PR #19, opened 2021-01-30
Typo

victorcuervo:
The right word is "árabe" instead of "árave".


chrisdavidmills:
Apologies, but we're not currently accepting changes to translations. Read https://hacks.mozilla.org/2020/12/an-update-on-mdn-web-docs-localization-strategy/ for an explanation.

PR #25, opened 2021-02-03
update tfoot translation to be HTML5 conform

d3tr0id:
since HTML5 it's not allowed anymore to have a `<tfoot>` elements after a `<thead>` .The English documentation is correct, but the german, french and Spanish ones need to be updated. See here: https://www.w3.org/TR/html51/tabular-data.html#elementdef-table


SphinxKnight:
Sorry, but this locale is currently frozen and changes/fixes are not currently accepted. Please see https://hacks.mozilla.org/2020/12/an-update-on-mdn-web-docs-localization-strategy/ and [README](https://github.com/mdn/translated-content/blob/main/README.md) of this repo.

PR #27, opened 2021-02-05
fix some macros for zh-cn

fiji-flo:


fiji-flo:
Some of these are still broken, but less broken!

escattone (commented):
The first thing that struck me is how strange it is that `files/uk/web/http/заголовки/accept/index.html`, which is located under the Ukrainian locale `uk`, is written in Chinese! Should we just delete that page?

escattone (approved):
@fiji-flo These changes look great to me. That strange Ukrainian page with Chinese content is a separate issue.

PR #70, opened 2021-02-25
use class="hidden" not style="display:none"

peterbe:
First of all, this solves the problem with `findMatchesInText()` in the flaw checker. 
Second of all, this is and important change because the Yari builder will be able to remove all things `class="hidden"` from the HTML before it goes into the React SSR rendering. 
A lot of live samples are based on the CSS/HTML/JS being within the HTML itself as hidden `div` tags. The cleanup of that can't work if they're `style="display:none"` but it can if it's `class="hidden"`.


fiji-flo (approved):
👍

PR #72, opened 2021-02-25
fix picky WebExtAllCompatTables invocations

fiji-flo:


escattone (approved):
Great catch @fiji-flo!

PR #78, opened 2021-03-01
add CI

fiji-flo:
Somewhat depends on https://github.com/mdn/content/pull/2773 so it won't fail on issues with the content repo.

### DIFF
--- a/.github/workflows/pr-check_redirects.yml
+++ b/.github/workflows/pr-check_redirects.yml
@@ -43,8 +43,8 @@ jobs:
 
       - name: Check redirects file(s)
         env:
-          content_root: ${{ github.workspace }}/mdn/content/files
-          content_translated_root: ${{ github.workspace }}/files
+          CONTENT_ROOT: ${{ github.workspace }}/mdn/content/files
+          CONTENT_TRANSLATED_ROOT: ${{ github.workspace }}/files
         working-directory: ${{ github.workspace }}/mdn/content
         run: |
           yarn content validate-redirects --strict
--- a/.github/workflows/pr-check_redirects.yml
+++ b/.github/workflows/pr-check_redirects.yml
@@ -1,0 +1,55 @@
+name: Check Redirects
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - files/**/_redirects.txt
+      - .github/workflows/pr-check_redirects.yml
+
+jobs:
+  check_redirects:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/checkout@v2
+        with:
+          repository: mdn/content
+          path: mdn/content
+          # Yes, this means fetch EVERY COMMIT EVER.
+          # It's probably not sustainable in the far future (e.g. past 2021)
+          # but for now it's good enough. We'll need all the history
+          # so we can figure out each document's last-modified date.
+          fetch-depth: 0
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v2.1.4
+        with:
+          node-version: "12"
+
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - uses: actions/cache@v2.1.4
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install all yarn packages
+        run: |
+          cd $GITHUB_WORKSPACE/mdn/content/
+          yarn --frozen-lockfile
+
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-check_redirects.yml
```diff
@@ -0,0 +1,55 @@
...
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install all yarn packages
+        run: |
+          cd $GITHUB_WORKSPACE/mdn/content/
```
peterbe, line 46: Instead of this, you can add, between the `name:` and `run:`:
```
working-directory: mdn/content
```

.github/workflows/pr-check_redirects.yml
```diff
@@ -0,0 +1,55 @@
...
+          yarn --frozen-lockfile
+
+      - name: Check redirects file(s)
+        run: |
+          export CONTENT_ROOT=$GITHUB_WORKSPACE/mdn/content/files
+          export CONTENT_TRANSLATED_ROOT=$GITHUB_WORKSPACE/files
```
peterbe, line 52: Can you move these into an env block above instead. So it becomes:

```yml
      - name: Check redirects file(s)
        env:
          CONTENT_ROOT=${{ github.workspace }}/mdn/content/files
          CONTENT_TRANSLATED_ROOT=${{ github.workspace }}/files
        working-directory: ${{ github.workspace }}/mdn/content
        run: |
          yarn content validate-redirects --strict
```

.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - uses: actions/checkout@v2
+        with:
+          repository: mdn/content
+          path: mdn/content
+          # Yes, this means fetch EVERY COMMIT EVER.
+          # It's probably not sustainable in the far future (e.g. past 2021)
+          # but for now it's good enough. We'll need all the history
+          # so we can figure out each document's last-modified date.
+          fetch-depth: 0
```
peterbe, line 27: Why this?

.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - uses: actions/checkout@v2
+        with:
+          repository: mdn/content
+          path: mdn/content
+          # Yes, this means fetch EVERY COMMIT EVER.
+          # It's probably not sustainable in the far future (e.g. past 2021)
+          # but for now it's good enough. We'll need all the history
+          # so we can figure out each document's last-modified date.
+          fetch-depth: 0
```
peterbe, line 34: Why do we need this?

peterbe (commented):
Before we merge this, we need *some* solution to keep this in sync. 

First of all, do you think you can rearrange the order of the checkouts. In particular, the way the `translated-content` is checked out. If we do that a lot of the `pr-test.yml` file will be a LOT more similar to the `pr-test.yml` in `mdn/content`. E.g.
Check out `mdn/content` to path `.` and *then* check out `translated-content` into `translated-content`. Like:

```
      - uses: actions/checkout@v2
        path: translated-content
      - uses: actions/checkout@v2
        repository: repository: mdn/content
        path: .
```

Then, we might be able to operate like you and I do on our own laptop terminals. We operate within the `mdn/content` but optionally run `export CONTENT_TRANSLATED_ROOT=...`

This might make the `.github/worksflows/*.yml` files much more similar and hopefully easier to maintain the sync.

fiji-flo:
@peterbe Initially I wanted to have a better folder structure but as far as I could tell `get-diff-action` requires the repo you're diffing to be checked out to `.`

peterbe:
> @peterbe Initially I wanted to have a better folder structure but as far as I could tell `get-diff-action` requires the repo you're diffing to be checked out to `.`

Good point! Didn't think about that.

fiji-flo:
@peterbe I think I resolved all of the gh details.

peterbe (approved):
Let's just do it. We are going to get out of sync sooner or later but there's no brilliant way to keep them in sync. We're doing something similar with `prod-build.yml` and `stage-build.yml` which are 95% the same. 

Let's see if this works.

.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,125 @@
...
+      - uses: technote-space/get-diff-action@v4.0.5
+        id: git_diff_content
+        with:
+          PATTERNS: files/**/*.html
+          SET_ENV_NAME: GIT_DIFF_CONTENT
+
+      - name: Build changed content
+        if: ${{ env.GIT_DIFF_CONTENT }}
+        env:
+          content_root: ${{ github.workspace }}/mdn/content/files
```
peterbe, line 59: Does that work?!
Normally, it's the same uppercase as we use it. I.e.
```suggestion
          CONTENT_ROOT: ${{ github.workspace }}/mdn/content/files
```
Either way, I wish we would write it out in upper case because it makes it more recognizable and easier to read.

fiji-flo: If I read the docs correct that should work but let's upper case them.

PR #83, opened 2021-03-03
Fix: dir=rtl in ja/web/javascript/reference/global_objects/number

kazuma0129:
Fixed alignment of code block.


peterbe:
@kazuma0129 Hi! It's not your fault but the CI tests failed here because of something related to redirects which changed just a couple of days ago. 

@fiji-flo Can you investigate the CI failures? Could it be related to that thing that landed on Friday?

mfuji09 (approved):
適切な修正だと思います。
I think it is a right correcting.

PR #88, opened 2021-03-08
sync translated content

fiji-flo:


chrisdavidmills (approved):
Well, I can't see anything here that looks alarming, but at the same time I don't really have any context for what you are doing here. It looks like setting some docs to have orphaned or conflicting status, and adding redirects to their new locations. Anything else I need to know here?

fiji-flo:
@chrisdavidmills This reflects the changes in _en-US_. If their counterpart has been removed without a redirect they are now *orphaned* if a redirect was added they moved to conflicting. We should come up with a policy for this.

Something like: localizers have x weeks to address these then we delete them and we delete them immediately for tier-2 locales.

chrisdavidmills:
I am happy to add something to the README to address this. So how about

---

## Synchronization with the en-US document structure

Before unfreezing the Tier 1 locales, we made an update to synchronize all the localized document tree structures with the `en-US` tree structure (English slugs only), to make the documentation easier to manage. This resulted in two new buckets of documents being created for each locale, existing as subdirectories of each local folder:

* `orphaned` — documents that are not associated with any parent `en-US` page.
* `conflicting` — documents with duplicate translations in existance (e.g. localized once under the existing `en-US` slug, and then again under a localized slug). The duplicate(s) are put in this folder.

Active locale maintenance teams are invited to spend some time exploring the orphaned and conflicting documents, to see whether any of this work is worth keeping (either adding to, or merging with an existing document in, the main tree), or whether it can just be deleted.

### Periodic synchronization updates

We periodically update the localized documentation to keep it in sync with the `en-US` tree structure, for example if documents are deleted for, or moved to a different location, in the tree.

When a synchronization occurs:

* Tier 1 (active) locale maintenance teams are given two weeks to decide what to do with the affected documents in their locales to keep things in sync.
* Tier 2 (frozen) locales have the affected documents deleted/moved immediately.

---

Does that work for you, @fiji-flo ? Let me know if you think this needs anything else. Maybe a mention of how often this sync is performed would be good (whatever works for you is fine), and also we should say how the sync request is communicated to the active locale maintenance teams?

fiji-flo:
I'm working on a daily sync action right now. 

Reads great. One addition for more context. `conflicting` will happen most likely if a documents get merged (I see this a lot when we _inline_ content like the `Foo/Bar` becomes just a section in `Foo` then we usually redirect `Foo/Bar -> `Foo#Bar` and this will result in a conflict as the sync job tires to more the translated `Foo/Bar` to `Foo` according to the redirect but `Foo` already existed.

PR #92, opened 2021-03-10
Fix typo in sample result

kptdobe:


SphinxKnight:
@kptdobe merci pour la correction :), c'est tout bon.
---
Thanks for the PR, looks good to me.

SphinxKnight (commented):
Forgot to add a review: looks good :)

PR #94, opened 2021-03-10
add sync cron action

fiji-flo:
Adds a GitHub Action that runs `yarn content sync-translated-content` once a day an pushes back any changes to `translated-content:main`


peterbe (approved):
If it will works, it works :)
Let's try this!

Issue #97, opened 2021-01-01
typos in french version


voronamanga:
https://developer.mozilla.org/fr/docs/Web/API/Element/getBoundingClientRect :
Title "Syntaxe", subtitle "Valeur de retour", § mdn/content#2: "est définit" -> "est défini" (grammar);
Title "Spécifications", subtitle "Notes", § mdn/content#2: "problèmes de comptabilité" -> "problèmes de compatibilité" (typo).

chrisdavidmills:
@SphinxKnight moved! I also had a look, and it turns out that to transfer issues you need write permissions on both the "from" and "to" repositories. So I've just given you write permissions on this one as well.

You should start to see a "Transfer issue" link in the right-hand column of issues, near the bottom.

PR #108, opened 2021-03-11
FR - several typofixes for #97

Scttpr:
My bad this pull request is linked to #97


SphinxKnight:
@peterbe, @fiji-flo hello, sorry to disturb but I can't quite understand why the CI is failing on this one. The error about missing `env-cmd` does not seem to relate with any modification from the PR itself.
I've tried rerunning the jobs a few times in case it was a transient error but it still fails.

Could you let me know if you see something obvious? :s 
Thanks!

fiji-flo:
We do have cache issues. That's what causes this. Working on it. @Scttpr could you rebase this on current main?

SphinxKnight:
@Scttpr are you sure you want to close this one? I'm not against avoiding having response.ok here (as it's not really the initial scope of the PR) but the rest looked fine :)

PR #110, opened 2021-03-11
fix fixable flaws in /es/docs/WebAssembly/Concepts

peterbe:
@fiji-flo This is part of testing the fixable flaws where the external image isn't downloaded but instead we simply switch to the en-US equivalent. 
I'll make the PR on mdn/yari shortly.


peterbe:
@fiji-flo ping

fiji-flo (approved):
👍

PR #114, opened 2021-03-12
fix caching

fiji-flo:


peterbe:
Why this?
What was wrong?

peterbe:
Caching the yarn download directory I generally find is performing much worse, over time, than just caching the whole `node_modules`. The yarn downloads aren't a huge time-spend, but the node_modules install is. Yes, a single change to `yarn.lock` will invalidate all the caching but I've found numerous times that it's worth blowing it away sometimes instead of saving a couple of seconds on some yarn package downloads.

fiji-flo:
I'm happy to revisit this. 
I've seen these: https://github.com/mdn/translated-content/runs/2092789996?check_suite_focus=true#step:8:58 a few times now.

peterbe:
In that case, it was just a bug. You copied the workflow from mdn/content and there it had its own `mdn/content/package.json` so it was probably something related to the `hashFiles()` and the "unusual" path. 
Do you want to revert this and fix it? Or if you're bored of it I can take a stab.

fiji-flo:
Feel free to take a shot.

PR #157, opened 2021-03-14
FIX: Fix the semicolon by a correct comma line 232 - Issue #101

### LINKED ISSUE #101: Typo in French article #l10n-fr
A user reported a typo in the French l10n Matrix channel regarding the semicolon used in https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Op%C3%A9rateurs/Affecter_par_d%C3%A9composition#valeurs_par_d%C3%A9faut_2

This is a trivial typo which could be fixed whenever the localization will be opened.

In the meantime, could you please triage it with the relevant label (#l10n-fr) please?

tristantheb:
**Related issue:** #101 
**Observed issue:** One block of code contained a sign error

**Commit changes:**
- [x] Typo: Fix a semilicon in the Javascript code block that should have been a comma
- [x] Tags is updated with french tags from en-US version

### DIFF
--- a/files/fr/web/javascript/reference/operators/destructuring_assignment/index.html
+++ b/files/fr/web/javascript/reference/operators/destructuring_assignment/index.html
@@ -2,10 +2,14 @@
 title: Affecter par décomposition
 slug: Web/JavaScript/Reference/Operators/Destructuring_assignment
 tags:
-  - ECMAScript 2015
-  - JavaScript
-  - Opérateur
-  - Reference
+- Destructuration
+- Affectation de déstructuration
+- ECMAScript 2015
+- ES6
+- JavaScript
+- Caractéristiques de la langue
+- Déstructuration des objets imbriqués et des tableaux
+- Opérateur
 translation_of: Web/JavaScript/Reference/Operators/Destructuring_assignment
 original_slug: Web/JavaScript/Reference/Opérateurs/Affecter_par_décomposition
 ---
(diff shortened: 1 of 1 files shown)


files/fr/web/javascript/reference/operators/destructuring_assignment/index.html
```diff
@@ -2,10 +2,14 @@
 title: Affecter par décomposition
 slug: Web/JavaScript/Reference/Operators/Destructuring_assignment
 tags:
-  - ECMAScript 2015
-  - JavaScript
-  - Opérateur
-  - Reference
```
SphinxKnight, line 8: I think Reference is missing in English and should not be deleted in fr (as far as I can tell when I'm looking at several pages in en-US / mdn/content)

SphinxKnight (changes requested):

SphinxKnight:
We need to see if this new platform is better for handling localized tags (in the Kuma era there were a few issues with them and I reverted my initial work on this). I'm not sure yet if we should generalize :/

After a few looks around, it seems they're only used in a few remaining KumaScript macros. Unless you think it's better to have them in French, I'd leave them in English for the time being. (Since they're not even displayed on the page, that wouldn't really matter to the end audience).

We can open an issue to discuss that more broadly.

files/fr/web/javascript/reference/operators/destructuring_assignment/index.html
SphinxKnight: After a few thoughts (cf. other comment in the PR), I'd leave those in English (but you may still add "Reference" ;) )

tristantheb: Updated with the missing "Reference" tag

tristantheb:
I put Reference back in the list of tags, but the trags were well translated into French ("Opérateur"), we'll see if we leave them in French or if we put everything back in English since many pages have French in their tags

SphinxKnight (approved):

SphinxKnight:
Allez ça part en prod un dimanche

PR #175, opened 2021-03-15
remove pre for CSSSyntax

fiji-flo:
Depends on mdn/yari#3244

```
s/<pre.*>(\{\{csssyntax(\(.*\))?\}\})\n?</pre>/$1/g
```


peterbe (approved):

chrisdavidmills (commented):
I'm assuming this is because you'll be generating the `<pre>` inside the macro instead?

Looking at the en-US article source, the `{{cssref}}` calls look to be wrapped in `<div>` rather than `<pre>`....are you intending to remove those as well?

fiji-flo:
@chrisdavidmills yes the pre moves to kumascript https://github.com/mdn/yari/pull/3244

As for the `cssref` (this is only addressing `csssyntax`), are they broken? We should to cleanup at some point I guess.

chrisdavidmills:
> @chrisdavidmills yes the pre moves to kumascript [mdn/yari#3244](https://github.com/mdn/yari/pull/3244)
> 
> As for the `cssref` (this is only addressing `csssyntax`), are they broken? We should to cleanup at some point I guess.

Ack, sorry, I got confused. the `{{csssyntax}}` calls in the pages I looked at in the `en-US` locale are also inside `<pre>`, e.g. `<pre class="brush: css">{{csssyntax}}</pre>`

PR #188, opened 2021-03-16
fail on unsafe html

peterbe:
Replica of https://github.com/mdn/content/pull/3187

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -61,7 +61,7 @@ jobs:
           # this to "*:error" which means the slightest flaw in the build
           # will break the build.
           # See https://github.com/mdn/yari/issues/715
-          export BUILD_FLAW_LEVELS="*:ignore"
+          export BUILD_FLAW_LEVELS="unsafe_html: error, *:ignore"
 
           # Note, it might be interesting to explore building *some*
           # pages in a PR build if the diff doesn't have any index.html


fiji-flo (approved):
🔥 the unsafe.

chrisdavidmills:
@lex111 share which PRs are failing wth us, and we'll have a look. The unsafe-html checks are specifically to catch things like executable JS being put into mdn source content, so shouldn't affect much of what you are doing.

lex111:
@chrisdavidmills please see https://github.com/mdn/translated-content/pull/203#issuecomment-801248474

Yes, I understand that, but given that we as translators work with legacy docs, I would not make such check a prerequisite for PR merging, at least for now.

chrisdavidmills:
> @chrisdavidmills please see [#203 (comment)](https://github.com/mdn/translated-content/pull/203#issuecomment-801248474)
> 
> Yes, I understand that, but given that we as translators work with legacy docs, I would not make such check a prerequisite for PR merging, at least for now.

I understand that it seems annoying, but I feel quite strongly that we need these checks in place when we are working with a lot of different communities.

I'd imagine that you could fix them fairly quickly by using an automated script.

@peterbe , do you have any thoughts here? Is there any admin-only override that could be used to get round this?

lex111:
I'm not against such checks, but please let me do a few more mass-edits PRs to further deal with each page in the usual workflow.

peterbe:
I'm trying to understand what's wrong with https://github.com/mdn/translated-content/pull/203 but I'll take a look. 
Basically, we can not do "admin-only override" because that'll just push the problem to another day. And an admin will have to look again. 

The inconvenience is real when all you want to do is fix typos. But I think it's a worthwhile thing overall. 
What I would recommend is; 
1. Pause the work on the mass-edit typo fixes
2. Make a PR that cleans up all existing unsafe HTML things in the whole `ru` tree. 
3. Once that clean-up is merged, go back to the mass-edit typo fix PR and merge in the latest `main` from upstream. 

I'll now take a look at what these unsafe HTML errors are.

lex111:
@peterbe it will take a very long time considering that there is no way to automate this for now. Why not just turn off this check for ru locale for one day, please?

peterbe:
> @peterbe it will take a very long time considering that there is no way to automate this for now. Why not just turn off this check for ru locale for one day, please?

I still don't think that's a good idea. 

However, I have now discovered that the Unsafe HTML doesn't come from the Russian content itself. It's from Yari!
It's the old `CompatibilityTable.ejs` macro that injects that HTML. I'll fix that right now.

PR #214, opened 2021-03-18
Fix: incorrect use atribute dir in code exemple

davipguimaraes:
The exemple code has atribute `dir=rtl`, but it's worn becose text write correct is `dir=ltr`

I remove this atribute to fix direction to writer

Error printscreen:
image


chrisdavidmills:
Apologies, but we are not accepting changes to this locale right now; see https://github.com/mdn/translated-content#tier-1-locales for an explanation of which locales are currently active, and why. https://github.com/mdn/translated-content#promoting-an-inactive-locale-to-tier-1 is also useful.

PR #219, opened 2021-03-18
enableVertexAttribArray : in french > attributes are disabled by defa…

Kurtil:
…ult, and not enabled


tristantheb (approved):
It's checked and it's ok

PR #221, opened 2021-03-18
Fix outdated German Wikipedia Link

andypillip:


chrisdavidmills:
Apologies, but we are not accepting changes to this locale right now; see https://github.com/mdn/translated-content#tier-1-locales for an explanation of which locales are currently active, and why. https://github.com/mdn/translated-content#promoting-an-inactive-locale-to-tier-1 is also useful.

PR #287, opened 2021-03-22
Unify "Syntax" heading in Russian translation

lex111:


myshov (approved):
Просто мимо пробегал, и можно было быстро поревьювить :)

hochan222:
@lex111  The branch is alive! If there are no more changes, is it okay to delete it?

PR #309, opened 2021-03-24
Update PT-PT docs for Math.ceil

carlos-menezes:
- translate "smallest" to PT-PT
- add note from English (US) version


SphinxKnight:
Sorry, but this locale is currently frozen and changes/fixes are not currently accepted. Please see https://hacks.mozilla.org/2020/12/an-update-on-mdn-web-docs-localization-strategy/ and [README](https://github.com/mdn/translated-content/blob/main/README.md) of this repo.

PR #314, opened 2021-03-24
Update index.html

rotival:
Encoding value "identity" is translated in "entity".


rotival:
fix typo on files/fr/web/http/headers/accept-encoding/index.html  entity instead of identity

PR #325, opened 2021-03-25
create general guidelines readme

chrisdavidmills:
I was talking to @lex111 about creating a doc to contain general guidelines for MDN translated content, so here it is.

This exists as a readme inside the docs directory, and so sits over the top of the locale-specific guides (we already have a ru one, and other locale teams are welcome to create their own locale-specific guideas required).

I'd also like to invite @SphinxKnight , @irvin , and @mfuji09 to comment on this PR. Ideas welcome.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,70 @@
+# General guidelines for MDN translated content
+
+In this README you can find a collection of general guidelines for translating
+MDN content, which apply to every locale.
+
+For guidelines relating to specific locales, we have locale-specific docs in
+subdirectories:
+
+- [Russian translation guide / Участие в переводе](ru/translation-guide.md)
+
+If you want to add a guide to document some specific guidelines for your locale
+and it does not already appear here, you are welcome to add one, or
+[talk to the locale teams](https://github.com/mdn/translated-content/blob/docs-readme/README.md)
+about it. Similarly, if you can think of a good general guideline that you'd
+like to add here, feel free to create an issue to talk about it.
+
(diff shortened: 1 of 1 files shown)


SphinxKnight:
Summoning @tristantheb for fr as well.

mfuji09:
I agree with it.
And I think it is needed to guide tranlators to the documents, from "readme.md" ("Contributing to the translated content of MDN Web Docs") and the MDN/Contribute/Localize page.

docs/README.md
```diff
@@ -0,0 +1,70 @@
...
+in the fr locale would be
+
+```html
+<h2 id="tutoriels">Tutoriels</h2>
+```
+
+We generally advise that you write all IDs in lower-case. The platform
+automatically converts them at render time anyway, but keeping them lower-case
+means that there is less chance of a manually-created anchor link not working
+because of the conversion.
```
SphinxKnight, line 37: Is there any technical reason to do that with Yari?
It seems to me that, with slugs not being translated anymore, it would a bit weird to have IDs translated (/foo/bar/baz#Baguette)

It doesn't look like IDs are dynamically generated as they were w/ Kuma.

Another benefit of keeping IDs untranslated is that it could allow comparison between en-US and localized page section per section.
For beginners' contribution, I would think it also makes it simpler not to bother with IDs.

chrisdavidmills: I mean, you are working on this content more than me, so if you'd rather not translate them, and that still works, then I'd be ok with that.

PR #336, opened 2021-03-27
Default :focus was not accessible

notabene:
> What was wrong/why is this fix needed? (quick summary only)

On the default `:focus` example the outline was removed completely, thus making it inaccessible to people using browsers that don't support `:focus-visible`

> MDN URL of main page changed

https://developer.mozilla.org/fr/docs/Web/CSS/:focus-visible

> Issue number (if there is an associated issue)

(none)

> Anything else that could help us review it


SphinxKnight (approved):
Merci @notabene :)

notabene:
@SphinxKnight pleasure :)

PR #340, opened 2021-03-27
fr - Update / revamp translation after pr337

SphinxKnight:
This page contained outdated wordings and subjective information. I removed xref macros and fixed wordings.


tristantheb (approved):
It's ok ^^

PR #380, opened 2021-04-02
Correct some sentence in article Operators/void

skyzhao1223:
Correct some sentence


irvin (changes requested):
https://github.com/mdn/content/blob/979f68fee5b9f9c1d0a352f033900736109f7362/files/en-us/web/javascript/reference/operators/void/index.html#L65
原文如下，我認為沒有提到「用戶點擊」的動作，還請參照調整。
`When a browser follows a <code>javascript:</code> URI`

PR #395, opened 2021-04-03
fr - Orphaned cleaning - Media event - Parity w/ en-US 23272ec

SphinxKnight:
This cleans a new orphan after https://github.com/mdn/content/pull/3663.
I've added the correspond id / redirect and also took care of removing domxref from the event listing


tristantheb (approved):
Looks good :D

PR #486, opened 2021-04-12
getting rid of all .alllink links

peterbe:
URLs like `/en-US/docs/tag/Base64` don't work anymore so no point in making the links.


myshov (approved):
Thanks!

PR #507, opened 2021-04-14
Fix escaped characters: 日本語

yomotsu:
[The section](https://developer.mozilla.org/ja/docs/Web/API/Document/createEvent#syntax) contains unexpected escaped characters and it breaks a link.
This PR is to fix the problem.

The problem in Japanese page
image

[Original page](https://developer.mozilla.org/en-US/docs/Web/API/Document/createEvent)
image


mfuji09 (approved):
良いと思います。
なお、日本語翻訳へのプルリクエストは日本語で結構です。

PR #577, opened 2021-04-19
Fix error in french example of fetch

Webcretaire:
There must have been some unfortunate copy / paste from english here, the header variable name is wrong


tristantheb (approved):
Hello, thank you for this correction, it will be corrected shortly on the live server. You can see when it's deployed here: https://whatsdeployed.io/s/16d/mdn/translated-content

PR #589, opened 2021-04-19
replace class="in-page-callout webdev" with class="callout"

peterbe:
We did this monster search-and-replace on the en-US content a long time ago but couldn't do it for the translated-content because it wasn't KS rendered at the time. Now it is. 

Once this is merged, we can finally delete this code from Yari:
https://github.com/mdn/yari/blob/b7e3ea7ed56f490c49a390fd99dda7e3c758501f/build/index.js#L157-L166


peterbe:
@schalkneethling I'm flagging you for review because you and I did this mass edit manually on the en-us content.

schalkneethling (approved):
🧹 Thanks!

PR #608, opened 2021-04-21
Replaced the heading

CreaTorAlexander:
Fixes  mdn/content issue https://github.com/mdn/content/issues/4333


irvin:
there is misunderstanding of the issue, the current PR is not fixing it.

PR #720, opened 2021-05-01
Fix: correcting link target id

ThesllaDev:
The link and id values were wrong and was not leading to the correct section


tristantheb:
Hello, this locale is currently frozen and `issues`/`pull request` are not currently accepted. Please see https://hacks.mozilla.org/2020/12/an-update-on-mdn-web-docs-localization-strategy/ and README of this repo.

Issue #692, opened 2021-04-29
KS: Removal of the least used macros - l10n-ko 


hochan222:
In order to get out of the risk of collision, to receive PR each time we create an issue separately, we plan to receive and process individual PRs, not only the above item macro but also other macro processing. (#613)

All macro related PRs will be managed here.

### 🎫 Related issues
See main PR : #611
mdn/yari#3063

### 👀 Observed issues
KumaScript macros are being phased out, to help the Yari dev team, here is a list of macros that will disappear from your files.

I invite you to check your files in your languages to make sure it doesn't break any translations. You can edit them directly on my branch. If you want to take care of it after merging this request, feel free :)

### 🔁 Changes in commits
*This list is taken from the issue quoted above, the numbers are not current.*
```
kumascript/macros/原語併記.ejs                         88
kumascript/macros/HTML5ArticleTOC.ejs              77
kumascript/macros/EditorGuideQuicklinks.ejs        42
kumascript/macros/gecko_callout_heading.ejs        38
kumascript/macros/IRCLink.ejs                      26
kumascript/macros/ref.ejs                          25
kumascript/macros/gecko_minversion_note.ejs        22
kumascript/macros/h1_gecko_minversion.ejs          20
kumascript/macros/FontAwesomeIcon.ejs              11
kumascript/macros/endnote.ejs                      11
kumascript/macros/h2_gecko_minversion.ejs          10
kumascript/macros/UserLink.ejs                     9
kumascript/macros/訳語.ejs                           6
kumascript/macros/LXRSearch.ejs                    2
kumascript/macros/tree.ejs                         2
kumascript/macros/B2GOnlyHeader2.ejs               1
```

```diff
! Only macros used less than 100 times for the moment.
! The pull request would be too big to process everything at once.
```

The compatibility tables are not done in this pull request (except when I stumble upon them). They require a query of their own, as there are many of them.

## Special note
You have time to do this, it is not an emergency at the moment. You can also do the other macros, this branch is open for modification.
The branch is also open to the entire community of the relevant language, so feel free to add as a reviewer.

---

## 예시

1.
![image](https://user-images.githubusercontent.com/22424891/116571863-6b6c5a00-a946-11eb-9f18-24d7ac810fb6.png)

2.
![image](https://user-images.githubusercontent.com/22424891/116572072-9a82cb80-a946-11eb-9621-7c0b54f57700.png)

yin1999:
Those macros have been removed.

Issue #329, opened 2021-03-25
Issue with "SharedArrayBuffer":  Incorrect references?


kazuma0129:
MDN URL: 
- https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer
- https://developer.mozilla.org/zh-CN/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer
- https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer
- https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer

#### What information was incorrect, unhelpful, or incomplete?

section of both  `property` and `method` in `SharedArrayBufer prototype object` are not parsed.

#### Did you test this? If so, how?

<details>
<summary>MDN Content page report details</summary>

* Folder: `ja/web/javascript/reference/global_objects/sharedarraybuffer`
* MDN URL: https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer
* GitHub URL: https://github.com/mdn/translated-content/blob/main/files/ja/web/javascript/reference/global_objects/sharedarraybuffer/index.html
* Last commit: https://github.com/mdn/translated-content/commit/f7f5454b72bd74808d1a9cf8cbaf69d8d0b378b9
* Document last modified: 2021-03-24T22:13:29.000Z

</details>

hochan222:
#519

PR #728, opened 2021-05-02
[FIX] parse err for method, property, sharedarraybuffer(#329)

hochan222:
fix parse err #329 

delete macro #692


hochan222 (commented):
check

PR #750, opened 2021-05-04
Method TRACE is safe

llfister:
There is a mistake when you are on french version, method trace is safe...


SphinxKnight (approved):
Merci pour cette correction !

PR #787, opened 2021-05-07
the example source is inverted

abelbaek:
**xpath location:** /html/body/div/div[1]/main/article/div[11]/pre

`<pre class="brush: js notranslate" dir="rtl">` Change to `<pre class="brush: js notranslate">`

bug image:
image


pje1740 (approved):
브라우저에서 확인했습니다 :) 고생하셨습니다!

PR #837, opened 2021-05-12
server-automation

hochan222:
## Purpose
Added bash and .gitmodules for server automation execution.

## Explanation
In the meantime, there was an inconvenience of having to clone two repositories and creating an .env file.
I made it possible to run it at once using gitmodules. Now we can run the server with only one command.

```sh
./start-server
```

This is a part of solving https://github.com/mdn/yari/issues/3730

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "content"]
 	path = content
 	url = https://github.com/mdn/content
+[submodule "yari"]
+	path = yari
+	url = https://github.com/mdn/yari.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "content"]
+	path = content
+	url = https://github.com/mdn/content
--- a/start-server.sh
+++ b/start-server.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 
-# Register content repo in git local config
+# Register content, yari repo in git local config
 git submodule init
-# Cloning content repo into translated repo
+# Cloning content, yari repo into translated repo
 git submodule update
 
 # Add .env file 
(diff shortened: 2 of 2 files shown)


peterbe:
No, the idea is that you use mdn/yari to start your preview server. But you add `CONTENT_TRANSLATED_ROOT` to the `.env` file as well. Here's what mine looks like:
```
# In my .env
CONTENT_ROOT=/Users/peterbe/dev/MOZILLA/MDN/content/files
CONTENT_TRANSLATED_ROOT=/Users/peterbe/dev/MOZILLA/MDN/translated-content/files
```

Now when I run `yarn start` I get all the translated content stuff on <HTTP://localhost:3000>

hochan222:
> No, the idea is that you use mdn/yari to start your preview server.

Oh... that's right..! I've seen translated-content run fine on HTTP://localhost:3000. Thank you. It seems to be a little more comfortable to contribute to the translated-content.

I changed this PR to run yari at once, is it okay?

peterbe:
We don't want git submodules. They're a possibility but leads to many other complications. For example, they're pointers to exact SHAs and it's unclear how they're supposed to be moved forward. 

We don't use bash for anything other than some of the CI. If you wanted to make some convenience scripts use `yarn` and `node` and the `scripts` portion in a `package.json`.

hochan222:
I understand. Thank you for reply. You must be busy, but for me it was time to learn the following three things. I'm very sorry and thank you. I'll think and try about how I can contribute in different ways. 

1. Clarify my goals and know exactly what to do.
2. If there is anything to contribute, put the PR draft first.
3. Not be afraid of failure because I learn there too.

Not using git submodules, you'd better close this PR! Could you please close it?

PR #842, opened 2021-05-12
fix external image and flaws on Web/MathML/Authoring (fr)

peterbe:
Leveraging https://github.com/mdn/content/pull/4946 to fix the broken external image URL. 
I also applied all the fixable flaws.


tristantheb (approved):
:+1: Fully checked and updated :) (In addition to the translation of a part that was not in French)

PR #852, opened 2021-05-13
fix EmbedLiveSample macro calls

peterbe:
This makes all the broken live samples working. 
I also applied all the fixable flaws.


nKiNk:
@peterbe Sorry for the late review. I found the fix was already updated to MDN. Is it Okay to close this PR?

peterbe:
> @peterbe Sorry for the late review. I found the fix was already updated to MDN. Is it Okay to close this PR?

You're right! Glad it's working now.

PR #858, opened 2021-05-14
Get rid of {{CompatibilityTable}} from HTML

alattalatta:
This PR replaces all {{CompatibilityTable}} and its component macros in HTML references to {{Compat}}.


yujo11 (approved):
확인했습니다! 고생하셨습니다!! 👍👍👍

PR #918, opened 2021-05-18
Get rid of {{CompatibilityTable}} from JavaScript

alattalatta:
This PR replaces all {{CompatibilityTable}} and its component macros in JavaScript references to {{Compat}}.

Additionally replaced spec tables of the pages to use {{Specifications}} macro.

See also #858, #859, #864.


pje1740 (approved):
고생하셨습니다! :)

PR #919, opened 2021-05-18
[FEAT] add Issue template

hochan222:
As the issue was raised, I received an inquiry because the form was not set to report the problem. https://github.com/mdn/translated-content/issues/907

We really needed some templates to help people file issues. I imported the issue template of the content repo. https://github.com/mdn/content/pull/2906

How about adding it?

### DIFF
--- a/.github/ISSUE_TEMPLATE/content-bug.md
+++ b/.github/ISSUE_TEMPLATE/content-bug.md
@@ -1,0 +1,33 @@
+---
+name: Content bug
+about: "Use this template to notify us of a content bug that needs fixing"
+labels: "needs-triage"
+assignees:
+title: "Content bug: <TITLE OF PROBLEM>"
+---
+
+## What page(s) did you find the problem on?
+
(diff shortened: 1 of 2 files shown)


SphinxKnight:
@mdn/core-yari-dev is it possible to review this?

.github/ISSUE_TEMPLATE/content-bug.md
```diff
@@ -0,0 +1,33 @@
+---
+name: Content bug
```
peterbe, line 2: Should we make it 

```suggestion
name: Translated content bug
```
?

peterbe (approved):
Actually, if this verbatim copy of mdn/content let's do it!

.github/ISSUE_TEMPLATE/content-bug.md
hochan222: Oh I thought it was a word meaning content, but it was a repository name! But you merged, I guess I was right!

PR #933, opened 2021-05-20
CREATE Glossary - Client Hints - FR version

stebenwolf:
Create the Clients Hints page.
Translated as "conseils de client" rather than "astuces client" but feel free to change it if "astuces" seems more appropriate

Matches item 141 on @tristantheb 's list


SphinxKnight:
Attention, il manque la propriété translation_of dans le préambule (je l'ajoute :))

SphinxKnight (approved):

PR #951, opened 2021-05-21
typo: French typo in role="presentation" page

mattlavio:
To validate the change, you can check [here](https://www.linguee.fr/francais-anglais/traduction/un+%C3%A9l%C3%A9ment+donn%C3%A9.html) for more examples with the same wording


SphinxKnight (approved):
Merci ! : )

PR #1019, opened 2021-05-28
Hochan222/ko translation guide

q464q:
可以合併嗎謝謝


hochan222:
Good morning. We've already done it at https://github.com/mdn/translated-content/pull/878, is there any problem? Can you please let me know if there is any problem?

PR #1096, opened 2021-06-03
Web/CSS/blend-mode を更新

mfuji09:
- 2021/05/08 時点の英語版に同期


files/ja/web/css/blend-mode/index.html
```diff
@@ -4,33 +4,32 @@
...
-  - CSS データ型
-  - CSS 合成
   - Compositing
+  - Compositing and Blending
+  - Data Type
   - Reference
+  - color
   - ブレンドモード
   - 合成
+browser-compat: css.types.blend-mode
```
peterbe, line 15: As of https://github.com/mdn/yari/pull/3955 you won't be needing to do this. In fact, I would discourage it. 
What you can do is omit this line and `<p>{{Compat}}</p>` will just work.

files/ja/web/css/blend-mode/index.html
```diff
@@ -4,33 +4,32 @@
...
   - CSS
-  - CSS Compositing
   - CSS Data Type
-  - CSS データ型
-  - CSS 合成
   - Compositing
+  - Compositing and Blending
+  - Data Type
   - Reference
+  - color
```
peterbe, line 12: As far as I know, and I'm reasonably confident, the tags never matter in translated content. 
All the KS macros I know of use the en-US document's tags for figuring out what and which links to include.

mfuji09: > As far as I know, and I'm reasonably confident, the tags never matter in translated content.
> All the KS macros I know of use the en-US document's tags for figuring out what and which links to include.

So, how should we do? Is it incorrect to adjust tags in translated articles?

peterbe: Yes it is. Because it won't be of any use. You could but all or anything in there and it wouldn't affect anything. 
We currently don't have the ability to filter-search by tags in the new site-search. Mostly, because the new site-search is so much better, there's no need. 

I can't think of any reason to keep the `tags` in the translated content. But I'm only 95% certain and it would take some time and investment to make that 100%. 

Perhaps you can prove me wrong. Perhaps there's some `.ejs` macro that actually does depend on the *translated* document's tags (unlike, say, the CSSRef sidebar which always relies on the tags in the `en-US` docs)

files/ja/web/css/blend-mode/index.html
mfuji09: Thanks, but I can't omit this line because the {{Compat}} macro is reported as a flaw if the line is omitted.
Please modify the English version first. After it has been accepted, we will adjust as so.

peterbe: That must mean your `yari` is out-of-date. I think. 
And the latest en-US content has it in there: https://github.com/mdn/content/blob/e60dc3616e284c9b36b17a4c12e694984e3ebfde/files/en-us/web/css/blend-mode/index.html#L13

Would you mind trying to merge in the latest upstream `main` into this branch and see if it stops being a flaw?

PR #1102, opened 2021-06-03
use plain Compat macro (sample, fr)

peterbe:
Thanks to https://github.com/mdn/yari/pull/3955 we can no use just `{{Compat}}` instead of `{{Compat("that.bcd.compat.id")}}` if the en-US parent document as a `browser-compat` key in the front matter, which I think it does for every single JS reference page. 

This PR is just a demo to prove that.


peterbe:
CC @SphinxKnight

SphinxKnight:
I'll check this one today. Thanks @peterbe

SphinxKnight (approved):
LGTM, I've seized the occasion to update the content as well and also updated the specification table as it worked locally. Let's see how it goes live :)

PR #1112, opened 2021-06-06
Firefox 89 for developers を翻訳

yyss:
Jun 3, 2021 の英語版に対応


SphinxKnight (changes requested):
This PR contains files which are unrelated to the topic of the PR itself. The commits' history should only target /ja/ pages. If you're unsure what to do with Git, you can start a new PR and reapply your changes on top of a new/clean branch.
If you want to keep this PR, some effort will be needed to fix this.

PR #1208, opened 2021-06-15
fr - Update JS Date.getMonth() example 

Eric013:
Fix of the variable name Xmas95 to noel95 to match the example above


SphinxKnight (approved):
Merci pour la correction @Eric013 :)

Issue #1008, opened 2021-05-28
All hidden paragraphs about Browser Compatibility can be deleted


peterbe:
See https://github.com/mdn/content/issues/2228
Basically, in mdn/content all occurrences of HTML like this:
```html
<div class="hidden">The compatibility table on this page is generated from structured data. If you'd like to contribute to the data, please check out <a href="https://github.com/mdn/browser-compat-data">https://github.com/mdn/browser-compat-data</a> and send us a pull request.</div>
```
or 
```html
<div class="hidden">
<p>The compatibility table in this page is generated from structured data. If you'd like to contribute to the data, please check out <a href="https://github.com/mdn/browser-compat-data">https://github.com/mdn/browser-compat-data</a> and send us a pull request.</p>
</div>
```
or
```html
<p class="hidden">The compatibility table in this page is generated from structured data.
  If you'd like to contribute to the data, please check out <a
    href="https://github.com/mdn/browser-compat-data">https://github.com/mdn/browser-compat-data</a>
  and send us a pull request.</p>
```

These can all be deleted from the source content.

peterbe:
Here's a great example of a PR that does it: https://github.com/mdn/content/pull/5351/files

peterbe:
@mdn/yari-content-fr @mdn/yari-content-ru @mdn/yari-content-ko @mdn/yari-content-ja @mdn/yari-content-pt-br @mdn/yari-content-zh 
I encourage you to grab https://gist.github.com/peterbe/8e64574dcfea77b73f60d0c5da9e07b2 and download it and modify it. As you can see there, in the first iteration I made it focus on `files/ja/web/http` and as you can see the `NEEDLE` string is for Japanese. 
WIth that script, I was able to create https://github.com/mdn/translated-content/pull/1009
Please have a look to see if you can use that automation and/or write your own automation and/or doing some manual search-and-replace.

t7yang:
@mdn/yari-content-zh done by PR #1010.

sashasushko:
Also removed in RU by #1062 

@t7yang thx for regexp

tristantheb:
Hi, I started this cleaning for the FR version, it will require some time and some PR because there are 1560 results in total...

---
[...]

PR #1246, opened 2021-06-19
Remove bcd hidden in JS section for fr

SphinxKnight:
This is for helping on https://github.com/mdn/translated-content/issues/1008
It targets the JavaScript section for the French locale. It should not contain any other document nor do anything else than removing such hidden part for bcd.


tristantheb (approved):
Let's go to prod !!! \o/
Part 2 of #1008

Issue #1231, opened 2021-06-17
All uses of $samples URLs are deprecated and need to be replaced


peterbe:
For example: See the diff in https://github.com/mdn/translated-content/pull/1230/files

Things like:
```html
<iframe class="live-sample-frame sample-code-frame" frameborder="0" height="180px" id="frame_Live_example" 
 src="https://mdn.mozillademos.org/en-US/docs/Web/API/Element/compositionstart_event$samples/Live_example?revision=1606277" width="100%"
 ></iframe>
```
is deprecated and needs to cleaned up. In `en-US` all instances of these have been replaced with `EmbedLiveSample`. 

Please help hunt these down to upgrade. The output of these URLs will soon stop working because it frees up MDN to stop storing the old WIki revisions from 2019 and older. 

Note; it's not just `<iframes>` there are sometimes links too. Try to replace them with regular links to file attachments.

peterbe:
@mdn/yari-content-fr 
```
▶ rg -i '\$sample'  --count-matches
web/html/element/img/index.html:1
web/svg/attribute/text-anchor/index.html:1
mdn/structures/live_samples/index.html:1
```

@mdn/yari-content-ja 
```
▶ rg -i '\$sample'  --count-matches
web/html/element/img/index.html:1
```

@mdn/yari-content-ru 
```
▶ rg -i '\$sample'  --count-matches
mdn/structures/live_samples/index.html:1
```

@mdn/yari-content-pt-br 
```
▶ rg -i '\$sample'  --count-matches
learn/javascript/first_steps/what_is_javascript/index.html:1
```

@mdn/yari-content-zh 
```
▶ rg -i '\$sample'  --count-matches
web/api/htmlelement/input_event/index.html:1
web/api/htmlelement/beforeinput_event/index.html:1
web/api/mouseevent/screeny/index.html:1
web/api/mouseevent/screenx/index.html:1
web/api/range/selectnodecontents/index.html:1
web/api/range/surroundcontents/index.html:1
web/api/htmlinputelement/setselectionrange/index.html:1
web/api/htmlselectelement/selectedindex/index.html:1
web/api/datatransfer/getdata/index.html:1
web/api/document/readystatechange_event/index.html:1
web/api/domtokenlist/index.html:2
web/api/element/mouseout_event/index.html:1
web/html/element/img/index.html:1
web/html/element/input/index.html:22
learn/html/multimedia_and_embedding/adding_vector_graphics_to_the_web/index.html:2
[...]

PR #1274, opened 2021-06-21
remove hidden bcd table paragraph from 'zh-cn'

peterbe:
Part of #1008


peterbe:
Ah. This is blocked on https://github.com/mdn/translated-content/issues/1231

PR #1285, opened 2021-06-22
remove hidden bcd table paragraph of 'ja' (part 2)

peterbe:


peterbe:
I found many more than https://github.com/mdn/translated-content/pull/1183/files

peterbe:
It's too big! Lemme break it down a bit.

PR #1320, opened 2021-06-27
Add some redirection for fr Intl link

romain-trotard:
Hello :)

Currently on the french translations for `Intl` in the Sidebar, the links do not lead to the right page as you can see in the gif below
IntlMdnNotRedirected

Actually the URLs in the sidebar are redirect to new ones with `.../Global_Objects/Intl/...` in the path.
In the PR I just added the redirection for the sidebar url because the pages for method are not translated yet (link to en-US version). I think it would be better if they are added when translating the pages. But if you think it should be added in this PR, I can add them ;)

Have a good day


SphinxKnight:
Thanks a lot @romain-trotard ! I think this is due to the way the `{{JSRef}}` macro (see https://github.com/mdn/yari/blob/main/kumascript/macros/JSRef.ejs ) works.

I'll have a look to review this :)

SphinxKnight (approved):
Merci beaucoup :) J'ai rédigé https://github.com/mdn/yari/pull/4096 pour éviter à d'autres locales d'avoir à apporter le même "contournement"

PR #1335, opened 2021-06-28
Fix the translated description of connectedCallback

tnzk:
Fixed the Japanese translation of `connectedCallback` since it doesn't seem to explain how the callback would be invoked correctly.

I'm just a stranger for this project who haven't read the guidelines for contribution, so ignore this if doesn't follow the rules.


mfuji09 (approved):
良いと思います。

PR #1410, opened 2021-07-06
fix: update the variable 'XMLHttpRequest' to 'xhr'

arginsen:
update the variable 'XMLHttpRequest' to 'xhr', cause 'xhr' is defined above.


t7yang (commented):
`XMLHttpRequest.DONE` is a enum, nothing wrong here.

PR #1418, opened 2021-07-06
delete conflicting/orphaned in es

peterbe:
Ever since early spring 2021, we decided to not index the conflicting (and orphaned) documents but leave them in the repo. These documents will not be found by Google searching or our own internal site-search.
The original motive was to keep these around until each localization team has decided that the slug that should be kept is more up-to-date.


fiji-flo (approved):
👍

PR #1423, opened 2021-07-06
delete conflicting/orphaned in ko

peterbe:
Ever since early spring 2021, we decided to not index the conflicting (and orphaned) documents but leave them in the repo. These documents will not be found by Google searching or our own internal site-search.
The original motive was to keep these around until each localization team has decided that the slug that should be kept is more up-to-date.


hochan222:
메모: 유사 내용의 대체된 페이지들 링크 찾기

peterbe:
@hochan222 If you think it's ready, can you review and merge?

hochan222:
@peterbe I read `Files Changed` about half, but I have a question. Most of the conflicting(and orphaned) documents were translated in 2013, and it is confirmed that the current version has been replaced. However, the content of the conflicting documents and the latest documents are completely different.

 Is there any data(or history) I can refer to as to what caused these documents to be separated from the current documents?

files/ko/orphaned/learn/how_to_contribute/index.html
```diff
@@ -1,106 +0,0 @@
----
```
hochan222, line 1: (a note just for me) check point for compare

hochan222 (approved):
LGTM :)

All were well replaced.

PR #1439, opened 2021-07-07
fix a bug of legacy http:// URL references

peterbe:


peterbe:
Sorry all reviewers pinged. Anybody can review this. And if you can merge, please go ahead and squash and merge.

hochan222 (commented):
LGTM :)

irvin:
i had rebase merge instead of squash, due to one thing per commit principle.

peterbe:
> i had rebase merge instead of squash, due to one object per commit principle.

I don't understand. If you use "Squash and merge" that's actually a form of rebase too. But it makes it so there's only 1 commit. 
If you use "Rebase and merge" it just pretends that all commits started from the tip of `main` but all commits are kept. 

We use "Squash and merge" on mdn/content and mdn/yari.

irvin:
one commit is “fix http:// URL” and another commit is “remove wiki. prefixes”, i think it would be better to keep them separate because it’s actually two separate issues?

peterbe:
> one commit is “fix http:// URL” and another commit is “remove wiki. prefixes”, i think it would be better to keep them separate and not merging into one commit because it’s actually two separate issues?

Fair! It was "sloppy" to fix two separate things in one PR. 

Great work! ...as always. Please keep it up!

PR #1468, opened 2021-07-09
labeler

### LINKED ISSUE #391: Meta - Automating labelling pull requests w/ matching l10n-tag
@chrisdavidmills / @fiji-flo, as PR are 'round-robin'ed according to code owners, would it be possible/sensible to automatically flag PRs with the maching "l10n" tag (e.g. "l10n-ja", "l10n-fr", "l10n-ru", "l10n-zh")?
This is something which is now manually done and having such tags helps knowing what PRs are pending for a given locale

+ @mdn/yari-content-ja / @mdn/yari-content-ru / @mdn/yari-content-zh / @mdn/yari-content-fr if you want to add anything.

peterbe:
Now any PR is automatically labeled with the right `l10n-*` label based on the files inside the PR. Nifty!

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,19 @@
+# Add 'repo' label to any root file changes
+repo:
+  - '*'
+
+# Add '@domain/core' label to any change within the 'core' package
+@domain/core:
+  - package/core/*
+  - package/core/**/*
+
+# Add 'test' label to any change to *.spec.js files within the source dir
+test:
+  - src/**/*.spec.js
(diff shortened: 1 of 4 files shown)


peterbe:
@escattone I've found one bug with the PR builds from the PR review companion. Because it doesn't build the "equivalent" en-US pages, also, it doesn't get the images from those en-us documents. 
I guess we should see about adding an option so that for every non-en-US file it builds, also automatically build the en-us one. Something for next week to ponder!

PR #1480, opened 2021-07-11
Fix typos

timeshiftseiya:
> What was wrong/why is this fix needed? (quick summary only)

日本語の入力間違い

> Issue number (if there is an associated issue)

No issue


mfuji09 (approved):
OKです。

PR #1483, opened 2021-07-11
fix: the 'eval' jump to the correct link

yhb-flydream:
the 'eval' has been translated


yhb-flydream:
the 'eval' has been translated

PR #1484, opened 2021-07-11
fix: the '主线程' jump to the correct link

yhb-flydream:
'主线程' has been translated


t7yang (approved):

t7yang:
fix: the '主线程' jump to the correct link #1484

PR #1508, opened 2021-07-12
[ja] fix: typo of global_objects date index

sititou70:
# What was wrong/why is this fix needed? (quick summary only)

日本語の修正

# Issue number (if there is an associated issue)

No issue


mfuji09 (approved):
OKです。

PR #1533, opened 2021-07-14
correct the description of zh-cn date getTimezoneOffset

liumulin614:
返回值部分关于正负值的描述有误，英文原文为：
The number of minutes returned by getTimezoneOffset() is positive if the local time zone is behind UTC, and negative if the local time zone is ahead of UTC.


irvin:
thanks

PR #1566, opened 2021-07-16
[fr] update WebExtensions API tabs.duplicate page

ariasuni:
It was lacking the `duplicateProperties` object introduced in Firefox 77.


peterbe:
The CI failed because of https://github.com/mdn/content/pull/6979 which is merged now.

PR #1573, opened 2021-07-16
purge all redirects to archived content

peterbe:
We used to keep a file `content/archived.txt` that was used to salvage some really ancient redirects in the translated content. This gets rid of all of those.


peterbe:
I'm going to boldly merge this because it solves the problem that lots of translated-content redirects pointed to a page that's archived and as of yesterday deleted. 
As of this PR, we no longer rely on the old `content/archived.txt` list of legacy redirects to archived content. And that's a very good thing. Refreshing even!

SphinxKnight:
Thanks @peterbe :)

escattone:
Nice work, thanks @peterbe!

PR #1574, opened 2021-07-17
Prepare JS French section for Markdown

SphinxKnight:
This is a follow-up of https://github.com/mdn/translated-content/discussions/1236
I tried to follow @wbamberg PRs and I assume review would be easier to follow commit per commit.

I did not yet try to convert the section using the conversion tool.


wbamberg:
Amazing! When you run the conversion tool you will get a report talking about all the things that couldn't get converted, and this may flag some extra things that need to get done. There's a bit of explanation of the report's format here: https://github.com/mdn/content/pull/5193#issuecomment-864328887. Of course please feel free to ask if you have questions :).

SphinxKnight:
Converting this as a draft as long as the output of the converter is not clean.
@tristantheb feel free to start this enchanting little review tho :x

SphinxKnight:
@tristantheb En me mettant à ta place, je me dis que celle-ci va être plutôt compliquée à revoir de façon classique. N'hésite pas si tu as besoin d'indications ou d'un coup de main (quitte à faire une revue "statistique" en prenant quelques docs au hasard...)

L'idée de cette PR est d'avoir un rapport de `yarn md h2m web/javascript --locale fr --mode dry` relativement clean (reste les tableaux, kbd, etc.
[md-conversion-problems-report-2021-07-30T17:08:26.459Z.md](https://github.com/mdn/translated-content/files/6909095/md-conversion-problems-report-2021-07-30T17.08.26.459Z.md)

SphinxKnight:
@wbamberg If you have a few moments, I'd really appreciate a look at the last report (cf. link from previous comment) to check if that seems ok for you or if I missed something.

tristantheb (approved):

wbamberg:
Hey @SphinxKnight .

I see a couple of things that might be worth a look:

## code (7)

7 unconvertible `code` elements. This is most likely because of a `code` element that contains other markup, where the other markup does not have the same text content as the `code`. For example:

    `<code>this one <em>here</em></code>`

The issue here is that you can't represent this directly in GFM. Something like:

    `this one *here*`

....will be rendered like 

`this one *here*`

...rather than:

<code>this one <em>here</em></code>

The converter is clever enough to invert the nesting in cases like:

    `<code><em>this one here</em></code>`

...to generate markdown like:

    *`this one here`*

....which renders properly:

*`this one here`*

But this trick only works, of course, if the inner content is the same as the outer content.

Probably the converter should just discard this internal markup, but so far it's been simpler to remove it manually in the content. I guess I removed them in https://github.com/mdn/content/pull/6880.

## img[alt="Stack, heap, queue"]...

Probably. isn't being converted because of inline styles?

***************

Apart from these there's nothing obviously needing work. It *might* be worth looking through some of the unconverted tables to see if any of them are easily fixed, but as noted we expect many tables to be unconvertible (and per https://github.com/mdn/content/issues/7342, sometimes we probably want to use HTML even if a table can be converted!).

Note that the summary only shows the top 20 unconvertibles, so there might be a few which aren't listed there. Honestly we should tweak the report so it shows them all.

Nice work, fr l10n people!

SphinxKnight:
Thanks @wbamberg, I'll address the code in another PR and then, I'll open the PR for HTML -> MD for JS section in fr :)

PR #1580, opened 2021-07-18
modified:   files/zh-cn/mdn/about/index.html

xinetzone:


t7yang:
You don't need to reformat content.

PR #1607, opened 2021-07-19
Fixed confusion between http headers and html head (RU)

zzzorgo:
There is term "HTML headers" mentioned in the russian translation where meant to be "HTTP headers"


sashasushko (approved):

sashasushko:
Thanks @zzzorgo

PR #1616, opened 2021-07-19
Remove previously orphaned file

peterbe:


escattone (approved):

irvin:
the files in this pr are not orphaned. I had moved them to the corresponding folder in last week's PR 1412.
It seems I miss adjusting the slug to the corrent path on these files.

peterbe:
> the files in this PR are not orphaned. I had moved them to the corresponding folder in last week's PR 1412.
> It seems I miss adjusting the slug to the correct path on these files.

We had to rush through this because of these files it broke the production builds.

irvin:
@peterbe no problem. Does the issue came from wrong slug?

peterbe:
> @peterbe no problem. Does the issue came from wrong slug?

Yeah, if the `slug: ...` value is not matching the name of the folder it sits in, then it crashes.

irvin:
> > @peterbe no problem. Does the issue came from wrong slug?
> 
> Yeah, if the `slug: ...` value is not matching the name of the folder it sits in, then it crashes.

ok, will make sure to fix those before bringing them back.

escattone:
@irvin Sorry! Like @peterbe said, we rushed this because these files were breaking the production build (their slug didn't match the name of the path of the folder they existed within).

PR #1617, opened 2021-07-19
remove all 'cke_pastebin' mentions

peterbe:
No idea what this is about but the "cke" is probably a relic from the distant past when we used "CKEditor" in the Wiki.


irvin (approved):

SphinxKnight (approved):
LGTM

PR #1655, opened 2021-07-22
Update index.html

xkx9431:
I think  this should update, right ?


t7yang (commented):
https://developer.mozilla.org/en-US/docs/Web/Performance/How_browsers_work

> To understand how to improve performance and perceived performance, it helps to understand how the browser works.

The original translation is correct.

PR #1679, opened 2021-07-24
Deleted a space after "！" following the rule

kisre:
Followed the rule of 約物, no space has after "？", and "！".
There is a space after "？" in this file because of the presence of "2".

https://github.com/mdn/translated-content/pull/911#


mfuji09 (approved):
OKです。

PR #1690, opened 2021-07-25
fix: word translation

henriquefontes:
"evolindo" to "evoluindo"


josielrocha:
Thank you @henriquefontes!!

PR #1705, opened 2021-07-26
PR test should build based on possible index.md too

peterbe:
`pr-test.yml` for translated-content is based on that for mdn/content, which [has a fix already for possible `index.md` files](https://github.com/mdn/content/blob/a4edee7f6efb876d94912b4853d44ab61b0b3b80/.github/workflows/pr-test.yml#L61)


peterbe:
This was based on this error: https://github.com/mdn/translated-content/runs/3159902851?check_suite_focus=true
which was the result of this PR: https://github.com/mdn/translated-content/pull/1698

PR #1714, opened 2021-07-26
remove link 'title' attributes that's just the 'href' (zh-cn)

peterbe:
The `title` tooltip on links is useful if it's just the same as the URL. 
I think the old Wiki used to by-default put these in.


irvin:
lgfm

PR #1731, opened 2021-07-27
remove link 'title' attributes that's just the 'href' (fr, part 2)

peterbe:
An extension of https://github.com/mdn/translated-content/pull/1713
I found a bug in my automation so it found many more instances that can be cleaned up now.


SphinxKnight (approved):
🔥 
Thanks : )

Issue #1634, opened 2021-07-21
Content bug: Lets get rid of all www.fxsitecompat.com mentions


peterbe:
## What page(s) did you find the problem on?

<details><summary>grep</summary>
files/ko/web/api/worker/index.html:<p>초기 명세에서는 교차 출처 워커 스크립트를 불러오려 시도하면 <code>SecurityError</code>가 발생했습니다. 요즘은 오류 대신 {{event("error")}} 이벤트가 발생합니다. <a href="https://www.fxsitecompat.com/en-CA/docs/2016/loading-cross-origin-worker-now-fires-error-event-instead-of-throwing-worker-in-sandboxed-iframe-no-longer-allowed/">이 글</a>에서 교차 출처 처리 방법을 알아보세요.</p>
files/ko/web/api/document/keyup_event/index.html:<p><font><font>Firefox 65부터 </font></font><code>keyup</code><font><font>CJKT 사용자의 브라우저 간 호환성을 개선하기 위해 </font><font>{{domxref ( "Document / keydown_event", "keydown")}} 및 </font><font>이벤트가 이제 IME 작성 중에 실행됩니다 ({{bug (354358)}}, 자세한 내용 </font></font><a href="https://www.fxsitecompat.com/en-CA/docs/2018/keydown-and-keyup-events-are-now-fired-during-ime-composition/"><font><font>은 IME 작성 중에 keydown 및 keyup 이벤트가 시작</font></font></a><font><font> 됩니다. </font><font>작성의 </font></font><code>keyup</code><font><font>일부인 </font><font>모든 </font><font>이벤트 </font><font>를 무시하려면 </font><font>다음과 같이 수행하십시오 (229는 </font></font><code>keyCode</code><font><font>IME에 의해 처리 된 이벤트 관련에 </font><font>대한 특수 값 세트입니다 </font><font>).</font></font></p>
files/ko/mozilla/firefox/releases/66/index.html: <li><a href="https://www.fxsitecompat.com/en-CA/versions/66/">Site compatibility for Firefox 66</a></li>
files/fr/mozilla/firefox/releases/50/index.html: <li><a href="https://www.fxsitecompat.com/en-US/versions/50">Site Compatibility for Firefox 50</a></li>
files/fr/mozilla/firefox/releases/69/index.html: <li><a href="https://www.fxsitecompat.com/en-CA/versions/69/">Site compatibility for Firefox 69</a></li>
[...]

PR #1741, opened 2021-07-27
fix: Lets get rid of all www.fxsitecompat.com mentions (#1634)

hochan222:
issue #1634

get rid of all `www.fxsitecompat.com` mentions


hochan222 (commented):
@peterbe Is there anything else I need to do? Can you merge it immediately after confirmation?

peterbe:
> Did the current PR do everything the issue expected?

Not sure I understand your question. 
But the issue is about getting rid of it in ALL locales. Once that's been done we can close the issue. 
Thanks for standing up and doing the Korean part!

peterbe (approved):

PR #1743, opened 2021-07-27
French translation for the `:nth-col` CSS pseudo-class reference.

audrasjb:
This PR creates a French translation for the `:nth-col` CSS pseudo-class reference.


SphinxKnight (approved):
👌

PR #1749, opened 2021-07-27
French translation for the `math-style` CSS property reference.

audrasjb:
This PR adds a French translation for the `math-style` CSS property reference.


SphinxKnight (approved):
🔥

PR #1756, opened 2021-07-27
French translation of learn/performance: add missing non-breaking spaces

audrasjb:
This PR adds missing non-breaking spaces on learn/performance.

### DIFF
--- a/files/fr/learn/performance/index.html
+++ b/files/fr/learn/performance/index.html
@@ -39,9 +39,9 @@ translation_of: Learn/Performance
 <dl>
  <dt><a href="/fr/docs/Learn/Performance/pourquoi_performance_web">Le "pourquoi" des performances Web</a></dt>
  <dd>Cet article explique pourquoi les performances Web sont importantes pour l'accessibilité, l'expérience utilisateur et vos objectifs commerciaux.</dd>
- <dt><a href="/en-US/docs/Learn/Performance/What_is_web_performance">Qu'est-ce que la performance Web?</a></dt>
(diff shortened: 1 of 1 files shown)


files/fr/learn/performance/index.html
```diff
@@ -39,9 +39,9 @@ <h2 id="Guides">Guides</h2>
...
- <dt><a href="/en-US/docs/Learn/Performance/Perceived_performance">Comment les utilisateurs perçoivent-ils les performances?</a></dt>
+ <dt><a href="/en-US/docs/Learn/Performance/What_is_web_performance">Qu'est-ce que la performance Web ?</a></dt>
+ <dd>Vous savez que les performances Web sont importantes, mais en quoi consistent les performances Web ? Cet article présente les composants de la performance, du chargement et du rendu de la page Web, en passant par la manière dont votre contenu est intégré dans le navigateur de votre utilisateur pour être visualisé, aux groupes de personnes dont nous devons tenir compte lors de la réflexion sur les performances.</dd>
+ <dt><a href="/en-US/docs/Learn/Performance/Perceived_performance">Comment les utilisateurs perçoivent-ils les performances ?</a></dt>
```
SphinxKnight, line 44: Entities (&amp;nbsp;) are preferred to litteral nbsps as they are somehow easier to spot for reviews.
A bit of history, the previous graphical editor for MDN had a quirk with some shortkeys / keys and now some pages are crippled by useless nbsp (even in en-US content). In other words, the "weight" of litteral nbsps is heavier than the typographic professionalism they bring.

Also, while we're here, links could use /fr/ instead of /en-US/ (even if the target page doesn't exist yet). I know this is not the topic of this PR and if you want to keep things atomic, let me know :)

SphinxKnight (changes requested):

files/fr/learn/performance/index.html
audrasjb: Fixed in a new commit

SphinxKnight (approved):

PR #1766, opened 2021-07-28
French translation of the CSS Counters Styles API reference

audrasjb:
This PR creates the French translation of the CSS Counters Styles API reference.

### DIFF
--- a/files/fr/web/api/css_counter_styles/index.html
+++ b/files/fr/web/api/css_counter_styles/index.html
@@ -1,11 +1,11 @@
 ---
-title: Module CSS Styles de compteurs
+title: Module CSS Styles de compteurs (Counter Styles)
 slug: Web/API/CSS_Counter_Styles
 translation_of: 'Web/API/CSS_Counter_Styles'
 ---
 <div>{{DefaultAPISidebar("CSS Counter Styles")}}{{SeeCompatTable}}</div>
 
-<p>Le module CSS Styles de compteurs permet de définir des styles de compteurs personnalisés qui peuvent être utilisés pour les marqueurs de listes CSS et les compteurs des contenus générés.</p>
+<p>Le module CSS Styles de compteurs (<i lang="en">Counter Styles</i>) permet de définir des styles de compteurs personnalisés qui peuvent être utilisés pour les marqueurs de listes CSS et les compteurs des contenus générés.</p>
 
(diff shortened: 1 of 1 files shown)


files/fr/web/api/css_counter_styles/index.html
```diff
@@ -0,0 +1,31 @@
+---
+title: Module CSS Styles de compteurs
```
SphinxKnight, line 2: I'm torn between the English "canonical term" for the name of the spec and the French equivalent here.
Maybe adding '(Counter Styles)' (within parenthesis) afterwards is a good compromise. I wonder how (French-speaking) people will discover this page: will they search for "counter style css" in their search engine, etc.

Let me know if you disagree or think of a better solution.

files/fr/web/api/css_counter_styles/index.html
```diff
@@ -0,0 +1,31 @@
...
+slug: Web/API/CSS_Counter_Styles
+translation_of: 'Web/API/CSS_Counter_Styles'
+---
+<div>{{DefaultAPISidebar("CSS Counter Styles")}}{{SeeCompatTable}}</div>
+
+<p>Le module CSS Styles de compteurs permet de définir des styles de compteurs personnalisés qui peuvent être utilisés pour les marqueurs de listes CSS et les compteurs des contenus générés.</p>
```
SphinxKnight, line 8: As this is inline, I would definitely add the English version in parenthesis here.

SphinxKnight (changes requested):
A slight """""debate"""" about adding the English name of the spec but apart from that: good to go.

files/fr/web/api/css_counter_styles/index.html
audrasjb: Agreed. New proposal in https://github.com/mdn/translated-content/pull/1766/commits/b819ee39bcf12645dec134db59438b571dbf2f48

files/fr/web/api/css_counter_styles/index.html
audrasjb: Agreed. Let's add both, then. New proposal in https://github.com/mdn/translated-content/pull/1766/commits/b819ee39bcf12645dec134db59438b571dbf2f48

SphinxKnight (approved):
Thanks :)

PR #1790, opened 2021-07-29
French translation of the HTMLIFrameElement csp reference

audrasjb:
This PR creates the French translation of the HTMLIFrameElement csp reference.

### DIFF
--- a/files/fr/web/api/htmliframeelement/csp/index.html
+++ b/files/fr/web/api/htmliframeelement/csp/index.html
@@ -1,0 +1,26 @@
+---
+title: HTMLIFrameElement.csp
+slug: Web/API/HTMLIFrameElement/csp
+browser-compat: api.HTMLIFrameElement.csp
+translation_of: 'HTMLIFrameElement.csp'
+---
+<p>{{SeeCompatTable}}{{APIRef("HTML DOM")}}</p>
+
+<p>La propriété <strong><code>csp</code></strong> de l'interface {{domxref("HTMLIFrameElement")}} spécifie la <a href="/fr/docs/Web/HTTP/CSP">Content Security Policy</a> (ou stratégie de sécurité du contenu) qu'un document embarqué doit accepter pour pouvoir s'auto-appliquer.</p>
+
+<h2 id="syntax">Syntaxe</h2>
+
+<pre class="brush: js">var <em>csp</em> = HTMLIFrameElement.csp
+HTMLIFrameElement.csp = <em>csp</em></pre>
+
+<h3 id="value">Valeur</h3>
+
+<p>Une content security policy (stratégie de sécurité du contenu).</p>
+
+<h2 id="specifications">Spécifications</h2>
+
+<p>{{Specifications}}</p>
+
+<h2 id="browser_compatibility">Compatibilité des navigateurs</h2>
+
+<p>{{Compat}}</p>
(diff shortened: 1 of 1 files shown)


files/fr/web/api/htmliframeelement/csp/index.html
```diff
@@ -0,0 +1,26 @@
...
+<p>La propriété <strong><code>csp</code></strong> de l'interface {{domxref("HTMLIFrameElement")}} spécifie la <a href="/fr/docs/Web/HTTP/CSP">Content Security Policy</a> (ou stratégie de sécurité du contenu) qu'un document embarqué doit accepter pour pouvoir s'auto-appliquer.</p>
+
+<h2 id="syntax">Syntaxe</h2>
+
+<pre class="brush: js">var <em>csp</em> = HTMLIFrameElement.csp
+HTMLIFrameElement.csp = <em>csp</em></pre>
+
+<h3 id="value">Valeur</h3>
+
+<p>Une content security policy (stratégie de sécurité du contenu).</p>
```
SphinxKnight, line 18: I'd leave only the French expression here -(I'd guess we can expect a reader who has made it this far that they know what a CSP is).
In order to be slightly better than the English docs, I would also add that this one is a string. (https://w3c.github.io/webappsec-cspee/#dom-htmliframeelement-csp) (or else, I wouldn't know what to do with such a value).

files/fr/web/api/htmliframeelement/csp/index.html
```diff
@@ -0,0 +1,26 @@
+---
+title: HTMLIFrameElement.csp
+slug: Web/API/HTMLIFrameElement/csp
+browser-compat: api.HTMLIFrameElement.csp
+translation_of: 'HTMLIFrameElement.csp'
+---
+<p>{{SeeCompatTable}}{{APIRef("HTML DOM")}}</p>
+
+<p>La propriété <strong><code>csp</code></strong> de l'interface {{domxref("HTMLIFrameElement")}} spécifie la <a href="/fr/docs/Web/HTTP/CSP">Content Security Policy</a> (ou stratégie de sécurité du contenu) qu'un document embarqué doit accepter pour pouvoir s'auto-appliquer.</p>
```
SphinxKnight, line 9: I think the intent of the English wording is closer to "qu'un document embarqué doit accepter de respecter" (cf. example from https://w3c.github.io/webappsec-cspee/#csp-attribute )

SphinxKnight (changes requested):
Slight modifications (due to the fact that the English page seems like a dry copycat of the spec :( )

files/fr/web/api/htmliframeelement/csp/index.html
audrasjb: Ah! Thank you, I had a hard time with that one sentence :) 
Fixed in https://github.com/mdn/translated-content/pull/1790/commits/f4d36cc635776c889b2ba9c37bd2e9743752a4f4

files/fr/web/api/htmliframeelement/csp/index.html
audrasjb: Agreed!
Fixed in https://github.com/mdn/translated-content/pull/1790/commits/f4d36cc635776c889b2ba9c37bd2e9743752a4f4

SphinxKnight (approved):
🙇‍♂️

PR #1800, opened 2021-07-29
French translation of the `preconnect` reference

audrasjb:
This PR creates the French translation of the `preconnect` reference.


SphinxKnight (approved):
👌

PR #1803, opened 2021-07-30
Update French translation of the `@media` reference

audrasjb:
This PR fixes a broken link on French translation of the `@media` reference.

### DIFF
--- a/files/fr/web/css/@media/index.html
+++ b/files/fr/web/css/@media/index.html
@@ -1,27 +1,20 @@
 ---
 title: '@media'
 slug: Web/CSS/@media
-tags:
-  - CSS
-  - Reference
-  - Règle @
+browser-compat: css.at-rules.media
 translation_of: Web/CSS/@media
 ---
 <div>{{CSSRef}}</div>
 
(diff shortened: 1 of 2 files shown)


SphinxKnight:
Hmm, I think the French page should stick to the way the English page is presenting information here. The {{page}} macro is a good tool as well as an old macro.
If that's OK, I'll fiddle a commit to fix this macro call and (hopefully) get the same result as in English rather than transforming the table into a link.

That being written: good spot :) and thanks for providing a fix.

audrasjb:
OK good point! Since I spotted the issue directly while using MDN as an end-user, I opened this PR directly from GitHub interface and not using my IDE so feel free to handle that the way you'd like :D

SphinxKnight:
Well, I think I'll actually update the whole page 😶

audrasjb:
:D feel free to close that PR then

SphinxKnight:
@audrasjb I'm shamelessly using this PR : ]
Let me know if this looks OK for you :)

audrasjb:
yeah sure, do as you wish!

tristantheb (approved):

PR #1804, opened 2021-07-30
[Ko] replace clientHeight example with image

raccoonback:
Replace a broken-looking example with an image, such as en-US.

#### ko
- https://developer.mozilla.org/en-US/docs/Web/API/Element/clientHeight

#### en-US
- https://developer.mozilla.org/ko/docs/Web/API/Element/clientHeight


hochan222 (approved):
LGTM :)

PR #1807, opened 2021-07-30
remove link 'title' attributes that's just the 'href' (ja, part 2)

peterbe:
Continuation of https://github.com/mdn/translated-content/pull/1712


mfuji09 (approved):
I think it is good.

PR #1820, opened 2021-07-31
French translation for the `prerender` reference.

audrasjb:
This PR creates a French translation for the `prerender` reference.


audrasjb:
I wonder what's with all these unwanted commits 🤔 
Not a blocker, but I need to investigate…

PR #1839, opened 2021-08-02
Removes any mention of fxsitecompat.com in fr

SphinxKnight:
This takes care of #1634 for French by:

- removing any mention in translated pages
- removing any page which was untranslated/incompletely translated/unmaintained (or orphaned for 1)


peterbe (approved):
Feels so much lighter!

PR #1840, opened 2021-08-02
remove link 'title' attributes that's just the 'href' (ja, part 3)

peterbe:
continuation of https://github.com/mdn/translated-content/pull/1807


mfuji09 (approved):
I think it is good.

PR #1850, opened 2021-08-02
delete all DevNews links in ja

peterbe:
There's no DevNews in en-US.


mfuji09:
How to handle DevNews is under discussion in the mozilla-japan community in https://github.com/mozilla-japan/translation/issues/555
Please wait until a conclusion is reached.

mfuji09 (approved):
I agree with this PR.

PR #1835, opened 2021-08-01
JavaScript section to Markdown - French

SphinxKnight:
Let's do this.

@tristantheb could I ask a favor of you to review how it plays on your side?

@wbamberg: this looks quite analogous to https://github.com/mdn/content/pull/7092, I hope I didn't miss a thing. Let me know if you spot something.

Last but least, @fiji-flo could you confirm or deny that https://github.com/mdn/yari/pull/4251 needs to be merged before this one?

PR #1863, opened 2021-08-03
Fix other h2m unhandled

SphinxKnight:
This fixes a bunch of content which I did not care of in the dry-run report of html 2 markdown conversion.
I've also removed useless spans.


SphinxKnight:
@tristantheb I hope you won't mind too much if I merge this one (as it's quite trivial and also... because I need it for #1835 :x )

PR #1893, opened 2021-08-04
Correct Content of MDN/Yari

pocketdr:


t7yang:
What about zh-TW?

pocketdr:
> What about zh-TW?

Same as zh-CN

PR #1936, opened 2021-08-06
Rewording of the `first_meaningful_paint` French glossary item

audrasjb:
This PR rewords the `first_meaningful_paint` French glossary item.


SphinxKnight (approved):
Let's not get complicated: I directly fixed a link but apart from that: all good.

PR #1940, opened 2021-08-06
French translation of the `Main thread` glossary item

audrasjb:
This PR adds a French translation for the `Main thread` glossary item.


SphinxKnight (commented):
I think adding "d'exécution" to have "Fil d'exécution principal" would be a bit more technically explicit (I think "fil" alone doesn't convey quite as much as thread/process in English).
Moreover, "fil d'exécution principal" matches what's already in Firefox: https://transvision.mozfr.org/?recherche=main+thread&repo=gecko_strings&sourcelocale=en-US&locale=fr&search_type=strings

audrasjb:
Makes sense to me @SphinxKnight 👍 
Fixed in https://github.com/mdn/translated-content/pull/1940/commits/34bb0a4e7cf8d99d5611f5afa557a98c4c9aa86a

PR #1951, opened 2021-08-06
links en-US to pt-BR

Larissa-Fernandes:
It would be helpful if links direct the reader to the page in the same language they are reading. So, as he is reading in Portuguese, if he clicks on a link, he will be taken to the article in Portuguese.


josielrocha (approved):
👍🏽 
Thank you @Larissa-Fernandes!

PR #1959, opened 2021-08-07
[fr] Corr link in DOM_Property_Viewer help page

tbertels:
#Settings -> #options


SphinxKnight (approved):
Thanks for the fix @tbertels! I've fixed the first link

PR #1961, opened 2021-08-07
Fix broken live examples in `zh-cn/learn/css`

dukecat0:
Fix broken live examples in `zh-cn/learn/css`.


t7yang (changes requested):
1. Do not change `id` or modify hidden content (or something else) which do not match what you declare — Fix broken live examples in `zh-cn/learn/css`:
  - All `id` will be remove after convert from html to markdown format.
  - Hidden contents will remove in the future.
2. Please make id of `EmbedLiveSample` same as the latest English version.

meowmeowmeowcat:
@t7yang However, will live samples work if we don't change `id` or modify hidden content?

meowmeowmeowcat:
I'll change `id` so id of `EmbedLiveSample` will be the same as the latest English version.

files/zh-cn/learn/css/css_layout/positioning/index.html
```diff
@@ -539,8 +541,9 @@ <h6 id="Sticky_positioning_example">Sticky positioning example</h6>
 }
 </pre>
 
+<div id="Sticky_2">
 <div class="hidden">
-<pre class="notranslate">body {
+<pre class="brush:css notranslate">body {
```
meowmeowmeowcat, line 546: I've added `brush:css` here. If this is missing, then the live sample won't work.

files/zh-cn/learn/css/css_layout/positioning/index.html
```diff
@@ -557,7 +560,7 @@ <h6 id="Sticky_positioning_example">Sticky positioning example</h6>
 }
 </pre>
 
-<pre class="notranslate">&lt;h1&gt;Sticky positioning&lt;/h1&gt;
+<pre class="brush:html notranslate">&lt;h1&gt;Sticky positioning&lt;/h1&gt;
```
meowmeowmeowcat, line 563: I've added `brush:html` here. If this is missing, then the live sample won't work.

meowmeowmeowcat:
> Do not change id or modify hidden content (or something else) which do not match what you declare — Fix broken live examples in zh-cn/learn/css

I've changed some `id` to match with `EmbedLiveSample`. (And also follow the latest English version)

files/zh-cn/learn/css/css_layout/grids/index.html
```diff
@@ -83,6 +83,7 @@ <h3 id="定义一个网格">定义一个网格</h3>
 
 <p>在规则里加入你的第二个声明。刷新页面后，你会看到子项们排进了新定义的网格中。</p>
 
+<div id="Grid_1">
```
t7yang, line 86: Please make sure for every added `<div id="Grid_*"` has it's own close tag.
This article a bit difference with others cause it do not wrap code in a `pre` tag instead a hidden `div`.

t7yang (changes requested):

files/zh-cn/learn/css/css_layout/grids/index.html
meowmeowmeowcat: You're right. Thanks for catching this!

meowmeowmeowcat:
All live samples should work properly now.

t7yang (approved):

PR #2011, opened 2021-08-10
Update MDN/Structures/Macros and MDN/Yari of l10n-zh

pocketdr:


t7yang (changes requested):
The added files should under `zh-*/mdn/tools/kumascript` instead of `zh-*/tools/kumascript`.

PR #2014, opened 2021-08-10
General rewording of the French translation of the `audio_and_video_delivery` guide

audrasjb:
This PR contains a general rewording of the French translation of the `audio_and_video_delivery` guide.

This is still a work in progress.

### DIFF
--- a/files/fr/web/guide/audio_and_video_delivery/index.html
+++ b/files/fr/web/guide/audio_and_video_delivery/index.html
@@ -350,29 +350,29 @@ AddType video/webm webmv</pre>
 <h3 id="Audio_uniquement">Audio uniquement</h3>
 
 <ul>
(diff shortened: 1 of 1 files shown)


audrasjb:
I think that one is good to be reviewed @SphinxKnight. That was a big one, so there's probably room for many enhancements, but I believe this proposal is a LOT better than the current translation :)

SphinxKnight:
Let's check this :)

SphinxKnight:
I've pushed a bunch of changes. Most importantly:
* Titles with words starting in uppercase > they should be lowercased (apart from the first word)
* Added / deleted parts to match the en-US current layout (except when those mentioned FxOS or were somehow irrelevant)

SphinxKnight (approved):

PR #2044, opened 2021-08-12
Update HTML tags and attributes translated to Portuguese

kayquer:
Several HTML attributes had been translated to Portuguese like, name and value. JS events for click where translated too.


josielrocha:
Thank you @kayquer

PR #2097, opened 2021-08-16
Unable to find a Prism grammar for 'example-bad'

peterbe:


mfuji09 (approved):
I approve of Japanese one.

PR #2154, opened 2021-08-20
Fix inadequate <mrow><mo>{</mo><mi>C</mi><mo>}</mo></mrow> and add +- sign where necessary.

borispinatel:
Suppress inadequate <mrow><mo>{</mo><mi>C</mi><mo>}</mo></mrow> and add +- sign where necessary.


SphinxKnight (approved):
Merci beaucoup @borispinatel !

PR #2153, opened 2021-08-20
Removes potential ambiguity on this MathML exemple.

borispinatel:
Removes potential ambiguity.

PR #2159, opened 2021-08-21
escape title with colon

escattone:
Adds back the quotes around the title that includes a `:` character. Reverts part of #2153 that broke the production build (cc @borispinatel).


escattone:
Sorry to bypass your review @audrasjb, but I want to fix the production build.

SphinxKnight:
Thanks @escattone ! no need for apologies

PR #2160, opened 2021-08-21
Fix typo

koddr:
Wrong form of the Russian word `все-так`. Right form is `все-таки` (with the ended `и` letter).


mpstv (approved):
Спасибо

PR #2162, opened 2021-08-21
FIX | Corrected a small typo in the Russian translation

Pobepto:
Corrected a typo in the word `броузер` in the Russian translation, the correct word should be `браузер`.


saionaro (approved):

Saionaro:
@Pobepto thank you for the fix!

PR #2179, opened 2021-08-21
Fix an external link on using_the_aria-invalid_attribute/index.html in JP

kenkoooo:
Since the external link to the working example is outdated, users who click on the link will be sent to a dangerous site. I fixed it to use an example of W3 as the US page.


mfuji09 (approved):
修正ありがとうございます。

PR #2186, opened 2021-08-22
add ARIA list and listitem role for zh-CN

mutoe:


irvin:
@mutoe I suggestion that to focus on one article per PR.

PR #2229, opened 2021-08-25
Remove note from replaceAll() pt-br docs

lebreRafael:
This note can make some people think that replaceAll is not supported by Chrome, which is not true.

This note is not present on the English version, so I think it is fine to remove.


josielrocha:
🎉 

Obrigado pela contribuição @lebreRafael!

PR #2232, opened 2021-08-26
fr-glossary/spa id: 514

ManuSquall:
suggest translation for [id:514 /glossary/first_cpu_idle](https://tristantheb.github.io/history-content/#514)


files/fr/glossary/spa/index.html
```diff
@@ -0,0 +1,38 @@
...
+
+<p>Certains des frameworks SPA les plus populaires incluent:</p>
+
+<ul>
+ <li><a href="https://reactjs.org/">React</a></li>
+ <li><a href="https://angular.io/">Angular</a></li>
+ <li><a href="https://vuejs.org/">Vue.JS</a></li>
+</ul>
+
+<section id="quick_links">
```
SphinxKnight: This ID should not be lowercased for the content to appear in the sidebar.

SphinxKnight (commented):
I've pushed a commit to fix typos:
* Web should not be uppercased except when using it as a noun (see https://github.com/mozfr/besogne/wiki/Guide-stylistique-pour-la-traduction#divers)
* Locales don't start with an uppercase letter in French.

files/fr/glossary/spa/index.html
ManuSquall: ho sorry about that, thanks for the review

SphinxKnight: no need for apologies this is quite a specific thing :)

PR #2268, opened 2021-08-29
fix typo on Animation performance and frame rate article

am0xff:


Saionaro:
@am0xff thanks for the fix!

PR #2271, opened 2021-08-29
Revamps WS landing page - translates WS interface subpages - fixes #2139

### LINKED ISSUE #2139: url MessageEvent KO
Documentation URI https://developer.mozilla.org/en-US/WebSockets/WebSockets_reference/MessageEvent is KO

Referenced here https://developer.mozilla.org/fr/docs/Web/API/WebSocket

SphinxKnight:
#2139 mentioned a source page which was obsolete (vs en-US) and a target page which was not translated. I've translated the whole interface and revamped the two existing pages to take care of this issue.


audrasjb (approved):
Looks good to me 👍

PR #2280, opened 2021-08-30
Update the output.

CC2m2:


t7yang (approved):

t7yang (changes requested):
> Pull request title can't just be "Update index.html"

Please change title according CI error message.

PR #2301, opened 2021-09-01
Fix content property description

pepelsbey:
It’s called attribute which is not correct and may cause (well, already caused) problems. Since it’s located in the Attributes section, it sounds like it’s `<template content>`, but it’s `HTMLTemplateElement.content`.


lex111 (approved):

lex111:
Great, thanks!

sashasushko:
@lex111 поторопился чуток :-)

sashasushko:
Оформлю ревью отдельным ПРом, тогда.

PR #2318, opened 2021-09-03
add code to make it more elegant

KevinZonda:


mfuji09 (changes requested):
It shouldn't be changed because `<code>`s in `<pre>` are reported as flaws by the Yari platform.
And this change is meaningless because the original version (in English) has been already convert to Markdown.

KevinZonda:
> It shouldn't be changed because `<code>`s in `<pre>` are reported as flaws by the Yari platform.
> And this change is meaningless because the original version (in English) has been already convert to Markdown.

So 
image
JP edition

image
EN edition

image
CN edition

could u tell me why Japanese one looks so different?

mfuji09:
@KevinZonda 
It is because the English version has been converted to Markdown and uses the syntax highlight of `js`. Japanese one hasn't been converted yet and isn't used a class `brush: js` in the `<pre>` element.
So the code in Japanese one isn't decorated and it is shown in normal monospace font. Perhaps the default monospace font of Japanese in **your** browser is MS Gothic. It is a very old font (created about 30 years ago) and it uses bitmap fonts for smaller characters for compatibility reason. So it looks ugly.

This problem should be corrected (to be shown as well as English version) by being converted to Markdown, or using a class `brush: js`.
Please note that a problem of presentation mustn't be fixed by modifying HTML markup because it should be fixed by CSS.

KevinZonda:
Cool, thx :)

PR #2361, opened 2021-09-06
fix: change 『一些列』 to 『一系列』

cqA7:
fix:  change "一些列" to "一系列"


t7yang (approved):

t7yang (changes requested):
Please change your PR title to describe what you do, never use "Update index.html" as PR title.

PR #2362, opened 2021-09-06
update link attributes translate for zh-CN

2heal1:
Attributes as  is only used when rel="preload" or rel="prefetch" has been set on the <link> element.  But the translation missed 'prefetch' .


t7yang (changes requested):
Change your PR title instead of just "Update index.html" which may cause CI error.

PR #2378, opened 2021-09-07
Fix 6种原始类型 to 7种原始类型

Sean2096:
Fix `6种原始类型` to `7种原始类型`


t7yang (changes requested):
Please change your PR title, instead of just "Update index.html" which may cause CI error.

PR #2555, opened 2021-09-24
sync one more paragraph from en doc

ZivyHu:
translate more information from English to Simplied-Chinese


t7yang (changes requested):
Change your PR title

PR #2588, opened 2021-09-27
Web API section to Markdown - French

SphinxKnight:
This is a followup of https://github.com/mdn/translated-content/pull/2464

The conversion report is available [here](https://gist.github.com/SphinxKnight/cefa8bfe7ec332542d86aba29d4a054e)

PR #2597, opened 2021-09-27
Fix web api fr upstream h2m

SphinxKnight:
This PR intends to fix issues which were detected via the CI after the conversion (see #2588)


SphinxKnight:
I'm shamelessly merging this one to advance #2588

PR #2660, opened 2021-10-05
Small translation fix

Meleipe:
Small translation fix. It should state that it produces some other positive value (and not a negative one).


josielrocha (approved):

josielrocha:
Obrigado pela contribuição @Meleipe!!

PR #2699, opened 2021-10-10
Fix, firebug in past and finished translation

poum:
Fix some misspelling, improve some sentences, Firebug related sentences in past for more clarity and add missing translation


audrasjb (approved):
Thanks for the work @poum :)

PR #2724, opened 2021-10-13
Fix typo

mosade:
"sLeft" to "scrollLeft"


irvin:
此處並沒有錯誤，請參考原文 https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft

PR #2732, opened 2021-10-14
Fix typo in code sample for JS Ref typeof page

yann2084:
Hello,
in console :
var maDonnee = 99;
typeof maDonnee + 'Toto';

prints 'numberToto' // without space


SphinxKnight (approved):
Thanks a lot @yann2084!
I pushed another commit to update the rest of the page to match the English reference.

PR #2801, opened 2021-10-21
fix ajax send formdata incorrect format

btea:
fix #2704


irvin (changes requested):
Please delete those two lines to make them identical with the Engish version
https://github.com/mdn/content/blob/main/files/en-us/web/api/xmlhttprequest/send/index.md#example-get

PR #2825, opened 2021-10-22
[ja] Changed hgroup section from deprecated to warning

grgr-dkrk:
The Japanese version of the `hgroup` page is deprecated, but in the original this part was changed to a warning at https://github.com/mdn/content/pull/4977.
This PR is a Japanese translation of it.


potappo (approved):
OK です。

PR #2832, opened 2021-10-22
I translated "The" into Portuguese

ghost:
I changed the position of the word "elemento" as it makes more sense in Portuguese


dukex (approved):
Thanks

josielrocha (approved):

PR #2837, opened 2021-10-23
[ko] Translation done of 'AudioWorklet' document

logic-finder:
https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet
위 문서를 translated-content폴더에 추가 및 번역하였습니다.

I added [this document](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet) to translated-content/ko folder and translated it into Korean.


yujo11 (approved):
고생하셨습니다! 👍 👍👍

PR #2935, opened 2021-11-01
Fix wrong link for offscreencanvas spec

nanokind:


SphinxKnight (changes requested):
Hello @fengzilong thanks a lot for the fix (especially across different locales :) ).
I think it would be best for the localized docs to follow the en-US version as per https://github.com/mdn/content/blob/main/files/en-us/web/api/offscreencanvas/index.md?plain=1#L116-L119. 
The {{Specifications}} call should rely on the `browser-compat key in the frontmatter.
Let me know if you need help adapting this.

fengzilong:
Very usefull tip. Files are updated, but I'm not sure whether it matches what you describle or not, please help me review the changes @SphinxKnight

SphinxKnight (approved):
@fengzilong Thanks a lot for this. It's good to me. I'm waiting a bit for other maintainers. We shall merge soon I think.

pje1740 (approved):
I have checked the Korean document and seems fine to me! Thank you :)

PR #2953, opened 2021-11-03
Fixed typos and errors in Classes reference code example.

jeronimoek:
Fixed typos and errors in expected output in the Classes reference's code examples.

### DIFF
--- a/files/es/web/javascript/reference/classes/index.html
+++ b/files/es/web/javascript/reference/classes/index.html
@@ -218,7 +218,10 @@ p.hablar();</pre>
 
 <pre class="notranslate"><code>var Animal = {
   hablar() {
-    console.log(this.nombre + 'hace ruido.');
+    console.log(this.nombre + ' hace ruido.');
+  },
+  comer() {
+    console.log(this.nombre + ' se alimenta.');
   }
 };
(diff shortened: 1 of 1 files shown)


JuanVqz:
it seems like we won't translate code to Spanish, so... I think this could be invalid. @Graywolf9, @sergiogval, @lasr21 somebody can confirm, please?

sergiogval (commented):
Yo no le veo problema al traducir al contenido que no es código en sí. Sólo hay qu tener cuidado de no traducir cosas como atributos, métodos, palabras reservadas, entre otras cosas que son parte del código. 
Dicho eso, @JuanVqz recuerdo que había que remover las clases "notranslate", es correcto? con ese aclarado yo daría mi aprobación.

I far as I understand, what it was agreed is that we will make sure that the code itself hadn't a translated word  such attributes, methods, reserved keywords etc. Those may mislead the reader. That said I remember that you ran a script to remove the "notranslate" attribute but this code still contains it. That is my only observation, everything else is ok. What do you think?

Graywolf9:
It seems that the code was in Spanish before the PR, we could accept it and make a standard rule for further situations

PR #2966, opened 2021-11-04
Added an example

SnowingFox:
A code example is added to the warning content at the beginning of the article. 

This example can clearly show its meaning.

 Why should I give this example? Because when I first read it, I found it difficult to imagine and understand this concept.

please accept


SnowingFox:
error code, i will fix it right now

PR #2986, opened 2021-11-06
fix: [FR] typo on Models page

cedlerouge:
Correct some mistakes like plurial


SphinxKnight:
Note for self: there will be a conflict with #2738

SphinxKnight (approved):
Merci @cedlerouge ! :)

PR #2990, opened 2021-11-06
[pt-BR] Anchors to /Global_attributes now point to pt-BR version

luizffgv:
Only when suitable, i.e. in sections written in Portuguese, and for anchors to pages available in Portuguese.

No unrelated changes, but some minimal fixes have been applied to the same line in some cases.

*Many* other anchors still point to en-US pages, but this PR is a step forward.


josielrocha (approved):

josielrocha:
Obrigado pela contribuição @luizffgv!!!

PR #2994, opened 2021-11-07
fix: [ja,ko,es,ru] update codes on Node.js server without a framework

yuto0214w:
1. fixed 404 process code [all]
`response.writeHead(200, { 'Content-Type': contentType });` -> `response.writeHead(404, { 'Content-Type': 'text/html' });`
2. fixed `var contentType = 'text/html';` [es,ko,ru]
3. fixed code on korean page doen't show correctly [ko]
`<pre>` -> `<pre class="brush: js">`

### DIFF
--- a/files/es/learn/server-side/node_server_without_framework/index.html
+++ b/files/es/learn/server-side/node_server_without_framework/index.html
@@ -34,27 +34,28 @@ http.createServer(function (request, response) {
 
     var extname = String(path.extname(filePath)).toLowerCase();
     var mimeTypes = {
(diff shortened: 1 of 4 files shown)


mfuji09 (changes requested):
There are still codes different from en-US version in the ja version.
(ex. line 45-59, line 73, etc.)
Please reflect the whole changes in en-US version.

mfuji09 (approved):
I approve about ja file.

mpstv (approved):

yujo11 (approved):

Issue #2842, opened 2021-10-24
All French pages with an EmbedLiveSample error


SphinxKnight:
After a local crawl of the docset (as of today), it seems all of the following pages have an EmbedLiveSample/MacroLiveSampleError.

Filing this as an issue to help fix this all at once rather than from reports from reader:

- https://developer.mozilla.org/fr/docs/mozilla/add-ons/webextensions/user_interface/browser_styles
- https://developer.mozilla.org/fr/docs/orphaned/web/css/linear-gradient%28%29
- https://developer.mozilla.org/fr/docs/orphaned/web/css/@media/prefers-reduced-transparency
- https://developer.mozilla.org/fr/docs/orphaned/web/api/body
- https://developer.mozilla.org/fr/docs/orphaned/web/api/detecting_device_orientation
- https://developer.mozilla.org/fr/docs/conflicting/learn/html/introduction_to_html/advanced_text_formatting_b235e00aa38ee1d4b535fc921395f446
- https://developer.mozilla.org/fr/docs/conflicting/learn/css/building_blocks/selectors
- https://developer.mozilla.org/fr/docs/conflicting/learn/css/css_layout/introduction
- https://developer.mozilla.org/fr/docs/webassembly/understanding_the_text_format
- https://developer.mozilla.org/fr/docs/glossary/css_selector
- https://developer.mozilla.org/fr/docs/web/html/element/menu
- https://developer.mozilla.org/fr/docs/web/html/element/style
- https://developer.mozilla.org/fr/docs/web/html/element/ul
- https://developer.mozilla.org/fr/docs/web/html/element/b
- https://developer.mozilla.org/fr/docs/web/html/element/li
- https://developer.mozilla.org/fr/docs/web/html/element/input/email
- https://developer.mozilla.org/fr/docs/web/html/element/input/password
- https://developer.mozilla.org/fr/docs/web/html/element/textarea
- https://developer.mozilla.org/fr/docs/web/html/attributes/pattern
- https://developer.mozilla.org/fr/docs/web/css/css_grid_layout/relationship_of_grid_layout
- https://developer.mozilla.org/fr/docs/web/css/css_grid_layout/box_alignment_in_css_grid_layout
- https://developer.mozilla.org/fr/docs/web/css/css_grid_layout/layout_using_named_grid_lines
- https://developer.mozilla.org/fr/docs/web/css/css_grid_layout/css_grid_logical_values_and_writing_modes
[...]

PR #2995, opened 2021-11-07
Fix live sample errors for HTML docs

SphinxKnight:
this is a followup for #2842 after HTML md conversion


SphinxKnight:
There is no build error and I hope this is now done for EmbedLiveSample errors.
Merging myself like there is no tomorrow.

PR #3072, opened 2021-11-11
fix typo

astroxii:
### Reason:

The expression "**a seguir**" is written with "**a**" not "**à**". Small change but makes it correct.


josielrocha:
Thank you for your contribution @astroxii!
👍🏽

PR #3188, opened 2021-11-23
fix translation of "sorting" and translate remaing

aopenna:
fix translation of "sorting" and translate the remaing of text not yet translate (from "Conditionals" to the bottom of page.


aopenna:
Translation to pt-BR

josielrocha (approved):

PR #3216, opened 2021-11-28
Fix: Ordered List Wrong Attribute

njesenberger:
`<ol start="i">` should be `<ol type="i">` in order to display Roman numeral type


audrasjb (approved):
LGTM, thank you!

PR #3251, opened 2021-12-03
[ko] translate object __proto__ page

roy-jung:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto
위 문서를 translated-content/ko 폴더로 옮기고 번역하였습니다.

I translated [this document](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto) into Korean.


yujo11 (approved):
고생하셨습니다! 👍 👍 👍

PR #3322, opened 2021-12-12
new translation for ArrayBufferView

yin1999:
Fixes the flow in pr: #3321.

### DIFF
--- a/files/zh-cn/web/api/arraybufferview/index.md
+++ b/files/zh-cn/web/api/arraybufferview/index.md
@@ -1,0 +1,33 @@
+---
+title: ArrayBufferView
+slug: Web/API/ArrayBufferView
+tags:
+  - API
+  - Interface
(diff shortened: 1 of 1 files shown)


michaelstephenson0613 (approved):

t7yang (approved):

SphinxKnight:
@t7yang do you have any idea why the check-redirects workflow was not applied here?

t7yang:
> @t7yang do you have any idea why the check-redirects workflow was not applied here? This check failing caused other PR's CI to fail as well :/

🤔 I have no idea too, but CI works in #3327.

SphinxKnight:
>  I have no idea too, but CI works in #3327.

It seems the redirects need to be updated, I'm currently seeing it fail: https://github.com/mdn/translated-content/runs/4514978574?check_suite_focus=true

PR #3350, opened 2021-12-13
Remove unnecessary redirect (which is preventing other PRs)

mfuji09:
This redirect became unnecessary by #3322 and it seems to cause a error for new PRs other than zh-CN locales.


SphinxKnight:
Thanks @mfuji09!

PR #3361, opened 2021-12-14
main/docs/README.md: Remove duplicate tags

PythonCoderAS:
These tags do not need to be separate since they render the same style.


SphinxKnight (approved):
:+1: for me, thanks @PythonCoderAS

PR #3388, opened 2021-12-17
Remove `shortcut` from the example for favicon

shqld:
As described in MDN itself, web authors must no longer use `shortcut` as the link type.

> Warning: The shortcut link type is often seen before icon, but this link type is non-conforming, ignored and web authors must not use it anymore.

https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types


potappo (approved):
修正内容 OK です。ありがとうございます。

PR #3404, opened 2021-12-19
fix: french typo in documentation

xandermann:
This fixes some typo on the page https://developer.mozilla.org/fr/docs/Web/HTTP/Redirections.


SphinxKnight (approved):
Merci beaucoup @xandermann : )

PR #3490, opened 2021-12-30
Add autocorrect-lint GitHub Action

huacnlee:
1. Add a GitHub Lint Action by use AutoCorrect for check copywriting.
2. A `.autocorrectignore` file to configure that which files wants to use AutoCorrect lint.
3. #3834 
4. #3835

[AutoCorrect](https://github.com/huacnlee/autocorrect) also supports for Korean and Japanese, is not enabled in current, I just updated the Chinese as a example for first PR review.

## Current issues

https://developer.mozilla.org/zh-CN/docs/Learn/HTML

- Red: Incorrect
- Green: Correct

截屏2021-12-30 11 55 12

### DIFF
--- a/.autocorrectignore
+++ b/.autocorrectignore
@@ -1,0 +1,3 @@
+**/*
+!files/zh-cn/web/html/**/*
+
--- a/.github/workflows/autocorrect-lint.yml
+++ b/.github/workflows/autocorrect-lint.yml
@@ -1,0 +1,16 @@
+name: AutoCorrect Lint
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: AutoCorrect
+        uses: huacnlee/autocorrect-action@v1.5.5


huacnlee:
截屏2021-12-30 11 57 47

Lint for make sure the next changes will follow the standard.

irvin (changes requested):
目前 zh-tw 的格式規範不在中英文間加入空格，請將相關檔案先移除。

此外在確認本機制可上限前，請先選取十個檔案為範例，讓這個 PR 更精簡，便於 MDN 負責工程師檢閱。
待本 PR 已確認並 merge 後，再另開 PR 將 zh-cn 下的所有檔案進行批次更新。

huacnlee:
@irvin Hi, I just updated for only change `files/zh-cn/web/html/element` directory for as an example.

And add a AutoCorrect lint configuration for only check this directory.

After this PR has merge, I will continue write others PRs for modify the lint config.

irvin:
@Rumyra can you review (or suggest someone) if we should add this kind of linter as a github actions to do the automatic format correction on CJK docs?

irvin:
@huacnlee  建議你先把你本機跑過的修正成果一部分一部分的開成 PR，我們先讓修正上線，自動檢驗機制再等工程師慢慢研議。

huacnlee:
https://github.com/mdn/translated-content/pull/3834

.github/workflows/autocorrect-lint.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: AutoCorrect
+        uses: huacnlee/autocorrect-action@main
```
schalkneethling, line 16: We tend to avoid using `main` and instead opt for a version or tag such as `v1.5.5`. I believe you can change the above to:

```suggestion
        uses: huacnlee/autocorrect-action@v1.5.5
```

.autocorrectignore
```diff
@@ -0,0 +1,2 @@
+**/*
+!files/zh-cn/web/html/element/**/*
```
schalkneethling, line 2: Looks like this file is missing a empty line at the end of the file:

```suggestion
!files/zh-cn/web/html/element/**/*

```

schalkneethling (changes requested):

huacnlee:
Just updated.

This PR will breaking CI, before we merge #3834, #3835

.github/workflows/autocorrect-lint.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: AutoCorrect
+        uses: huacnlee/autocorrect-action@v1.5.5
```
huacnlee, line 16: image

`huacnlee/autocorrect-action@v1.5.5` version it works.

schalkneethling (approved):
This looks good to me now. Thank you, @huacnlee. Merge when the dependency pull requests are merged.

schalkneethling:
I see @irvin also requested changes. Not sure if those have been addressed.

irvin:
@schalkneethling I'm suggested that we can first commit the locally fixing result before the mechanism been reviewed. So feel free to merge this one if you are ready.

irvin (approved):

SphinxKnight:
@huacnlee @schalkneethling @irvin I'm seeing failures for PRs' CI with this new workflow. Could you let me know if something needs changing in the PRs' branches or if it is somehow related to this?

SphinxKnight:
I think it would be best to include `paths` targeting in the workflow itself rather than only in the autocorrectignore file.

SphinxKnight:
For the record, I decided to revert this PR with https://github.com/mdn/translated-content/pull/4173

huacnlee:
> I think it would be best to include `paths` targeting in the workflow itself rather than only in the autocorrectignore file. Another (quite elegant and complementary) solution that would work for zh-CN current content (with potential linting errors) wuold be to use the same mechanism as in `pr-test` which uses `technote-space/get-diff-action@v4.2` to provide only the _changed_ files to subsequent actions.
> 
> One could then "feed" autocorrect with the exact list of files to check (instead of the whole tree)
> 
> Meaning this workflow would help improve "newly edited" content but not block because of existing errors on files which were not edited by the PR which triggered the workflow: https://github.com/mdn/translated-content/blob/main/.github/workflows/pr-test.yml#L70-L124

Okey, I will try to fix that and make a new Pull Request.

schalkneethling:
> For the record, I decided to revert this PR with #4173

Sorry about that @SphinxKnight. Irvin and the contributor seemed to feel good about it and I really did not have enough context to make a call. It is also so dang tricky to test these workflows until they are merged. I should look into getting https://github.com/nektos/act going.

irvin:
@SphinxKnight for any further PR relative to the CI, should we request a review from you or any other from the team, to ensure it works fine?

schalkneethling:
> @SphinxKnight for any further PR relative to the CI, should we request a review from you or any other from the team, to ensure it works fine?

That might be OK, but it really is up to @SphinxKnight to decide how much time he has for this. I reckon we need to ensure we use https://github.com/nektos/act to test workflow changes.

schalkneethling:
Btw. @irvin I was not passing blame to anyone. This is just a hard part of the ecosystem to test properly. 😄

irvin:
we could have some reviewer team like @mdn-core or @mdn-system for any PR that involved not only translate can request from.

schalkneethling:
> we could have some reviewer team like @mdn-core or @mdn-system for any PR that involved not only translate can request from.

Yup, there is https://github.com/orgs/mdn/teams/core

SphinxKnight:
> @SphinxKnight for any further PR relative to the CI, should we request a review from you or any other from the team, to ensure it works fine?

I'd be glad too, just to add another pair of eyes to such PRs! (if time allows of course, I don't need to be a blocker/single source of failure here)
And to be honest, I may have missed the issue on this one as well :) as @schalkneethling testing those is an issue of its own (and I definitely need to read / practice regarding GH actions)

irvin:
> > @SphinxKnight for any further PR relative to the CI, should we request a review from you or any other from the team, to ensure it works fine?
> 
> I'd be glad too, just to add another pair of eyes to such PRs! (if time allows of course, I don't need to be a blocker/single source of failure here) And to be honest, I may have missed the issue on this one as well :) as @schalkneethling testing those is an issue of its own (and I definitely need to read / practice regarding GH actions)

we should probably add this as a guide somewhere, guess other locale teams would have the same questions when they got similar PRs. I had searched around before but did not find the instructions.

Rumyra:
> Ideally, I'd think of one team where each member would be notified for each PR touching core elements of mdn/translated-content (incl. GH Actions Workflows) and where one member could give their go. There is definitely a need for documentation but I think automating this a bit would be helpful (and would reduce the risk of error/missing the docs)
> 
> Regarding a potential team :
> 
> * @mdn /core, its members are Mozilla employees and is not really suitable here as far as I can tell
> * There might be some other "l10n veterans" (such as @mfuji09) who might be interested in being consulted for such changes on mdn/translated-content so it may be a specific team for this repo/l10n

@SphinxKnight I'd be on board with this. Do you want to open a separate issue/discussion? I can create a team and I guess CODEOWNERS would need updating. But I think a core translated content team that gets pinged on any prs which touch files that don't automatically get assigned to specific teams would be good. We do this on content so I don't see any readon not to do it here as well 👍

SphinxKnight:
Thanks @Rumyra, I'll open another issue for this in the next hours/days.

Issue #3497, opened 2021-12-30
Update needed for "Apprendre : Formulaires"


cw118:
The French "Learn/Forms" module would benefit from an update, starting with the "Les widgets de formulaire natifs" page (https://developer.mozilla.org/fr/docs/Learn/Forms/Basic_native_form_controls).

<details>
<summary>Doing so would also fix/"remove" the following issues (which are difficult to correct without updating most/all page content):</summary>

- None of the images on the page are available/displaying (the en-US version no longer uses them)
- Currently, in the table starting at [line 299](https://github.com/mdn/translated-content/blob/main/files/fr/learn/forms/basic_native_form_controls/index.md?plain=1#L299), there are two external image image links:
  - `https://developer.mozilla.org/files/4583/datalist-safari.png`
  - `https://developer.mozilla.org/files/4581/datalist-firefox-macos.png`
- Various sections do not match the en-US version (have been deleted from en-US/are missing in the French version)
</details>

Parts of "Les widgets de formulaire natifs" were moved and became their own pages, including (there may be more) "HTML5 input types" and "Other widgets", while some new pages were added in en-US. Links and `{{PreviousMenuNext}}` macros will likely need to be modified/updated across all files in the "Apprendre : Formulaires" module as well.

Tasks:

- [ ] Update ["Les widgets de formulaire natifs"](https://developer.mozilla.org/fr/docs/Learn/Forms/Basic_native_form_controls)
- [ ] Create fr ["HTML5 input types"](https://developer.mozilla.org/en-US/docs/Learn/Forms/HTML5_input_types) from various removed sections of "Les widgets de formulaire natifs"
- [ ] Create fr ["Other form controls"](https://developer.mozilla.org/en-US/docs/Learn/Forms/Other_form_controls) from various removed sections of "Les widgets de formulaire natifs"
- [ ] Translate ["UI pseudo-classes"](https://developer.mozilla.org/en-US/docs/Learn/Forms/UI_pseudo-classes) for fr
- [ ] Translate ["Test your skills: Basic controls"](https://developer.mozilla.org/en-US/docs/Learn/Forms/Test_your_skills:_Basic_controls) for fr
- [ ] "Re-link"/re-order modules to match en-US (+update links from `Web/Guide` to `Learn`)

Issue #3496, opened 2021-12-30
Update needed for "Composition avancée des formulaires HTML"


cw118:
## What page(s) did you find the problem on?

The fr "Composition avancée des formulaires HTML" page: https://developer.mozilla.org/fr/docs/Learn/Forms/Advanced_form_styling

## Specific page section or heading?

"Un exemple un peu plus compliqué" in [Cases à cocher et boutons radio](https://developer.mozilla.org/fr/docs/Learn/Forms/Advanced_form_styling#cases_à_cocher_et_boutons_radio) is where the issue concerning an image file is, though it seems the entire page would benefit from an update (vs. the en-US counterpart).

## What is the problem?

- The image `checkbox-sprite.png` is unavailable
- The same image is provided as the `background: url` source as an external link, `https://developer.mozilla.org/files/4173/checkbox-sprite.png`
- The "Un exemple un peu plus compliqué" section should be updated to match the en-US "Using appearance: none on radios/checkboxes" (subsection of [Styling checkboxes and radio buttons](https://developer.mozilla.org/en-US/docs/Learn/Forms/Advanced_form_styling#styling_checkboxes_and_radio_buttons)). 
  - **Doing so would fix both of the problems mentioned directly above, and from what I've seen, the entire page should probably be updated to match the en-US version.**

PR #3498, opened 2021-12-30
Remove external image sources on French pages pt. 1

cw118:
Removing external image sources in French files (e.g. `https://mdn.mozillademos.org/files/13090/colorful-heart.png`). Part 1 of a few more PRs to come, though I'm not sure if these edits are necessary for the "conflicting" and "orphaned" folders (I don't yet know what they're for)?

For `mdn.mozillademos.org/files` links, there are 651 results in 167 files so far in all French folders (I ended up creating #3566 to potentially track this).
 
This PR should fix all occurrences in the Learn section. I also tried to fix `EmbedLiveSample`s and made various small edits since there weren't too many files concerned, but I'm not sure what the scope of this PR should be (would adding `&nbsp;`s be a good idea?). Most of the pages included would benefit from an update, I mostly just did what I could to get the images (and links) working — I hope that's alright, but if not, what would be a better solution?

There are also 2 files with `developer.mozilla.org/files` image links, but from what I've seen of the 2 pages, updates are needed to match the en-US versions — doing so would then fix these image links *(it's quite difficult to fix them without modifying entire sections of the files)*. I've opened #3496 and #3497 for these.


SphinxKnight (approved):
Merci beaucoup, y compris d'avoir traduit l'image avec le schéma ! :100:

PR #3528, opened 2022-01-02
[ko] Translation done of `AudioNode.disconnect()` document

logic-finder:
https://developer.mozilla.org/en-US/docs/Web/API/AudioNode/disconnect
위 문서를 translated-content/ko 폴더로 옮기고 번역하였습니다.

I translated [this document](https://developer.mozilla.org/en-US/docs/Web/API/AudioNode/disconnect) into Korean.


yechoi42 (approved):
감사합니다!

Issue #2846, opened 2021-10-24
All zh-CN pages with an EmbedLiveSample error


SphinxKnight:
After a local crawl of the docset (as of today), it seems all of the following pages have an EmbedLiveSample/MacroLiveSampleError.

Filing this as an issue to help fix this all at once rather than from reports from readers:

- https://developer.mozilla.org/zh-cn/docs/web/html/applying_color
- https://developer.mozilla.org/zh-cn/docs/web/html/global_attributes/itemscope
- https://developer.mozilla.org/zh-cn/docs/web/html/global_attributes/itemref
- https://developer.mozilla.org/zh-cn/docs/web/html/element/map
- https://developer.mozilla.org/zh-cn/docs/web/html/element/meter
- https://developer.mozilla.org/zh-cn/docs/web/html/element/blockquote
- https://developer.mozilla.org/zh-cn/docs/web/html/element/aside
- https://developer.mozilla.org/zh-cn/docs/web/html/element/menuitem
- https://developer.mozilla.org/zh-cn/docs/web/html/element/menu
- https://developer.mozilla.org/zh-cn/docs/web/html/element/address
- https://developer.mozilla.org/zh-cn/docs/web/html/element/figure
- https://developer.mozilla.org/zh-cn/docs/web/html/element/bdo
- https://developer.mozilla.org/zh-cn/docs/web/html/element/details
- https://developer.mozilla.org/zh-cn/docs/web/html/element/var
- https://developer.mozilla.org/zh-cn/docs/web/html/element/progress
- https://developer.mozilla.org/zh-cn/docs/web/html/element/style
- https://developer.mozilla.org/zh-cn/docs/web/html/element/u
- https://developer.mozilla.org/zh-cn/docs/web/html/element/dfn
- https://developer.mozilla.org/zh-cn/docs/web/html/element/cite
- https://developer.mozilla.org/zh-cn/docs/web/html/element/heading_elements
- https://developer.mozilla.org/zh-cn/docs/web/html/element/marquee
- https://developer.mozilla.org/zh-cn/docs/web/html/element/ol
- https://developer.mozilla.org/zh-cn/docs/web/html/element/pre
- https://developer.mozilla.org/zh-cn/docs/web/html/element/iframe
- https://developer.mozilla.org/zh-cn/docs/web/html/element/abbr
- https://developer.mozilla.org/zh-cn/docs/web/html/element/input
- https://developer.mozilla.org/zh-cn/docs/web/html/element/input/hidden
[...]

PR #3589, opened 2022-01-08
convert to markdown and add some translation for speechsynthesis.getvoices()

yin1999:
Changes:

- convert to markdown.
- add some translation.
- fix live sample.

Referenced Issues: #2846.


yin1999:
Migrate to #3630 to fix issue: #3618.

PR #3591, opened 2022-01-08
[fix] : entity references where shouldn't be used

mirumirumi:
I couldn't figure out what the clause starting with "#" in the brackets meant before the change, but it seemed unnecessary from the original English version.
Am I making the right decision?


mfuji09 (approved):
良いと思います。
なお、日本語版記事のレビュアーは日本人なので、プルリクエストは日本語で結構です。

PR #3635, opened 2022-01-11
Javascript reference error getter_only page in pt-br

nathipg:
Página atualizada de acordo com a content e traduzida


SphinxKnight:
Hello @nathipg this looks like this page should live under /pt-BR/ and not /fr/ :) 
Could you move it at the right place? Let us know if you need help with that (gently poking @mdn/yari-content-pt-br )

SphinxKnight (changes requested):
cf. earlier comment in the discussion

nathipg:
> Hello @nathipg this looks like this page should live under /pt-BR/ and not /fr/ :) Could you move it at the right place? Let us know if you need help with that (gently poking @mdn/yari-content-pt-br )

I'm so sorry, I mix things up. Thanks for letting me know

PR #3687, opened 2022-01-14
Minor update for fr X-Frame-Options and <frame>

cw118:


SphinxKnight (approved):
:100: Merci encore @cw118 !

PR #3756, opened 2022-01-21
Create README l10n-es

lasr21:
README for l10n-es MDN localization

### DIFF
--- a/files/es/README
+++ b/files/es/README
@@ -1,0 +1,1 @@
+Bievenido a la localización de MDN en español 


lasr21:
@Rumyra we are planning to add a Spanish README but we wonder that if that could broke something or is it ok?

Rumyra:
I think technically there should be an issue - I'll merge this and check again tomorrow :)

JuanVqz:
@Rumyra It seems like nothing was broken. isn't it?

Rumyra:
Yes it seems fine @JuanVqz :D

JuanVqz:
@Rumyra we just discover that the reason why the build hasn't broken was that the file hasn't extension it was just `README` instead of `README.md`.

[this](https://github.com/mdn/translated-content/runs/4953693637?check_suite_focus=true) is the failure when we changed it to `README.md`.

PR #3825, opened 2022-01-25
Vea también

josecarbajalbolbot:
Additional info about 403 in a website


JuanVqz:
Hi @josecarbajalbolbot thank you for your contribution but we cannot accept it because we are aligned with the English's documentation, also, we only add links to the official documentation or well-known webs like for example [https://css-tricks.com/](https://css-tricks.com/).

But we are happy to have you, this is the [matrix chat](https://chat.mozilla.org/#/room/#mdn-l10n-es:mozilla.org) if you want to discuss more about it.

PR #3850, opened 2022-01-27
Fix Markdown errors in French files

cw118:
Should fix most `** text**`, `**text **`, and single escaped asterisks `\*` (some escaped asterisks were kept since they seemed to be intentional)


SphinxKnight (approved):
Merci beaucoup @cw118, j'ai échantillonné la revue sur la moitié des pages et tout semble OK.

Issue #3851, opened 2022-01-27
Issue with 'async function'page syntax code error


kraccoon-dev:
## What page(s) did you find the problem on?
https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Statements/async_function

## Specific page section or heading?
Syntax

## What is the problem?
syntax코드에 문제가 있습니다.

<img src="https://i.ibb.co/yYYQCry/koerr.png" alt="koerr" border="0">

## What did you expect to see?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#syntax

<img src="https://i.ibb.co/VNpY0d4/en.png" alt="en" border="0">

PR #3903, opened 2022-01-30
fix: Issue with 'async function'page syntax code error

choinashil:
https://github.com/mdn/translated-content/issues/3851 텍스트 정렬 수정했습니다 :) 

AS-IS
Screen Shot 2022-01-30 at 3 44 29 PM

TO-BE
Screen Shot 2022-01-30 at 3 44 42 PM


yechoi42 (approved):
기여 감사합니다!

PR #3932, opened 2022-01-31
Fix Chinese CSS cursor image sources, Specs and Compat

cw118:
- Fixes image sources to just the file names (e.g. from `/@api/deki/files/3438/=default.gif` to `default.gif`)
- Updates/fixes specifications (use `{{Specifications}}` macro)
- Updates/fixes browser compat (moves `css.properties.cursor` into front-matter)

N.B. The `style="width: 7.5em;"` ensures the images have the "correct" width and height. Without/before this style:
zh-cursor-before

With/after this style:
zh-cursor-after

### DIFF
--- a/files/zh-cn/web/css/cursor/index.html
+++ b/files/zh-cn/web/css/cursor/index.html
@@ -9,14 +9,13 @@ tags:
   - NeedsTranslation
   - TopicStub
 translation_of: Web/CSS/cursor
+browser-compat: css.properties.cursor
 ---
 <div>{{CSSRef}}</div>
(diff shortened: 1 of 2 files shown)


cw118:
I originally thought it would be difficult for multiple review teams to check the files for their locales if all of them were combined into one PR, but it seems that the number of changes is small enough. 

Closing this one to combine the `zh`, `pt-BR` and `ko` locales (`ru` will remain as a standalone, it also happens to be the only one linked to an issue) and migrating to #3934.

PR #3972, opened 2022-02-03
Code  example : "fantome" instead of "shadow" everywhere [fr]

moinal:
"fantome" was used only once


SphinxKnight (approved):
Merci @moinal !

PR #4055, opened 2022-02-08
RU: Storage.key add null return case and specs

Zulcom:
Pack of small corrections for `Storage.get` method:
- Transferred "The order of keys is user-agent defined, so you should not rely on it." phrase from en version;
- added "key not found" case;
- copy-pasted "To see this used within a real-world example, see our Web Storage Demo" translation from getItem article to use more human-like phrase;
- replaced broken "standards" section with "specifications" sections (not sure that works how I suppose - please comment if I did it incorrectly);


lex111 (approved):
Преобразовал в маркдаун, должно быть ок

lex111 (approved):

lex111:
@Zulcom спасибо!

Issue #3886, opened 2022-01-29
 {{page}} macro removal for French pages - Misc.


SphinxKnight:
Fragment of #3881 in order to remove `{{page}}` calls from pages.
There are 10 pages (after tools removal) in various sections:

- [x] glossary/denial_of_service/index.md (dealt with #4058)
- [x] glossary/second-level_domain/index.md (dealt with #4058)
- [x] ~learn/server-side/django/generic_views/index.md~ false positive with Django template in the content
- [x] web/css/@media/index.md (#5238)
- [x] web/css/css_transforms/using_css_transforms/index.md (#5239)
- [x] web/css/max-block-size/index.md (#5240)
- [x] web/svg/attribute/accent-height/index.md (#5241)
- [x] web/svg/attribute/height/index.md (#5245)
- [x] web/svg/attribute/width/index.md (#5252)
- [x] web/svg/attribute/x/index.md (#5253)

cw118:
Done after #5253 🎉

PR #4058, opened 2022-02-08
Remove {{page}} macro from Glossaire files

cw118:
Part of #3886

### DIFF
--- a/files/fr/glossary/denial_of_service/index.md
+++ b/files/fr/glossary/denial_of_service/index.md
@@ -1,13 +1,9 @@
 ---
 title: Déni de Service
 slug: Glossary/Denial_of_Service
-tags:
-  - Attaque
-  - Déni de Service
-  - Glossaire
-  - Intro
-  - Sécurité
 translation_of: Glossary/Denial_of_Service
 original_slug: Glossaire/Déni_de_Service
 ---
-{{page("/fr/docs/Glossaire/Attaque_DOS")}}
+Le déni de service ou DoS («&nbsp;<i lang="en">**D**enial **o**f **S**ervice</i>&nbsp;» en anglais) est une catégorie d'attaques réseau. Il consomme les ressources disponibles d'un serveur, souvent en surchargeant celui-ci de requêtes, ce qui le fait fonctionner plus lentement ou le rend indisponible pour ses utilisatrices et utilisateurs légitimes.
+
+Voir [attaque DOS](/fr/docs/Glossary/DOS_attack) pour plus d'informations.
--- a/files/fr/glossary/denial_of_service/index.md
+++ b/files/fr/glossary/denial_of_service/index.md
@@ -1,5 +1,5 @@
 ---
-title: Déni de Service
+title: Déni de service
 slug: Glossary/Denial_of_Service
 translation_of: Glossary/Denial_of_Service
 original_slug: Glossaire/Déni_de_Service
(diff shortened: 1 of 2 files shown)


files/fr/glossary/denial_of_service/index.md
```diff
@@ -1,13 +1,9 @@
 ---
 title: Déni de Service
 slug: Glossary/Denial_of_Service
-tags:
-  - Attaque
-  - Déni de Service
-  - Glossaire
-  - Intro
-  - Sécurité
```
JuanVqz, line 9: Hi @cw118 should we avoid adding tags? I know you told me that they are optional for us, which one is the best approach?

cw118: If I remember correctly tags don't do much for translated-content, but I'm not the best person to ask — @SphinxKnight is it just for `fr` or should all locales avoid adding tags?

SphinxKnight: @JuanVqz hello, tags for localized pages are mostly a relicate of the old platform (kuma) and are not used anywhere or presented to the reader in any form. As such, I'd advise to remove/avoid them. That being written, they're not harmful so it's not a really big deal if they're kept.

SphinxKnight: To be completely honest, some tags may be used by some macros (cf. https://github.com/mdn/yari/blob/main/kumascript/macros/HTMLRef.ejs#L34 ) but this is not consistent depending on the section (in other words, not all sidebars (ex. JSRef or HTTPSidebar) use them)

SphinxKnight: On further inspection, 
1. A localized page shoud automatically/implicitly get the tags of its "parent" (i.e. English) page: https://github.com/mdn/yari/blob/main/kumascript/index.js#L67 (this also applies for other info in the frontmatter)
2. Here are the macros which are still used in English and which are using tags:

- `APIListAlpha` (used only on Web/API and targeting subpages with `Non-standard`, `Obsolete`, `Experimental`, `junk` )
- `HTMLRef` (used for many element pages, tag must have the "HTML XYZ" form)
  - apart from this form: current calls on en-US only relies on `Sectioning Root Element`, `multimedia`, `Deprecated`, `Web Components`, `Obsolete`
- `HTMLRefTable` (used only on Web/HTML/Element but scanning subpages)
- ~`LearnBox` which uses a parameter on tags to filter / there are only two calls for this macro, none of them uses this parameter~
- `ListGroups` used on Web/API and Web/Guide/API: only uses `Non-standard`, `Experimental` or `Obsolete` tags
- `MathMLRef` which only checks the tag `MathML:Element`
- `Non-standard*` which only uses the `Firefox OS` tag (present only on 9 pages in en-US)
- - `SVGRef` (used for many element pages, tag must have the "SVG XYZ" form)

So:
- Tags are not strictly necessary on localized pages
- Even for English pages, only a small subset of tags are still useful.

files/fr/glossary/denial_of_service/index.md
```diff
@@ -1,13 +1,9 @@
 ---
 title: Déni de Service
```
SphinxKnight, line 2: A single/minor nitpicking for this PR  : ) 
```suggestion
title: Déni de service
```

SphinxKnight (changes requested):
one case nitpicking to go before merging :)

SphinxKnight:
Thanks @cw118 !

files/fr/glossary/denial_of_service/index.md
JuanVqz: @SphinxKnight @cw118 Thank you for your time and great explanation. :bookmark_tabs: 
we will remove the tags, let's do it...

SphinxKnight (approved):
🙏

PR #4107, opened 2022-02-11
25. Screen [ES]

### LINKED ISSUE #4057: 25. Screen [ES]

JuanVqz:
- [x] Change from HTML to MD
- [x] Sync content

Fixes #4057


files/es/web/api/screen/index.md
```diff
@@ -0,0 +1,78 @@
...
+## Métodos
+
+- {{DOMxRef("Screen.lockOrientation")}} {{Non-standard_Inline}} {{Deprecated_Inline}}
+  - : Bloquea la orientación de la pantalla (sólo funciona en pantalla completa o para aplicaciones instaladas)
+- {{DOMxRef("Screen.unlockOrientation")}} {{Non-standard_Inline}} {{Deprecated_Inline}}
+  - : Desbloquea la orientación de pantalla (sólo funciona en pantalla completa o para aplicaciones instaladas)
+
+_Métodos heredados de {{domxref("EventTarget")}}:_
+
+{{Page("/es/docs/Web/API/EventTarget", "Methods")}}
```
JuanVqz, line 59: Aún sigue esta macro en la versión en Inglés y esta funcionando...

@mdn/yari-content-es, que se hace en esos casos?

https://github.com/mdn/content/blob/main/files/en-us/web/api/screen/index.md

files/es/web/api/screen/index.md
```diff
@@ -0,0 +1,78 @@
...
+- {{DOMxRef("Screen.top")}} {{deprecated_inline}}{{Non-standard_Inline}}
+  - : Devuelve la distancia en píxeles desde el borde supierior de la pantalla principal.
+- {{DOMxRef("Screen.width")}}
+  - : Devuelve la anchura de la pantalla.
+- {{DOMxRef("Screen.mozEnabled")}} {{Non-standard_Inline}} {{Deprecated_Inline}}
+  - : Booleano. Estableciendo su valor a `false`, se apagaá la pantalla del dispositivo.
+- {{DOMxRef("Screen.mozBrightness")}} {{Non-standard_Inline}} {{Deprecated_Inline}}
+  - : Controla el brillo de la pantalla del dispositivo. Espera un valor `double` entre `0` y `1.0`.
+
+### Manejador de eventos
```
Vallejoanderson, line 45: Creo que seria de la siguiente forma segun [la guia de traduccion](https://github.com/mdn/translated-content/discussions/4029).
```suggestion
### Detector de eventos
```

files/es/web/api/screen/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+- {{DOMxRef("Screen.onorientationchange")}} {{Deprecated_Inline}}
+  - : Un manejador para el evento {{event("orientationchange")}}.
```
Vallejoanderson, line 48:
```suggestion
  - : Un detector para el evento {{event("orientationchange")}}.
```

Vallejoanderson (changes requested):

files/es/web/api/screen/index.md
davbrito: En este caso el original dice `handler` y no `listener`. ¿Sería correcto traducirlo igual en ambos casos?

files/es/web/api/screen/index.md
davbrito: Lo mismo que en https://github.com/mdn/translated-content/pull/4107#discussion_r804799828

files/es/web/api/screen/index.md
davbrito: @JuanVqz @Vallejoanderson El original "manejador" está correcto. Detector lo usamos cuando es _listener_.

JuanVqz: Para mi no es lo mismo

files/es/web/api/screen/index.md
JuanVqz: @cw118 @SphinxKnight sorry for pinging you but we don't get how to handle this.

I've seen the [English version](https://github.com/mdn/content/blob/main/files/en-us/web/api/screen/index.md) is still using the `page` macro.

We're wondering if should we remove it or what should we do in this case? Do you know?

files/es/web/api/screen/index.md
JuanVqz: Listo! revertí los cambios

files/es/web/api/screen/index.md
JuanVqz: revertí los cambios! :+1:

Vallejoanderson: Genial, bien observado @davbrito.

files/es/web/api/screen/index.md
cw118: All good (and sorry for the late reply), I think it'd be best to wait for SphinxKnight's response but maybe this page macro could be replaced with the [methods found here](https://developer.mozilla.org/es/docs/Web/API/EventTarget#methods) (the heading that this `{{Page}}` transcludes), and a PR can opened at mdn/content to remove the `{{page}}` there?

JuanVqz: we thought that! just remove the `{{page}}` macro and use the methods there, okay, sounds good, let's wait for SphinxKnight's response, thank you for your help :pray:

davbrito: I think that the replacement of the macro should be a link to the page that it makes reference, instead of duplicating the content.

JuanVqz: > In this case it works because [the target page of the macro has an explicit `name` attribute](https://github.com/mdn/translated-content/blob/main/files/es/web/api/eventtarget/index.html#L16) for the corresponding section. Keeping this macro call as is would fail if the `EventTarget` page would be updated with `Métodos` (where, in this case `{{Page("/es/docs/Web/API/EventTarget", "Métodos")}}` would work.
> 
> 
> 
> Also in this case, I concur with @davbrito and @cw118, duplicating the content might not be the best call, here should EventTarget be updated (also I assume it's being used in a lot of pages). **Adding a link would be quite a simple solution which would fix the coupling issue (between source and target page) from the `{{page}}` call.**
> 
> 
> 
> To elaborate a bit more, as a rule of thumb, English content is the reference, so as long as the translated page does not have any issue for the reader, sticking to what mdn/content has (in terms of macro calls for instance) is a right "default strategy" (in other words, in this case, if I were to review this PR, I wouldn't block it because it uses the macro since it works and since it is present in the English page).

Thank you for your guidance and justify suggestion! 🙏

davbrito:
Tomando en cuenta lo mencionado [aquí](https://github.com/mdn/translated-content/pull/4107#discussion_r806121658), creo que ya podemos hacer merge a este PR ¿no?

files/es/web/api/screen/index.md
Graywolf9: Nota de referencia, se queda como está funcionando en inglés y se esperará a que se actualice ahí

PR #4121, opened 2022-02-12
Add quote tags with english language in modules guide

mkrtchian:
This is more semantic, and improves also accessibility.


tristantheb (approved):
It's ok, thanks for your contribution :D

PR #4165, opened 2022-02-16
convert to markdown and correct the reference of images

yin1999:
Changes:

- convert to markdown
- sync with english version


yin1999:
Images may cannot see in `pr preview` because images are only available on domain "developer.mozilla.org", please use local repo to see live-samples in this page.

irvin:
I guess we can probably wait for this to merge and reduce the need to re-remove the image absolute URL again. 
https://github.com/mdn/yari/pull/5215

let's wait for a couple of weeks to see if it can be landed in Q1.

yin1999:
Using relative paths for the pr mdn/yari#5215 has been merged.

PR #4167, opened 2022-02-16
#4018

alinggo:
#4018


SphinxKnight:
(failure for AutoCorrect is due to https://github.com/mdn/translated-content/issues/4170 and not due to this PR content (as far as I can tell, even if the files are for zh-CN, feel free to ignore it or to rebase this branch against a more recent state to avoid this workflow running)

SphinxKnight (changes requested):
It seems that this PR touches more files than it should.
If I refer to #4018 which is mentioned here, I can see there that only `files/zh-cn/learn/forms/how_to_structure_a_web_form/index.html` was edited.

@alinggo could you please change what's necessary to reduce the scope of changed files or explain why others are edited in this PR? thanks!

alinggo:
> It seems that this PR touches more files than it should. If I refer to #4018 which is mentioned here, I can see there that only `files/zh-cn/learn/forms/how_to_structure_a_web_form/index.html` was edited.
> 
> @alinggo could you please change what's necessary to reduce the scope of changed files or explain why others are edited in this PR? thanks!

Sorry, I'm not very good at using git yet

SphinxKnight:
@alinggo no worries, don't hesitate to ask :) 

if you're working on a single file, there is a relatively minor risk of conflict so rebasing/merging main onto your branch might not be necessary :)

PR #4183, opened 2022-02-17
Correct code order，And solve #4018 #4167

alinggo:


alinggo:
@SphinxKnight Is this correct？
Thank you so much！

SphinxKnight:
@alinggo regarding the "git" aspect of things, yes :) the scope looks right :)
Regarding the content of the PR itself, I'm no Chinese-speaker but looking at this code sample, it does match what's currently in English so it's correct to me.

I'll let @irvin check as well :)

irvin (approved):
yap the issue is just on the order of example code, nice.

PR #4296, opened 2022-02-27
Rename files/es/web/http/status/200/index.html to iptvlatina.cr.m3u

iptv2825:
https://tna5.bozztv.com/tvretro1971/index.m3u


sergiogval (changes requested):
Please report this user, I did it already. We need to do a dev branch it is a bad idea touching directly the main one.

PR #4304, opened 2022-02-27
minor formatting typo

hchiam:
I noticed that:
rp with bold star characters showing

probably should look more like:
**`<rp>`**
(bolded, and without the star characters)


hchiam (commented):
Ready for review.

audrasjb (approved):
LGTM.

PR #4330, opened 2022-03-01
Fix directives list syntax

tomsihap:
Use the same syntax used for the list of directives


SphinxKnight (approved):
Merci @tomsihap pour ces corrections.

Issue #4394, opened 2022-03-04
Issue with "继承": (short summary here please)


lvfangren:
MDN URL: https://developer.mozilla.org/zh-CN/docs/Web/CSS/inheritance

#### What information was incorrect, unhelpful, or incomplete?
translate repeat

#### Specific section or headline?
Notes part content repeat in Non-inherited properties part

#### What did you expect to see?
remove Non-inherited properties part repeat part content

#### Did you test this? If so, how?

<details>
<summary>MDN Content page report details</summary>

* Folder: `zh-cn/web/css/inheritance`
* MDN URL: https://developer.mozilla.org/zh-CN/docs/Web/CSS/inheritance
* GitHub URL: https://github.com/mdn/translated-content/blob/main/files/zh-cn/web/css/inheritance/index.html
* Last commit: https://github.com/mdn/translated-content/commit/34ed61153801e8c35ccd931afd994a3c7cfbcb20
* Document last modified: 2022-03-03T22:39:44.000Z

</details>

lvfangren:
对照英文原文  非继承属性部分中内容  存在  下一部分的内容重复。（重复翻译）

lvfangren:
add:  https://developer.mozilla.org/zh-CN/docs/Glossary/Node.js this page translate miss something and https://developer.mozilla.org/zh-CN/docs/Web/JavaScript/Guide/Modules jumping incorrect to english version page

lvfangren:
补充： [这个页面]( https://developer.mozilla.org/zh-CN/docs/Glossary/Node.js)翻译不完善。

[这个页面](https://developer.mozilla.org/zh-CN/docs/Web/JavaScript/Guide/Modules)的Node.JS跳转到英文版

cw118:
Thank you for the report and for fixing this issue :) Closing since this was dealt with in #4398

PR #4398, opened 2022-03-04
fix: delete repeat content

lvfangren:
fix translate repeat https://github.com/mdn/translated-content/issues/4394


lvfangren:
i don't sure something e.g: change /zh-cn/web/javascript/guide/modules/index.html jump href and /zh-cn/glossary/node.js/index.html translate sync

PR #4401, opened 2022-03-04
Fix for vertical-align examples in Dark Mode

### LINKED ISSUE #4402: vertical-align article’s interactive example has dark text on dark background
## What page(s) did you find the problem on?

https://developer.mozilla.org/zh-CN/docs/Web/CSS/vertical-align

## What is the problem?

When I switch to black mode, I can't see the specific content selected in my box in the figure，but I like to browse in black mode very much.

b31b04a73694a8c32946c2236320e01
56e81e4338747044827debd8d081bae

NiedziolkaMichal:
It should fix #4402 by making iframes use "sample-code-frame" class instead of "example-outcome-frame".
en-US examples have the same tables written as
{{EmbedLiveSample("Vertical_alignment_in_a_line_box", '100%', 160, "", "")}}
{{EmbedLiveSample("Vertical_alignment_in_a_table_cell", '100%', 230, "", "")}}

So I am guessing that if last argument is undefined, "sample-code-frame" class will be used, which gives white background to whole iframe.

Additionally I increased height, so there will be no scroll bar.


caugner (changes requested):
I think it's a good practice to omit the empty parameters.

~~As for increasing the height, we should probably wait until we resolved https://github.com/mdn/yari/issues/5431, because we might be able to fix this globally in yari. Could you revert to the old height for the time being? Sorry for that!~~

**edit**: Never mind, you're just using the same height for both, that makes sense!

NiedziolkaMichal:
I could make make size match en-US, so 160-230 and also change width to 100%. Now I can see that other translations have the same problem, so I can set same size for all of them in this PR.

caugner:
Good catch and great initiative, thanks!

If you could make those consistent, that would be terrific.

NiedziolkaMichal:
Alright, it's done.

caugner (changes requested):
It seems that you have accidentally committed your `.idea` directory.

If you don't know how to remove them, `git rm -r --cached .idea` should do the trick.

NiedziolkaMichal:
Yes but I have reverted commit and added another one without .idea folder, so it should be fine, right?
When I requested for all changes file names by "git diff --name-only patch-1 main", it git shown:
files/de/web/css/vertical-align/index.html
files/ko/web/css/vertical-align/index.html
files/pt-br/web/css/vertical-align/index.html
files/ru/web/css/vertical-align/index.html
files/zh-cn/web/css/vertical-align/index.html

Also "git rm -r --cached .idea" returned "fatal: pathspec '.idea' did not match any files".

Sorry about the mess.

NiedziolkaMichal:
Could you redo checks please? I have removed invalid quote.

caugner:
@NiedziolkaMichal Unfortunately, the tests don't seem to like passing "100%" as width. Can we stick with 1200 for now, or was that to narrow/wide?

There might be a good reason why we don't use a width of 100%, but I don't know it (yet). To me it would make sense if the width parameter was optional and defaulted to 100% (or 1200). Then again, if we provide a width, we also have to provide a height, unless named parameters were supported.

NiedziolkaMichal:
Ok it's done. Visually there is no difference between 1200px and 100%, even on very big screen sizes.

files/de/web/css/vertical-align/index.html
```diff
@@ -61,7 +61,7 @@
 </pre>
 </div>
 
-<p>{{EmbedLiveSample("vertical-align-inline", 1200, 120, "", "", "example-outcome-frame")}}</p>
+<p>{{EmbedLiveSample("vertical-align-inline", '1200', 160)}}</p>
```
caugner, line 64: The quotes were only necessary because of the % sign. For numeric values, it's better to avoid them:

```suggestion
<p>{{EmbedLiveSample("vertical-align-inline", 1200, 160)}}</p>
```

caugner (changes requested):
Thanks!

Apart from the quotes, the code changes look good.

I will have a look at the previews tomorrow.

NiedziolkaMichal:
You are right, it's done.

caugner (approved):

NiedziolkaMichal:
Ok, thank you

caugner:
LGTM!

I verified the previews and the `vertical-align` examples are indeed now readable even in dark-mode, which is an improvement and the first half of the solution:
image

The second half of the solution (i.e. to make these live examples look dark in dark-mode), we have to update yari here:
https://github.com/mdn/yari/blob/0f206cee92975b9e73f1cb8518528e8857d2a5fd/kumascript/src/live-sample.js#L23-L32

NiedziolkaMichal:
That would be a nice improvement, currently white background doesn't fit dark mode, but at least examples are readable. If you would review my other PRs in your free time, I would appreciate it.

caugner:
I created https://github.com/mdn/yari/issues/5491 to track the remaining issue with the white background.

PR #4479, opened 2022-03-08
Update console.dir

SphinxKnight:
A small update after seeing this "non up to date" document on a browsing session


cw118 (approved):
Merci encore sphinx 😊

PR #4525, opened 2022-03-10
Patch #2846 zh-cn/docs/web/html/* pages with an EmbedLiveSample error

UESTCzhouyuchuan:
## Event
It seems that  EmbedLiveSample error occur in several pages example sections at zh-cn/docs/web/html/*, like this
image
[https://developer.mozilla.org/zh-CN/docs/web/html/element/pre](https://developer.mozilla.org/zh-CN/docs/web/html/element/pre#%E7%BB%93%E6%9E%9C)
After hours checking, I found these pages had the same issue, then I fix them
- https://developer.mozilla.org/zh-cn/docs/web/html/element/pre
- https://developer.mozilla.org/zh-cn/docs/web/html/element/u
- https://developer.mozilla.org/zh-cn/docs/web/html/global_attributes/itemscope
- https://developer.mozilla.org/zh-cn/docs/web/html/element/iframe
- https://developer.mozilla.org/zh-cn/docs/web/html/element/input
- https://developer.mozilla.org/zh-cn/docs/web/html/element/input/date
- https://developer.mozilla.org/zh-CN/docs/web/html/element/textareahr
- https://developer.mozilla.org/zh-CN/docs/web/html/element/hr

## reason
1. code demo dosen't have class named `class="brush:language-type"`
2. EmbedLiveSample need The ID of the heading or enclosing block to draw the code from, which means give the specific ID or let EmbedLiveSample and code demo in the same header
https://developer.mozilla.org/en-US/docs/MDN/Structures/Live_samples


UESTCzhouyuchuan:
Anybody there?

UESTCzhouyuchuan:
为什么还没code review啊

irvin:
Hi,

All of the zh reviewers were busier this past week (like you, we were all volunteers too) on the day job and your PR contains 8 files, thus requiring a longer time to review and cannot be processed in daily handy time. Please accept a couple of days to no longer than two weeks.

Also suggest to keep one file per PR in the future.

UESTCzhouyuchuan:
> Hi,
> 
> All of the zh reviewers were busier this past week (like you, we were all volunteers too) on the day job and your PR contains 8 files, thus requiring a longer time to review and cannot be processed in daily handy time. Please accept a couple of days up to no longer than two weeks.
> 
> Also suggest to keep one file per PR in the future.

Althought one file one pr is ineffective, but all right

PR #4592, opened 2022-03-13
Fixes 4586 - minor fixes for <em> page

### LINKED ISSUE #4586: [French] HTML's <em> page contains confusing inclusive typography (interpunct)
MDN url : https://developer.mozilla.org/fr/docs/Web/HTML/Element/em

This page contains what's wrongfully called "inclusive" writing. "Inclusive" writing is a political way of writing based on a political and subjective statement that french's gender-oriented language has to be replaced by writing human language as programmation code and including all genders and variations of a word at the same time and separating them with a midpoint. "dérouté" becomes "dérouté⋅e", as in the MDN's page. However, it has been discussed AND denied by both the government and France's Académie Française that is dedicated to determine our language's standards. It is even forbiden to use it in any of our public services. It is also rejected by a majority of french people and is used by only a handful of individuals.

Last but not least, as someone who suffers of multiple reading disabilities, it makes the task even mor painful for me and people like me.

"dérouté⋅e" should be written "dérouté⋅e" and no midpoint should be used until an official change in the way of writing french.

cw118:
Fixes #4586


SphinxKnight (approved):
Merci @cw118

PR #4676, opened 2022-03-17
fix: scroll event Bubbles should be No

hnsxxscyx:


yin1999:
The `Bubbles` property does be `Yes`.

You can try the following code in console:

```js
window.addEventListener('scroll', e => { console.log(e.bubbles)} ) // true
```

hnsxxscyx:
> https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll_event Yes, the document scroll event is different from the scroll event of the element.

Got it, I missed that, scroll event on document can Bubble

PR #4677, opened 2022-03-17
Fixes broken example after #4270

SphinxKnight:
see https://github.com/mdn/translated-content/pull/4270#issuecomment-1070912880


SphinxKnight:
waiting for the build / test to pass and will merge this shamelessly to fix my own mistake 😞

SphinxKnight:
@julban here it is, https://whatsdeployed.io/s/16d/mdn/translated-content will tell you once this is public.

PR #4697, opened 2022-03-18
fixes 4269-introduced breakage in some code samples

SphinxKnight:


SphinxKnight:
did a check looking for successive lines with only whitespace, did not find any other file with this pattern

SphinxKnight:
merging shamelessly for the same reason as in https://github.com/mdn/translated-content/pull/4677#issuecomment-1070924931 
(not making an habit of it)

PR #4714, opened 2022-03-18
Convert to markdown type and translated based on latest es-US content

Malaaaa:
I just retranslate the learn/HTML page based on the english version. I don' know why there has too commits, Is that because I just fetch the MDN/translated-content?


yin1999:
For convert document from `index.html` to `index.md`, please using the following commits:

1. use a single commit to rename the file (from `.html` to `.md`), this will help us to track the history

    commit: convert to `.md` for some document

       git mv /path/to/some/index.html /path/to/some/index.md # 

2. use the second commit to convert the content and sync with english version

    commit: convert to markdown and sync with english version

You can refer: [commits in #4583](https://github.com/mdn/translated-content/pull/4583/commits)

PR #4764, opened 2022-03-21
Translated mixed content pages

SphinxKnight:
This page (and its children), linked from the Firefox DevTools was not available in French.


cw118 (approved):
Merci encore @SphinxKnight 🔥

PR #4774, opened 2022-03-22
Update pt-BR parseFloat page

### LINKED ISSUE #4769: pt-BR parseFloat page needs some work
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat

It is missing some macros like {{Compat}} and {{Specifications}}. To Update content itself would be great as well.

en-US version: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat

ghost:
Feito com muito carinho, então faça a review, por favor. Fix: [#4769](https://github.com/mdn/translated-content/issues/4769)


victor0x16:
Hi @cw118, do you know why EmbedInteractiveExample macro has ```Try it``` as a title?

I don't see that in my local machine. Over here looks fine but in the preview not so much.

cw118:
@victor0x16 I believe the title is part of the macro itself, so there's no need to worry about that. I also see it on the [English page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat) and [French page](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/parseFloat).

As for why it didn't show up locally, I'm not quite sure but I'll try to look into it in a bit :)

cw118:
@victor0x16 alright, I think I figured out why. Turns out the "Try it" title for `embedinteractiveexample` macros was just recently added (cf. mdn/yari#5667) and the version of yari used by mdn/content (and translated-content) hasn't been updated to have that "feature" yet, which is why we don't see it locally but it's there on deployment and preview.

victor0x16:
@cw118 ,  thank you for looking it up!

I do believe that this title on the macro seems unnecessary and it doesn't fit in the pages specially to translated ones. But maybe there is something I'm missing here.

cw118:
> I do believe that this title on the macro seems unnecessary and it doesn't fit in the pages specially to translated ones. But maybe there is something I'm missing here.

There is valid motivation for adding the title, notably improving navigation for readers (cf. [this discussion on yari](https://github.com/mdn/yari/discussions/5453)); yes, we're also concerned about the effect of this addition on locales and are discussing ways to potentially localize it :)

SphinxKnight:
@mdn/yari-content-pt-br is there something blocking this from being reviewed?

josielrocha (approved):
LGTM ;)

josielrocha:
Obrigado pela contribuição @victor0x16!

PR #4824, opened 2022-03-25
add notecard block

unsky559:
Link to another similar page


sashasushko:
@unsky559 unfortunately, I have to close this PR.

We cannot deviate from the original in translation, and there is no such link in the English version.

sashasushko:
@unsky559 but thank you for the help.

p.s. There is a variant how to add this link - is to start from the English version.

Issue #4822, opened 2022-03-25
`Firefox Developer Tools docs` is removed from `mdn/content`


yin1999:
## What is the problem?

`Firefox Developer Tools docs` in `en-US` is now redirected to [Firefox DevTools User Docs](https://firefox-source-docs.mozilla.org/devtools-user/index.html#firefox-devtools-user-docs) by mdn/content#14140

Should we follow this changes in this repo?

SphinxKnight:
Thanks for the heads up @yin1999, I guess the workflow for this would be to use `yarn content --delete redirect` for every current page under tools

SphinxKnight:
After fixing a missing logic w/ https://github.com/mdn/yari/pull/5814, I believe using the command line with 
` yarn content delete <slug> <locale> --redirect https://firefox-source-docs.mozilla.org/<sthg>`
will work.
For instance, to delete/redirect one of the French page, I'll use 
` yarn content delete Tools/3D_View fr --redirect https://firefox-source-docs.mozilla.org/devtools-user/3d_view/index.html`

(quick reminder: `yarn content` works from `/content` directory whereas `yarn tool` works from `/yari` directory)

SphinxKnight:
if it may be of any help to any maintainer out there, here is the script I used for the French "surface" of Tools: https://gist.github.com/SphinxKnight/16d971e073338c5da4403874a3d8227b
I hope this would help reduce the workload (leaving only the pages for Tools that you may have in addition to the ones French had)

yin1999:
I have wrote two scripts(in `go` and `python`) to deal with removing `Tools` section.

It resolves the `Tools` section with the following steps:

- List all the file with `.html` and `.md` ext, save the path to a string list.
- sort those path with inverse lexicographical order
- get the redirect url by `GET` "https://developer.mozilla.org/en-US/docs/" + `<slug>`
- exec command `yarn content delete <slug> <locale> --redirect <redirect url> -y` at `content` repo dir

Some documents need to be resolved manually. (then run the script below again)

I have precompiled the executable files below (for go).

## source

### go

#### usage

```bash
./main -pwd /path/to/mdn/content -d /path/to/mdn/translated-content/ -locale <locale>
```

```go
package main

import (
	"bufio"
	"bytes"
	"flag"
	"fmt"
	"io/fs"
	"log"
	"net/http"
	"os"
	"os/exec"
	"path/filepath"
	"sort"
	"strings"
)

var (
[...]

PR #4841, opened 2022-03-26
Remove Tools section from MDN - French

SphinxKnight:
This takes care of #4822 / is the fr counterpart of https://github.com/mdn/content/pull/14140

poking @solfen for old times' sake

Regarding the review, I think the most important file to check here is `_redirects.txt`


cw118 (approved):
J'ai verifié la plupart des redirections et tout semble OK, merci pour tout ce travail :)

PR #4845, opened 2022-03-27
remove `Tools` section from `zh-TW`

yin1999:
Part of #4822

Removed by running go script [here](https://github.com/mdn/translated-content/issues/4822#issuecomment-1079841949)

There is still an error on file `files\zh-tw\tools\firefox_os_1.1_simulator\index.html`, this seems cannot be found in `firefox-source-docs.mozilla.org`. Any suggestion ?


SphinxKnight:
Firefox OS doc should be deleted

yin1999:
> firefox_os_1.1_simulator

Done in 6226fbe

PR #4765, opened 2022-03-21
Fix #4762 (markdown detection) and normalize line endings 

SphinxKnight:
Fixes #4762
See https://github.com/mdn/content/pull/809 and https://github.com/mdn/content/pull/14123 for context from mdn/content.

Thanks @yin1999 for filing https://github.com/mdn/translated-content/issues/4762 and @sideshowbarker for fixing an itch of mine :)

N.B. If I understood https://github.com/github/linguist/blob/master/lib/linguist/documentation.yml correctly, the `docs` folder is rightly excluded from language stats.

PR #4923, opened 2022-03-31
Make GitHub count all files/** for the Languages statistics

caugner:
# Before

image

# After

image

See:
- https://github.com/mdn/content/pull/14123
- https://github.com/github/linguist/blob/323ec129c4a359f9c9430175bc391e0176aefb6a/docs/overrides.md

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*	text=auto eol=lf
+files/**	linguist-detectable


cw118:
@caugner Thanks for the PR! However, I believe this is a duplicate of #4765

caugner:
Good catch @cw118!

I believe the difference is that this PR only adds `files/**` to the Languages statistics (and not `docs/**`), independently of their file type, which reveals JSON, Text and SVG, and #4765 adds all Markdown files (except the ones in the root folder) and converts the line-endings of some files.

But I will close this PR in favor of the existing one.

PR #4924, opened 2022-03-31
chore(git): ignore .DS_Store

caugner:
See:
- https://github.com/mdn/content/blob/107f88d6387ffa33f1efba84228e5dd0bddf88a0/.gitignore#L7
- https://en.wikipedia.org/wiki/.DS_Store


schalkneethling (approved):
🎉

PR #4950, opened 2022-04-02
Clarify meaning of `Strict` attribute of `SameSite` cookie policy

pierre42100:
It make more sense to explain `Strict` policy using `provenant de` rather than `vers`.

English original text:
```
Cookies will only be sent in a first-party context and not be sent along with requests initiated by third party websites.
```


cw118 (approved):
Merci @pierre42100 et félicitations pour cette première contribution !

Issue #4545, opened 2022-03-11
[ko] Duplicated mdn/content images/assets files should be removed from the tree


SphinxKnight:
Since https://github.com/mdn/yari/pull/5215 has been merged, there is no apparent reason to have duplicated files from mdn/content in mdn/translated-content. After fiddling [a small script](https://gist.github.com/SphinxKnight/60d766b8d6cd970b42eaf967b8dac3ff) (adapted for ko), I was able to identify the following files under the tree which are identical and should be removed.

- `translated-content/files/ko/web/html/element/img/favicon72.png`
- `translated-content/files/ko/web/html/element/img/clock-demo-400px.png`
- `translated-content/files/ko/web/html/element/img/favicon144.png`
- `translated-content/files/ko/web/media/formats/audio_concepts/human-hearing-range.svg`
- `translated-content/files/ko/web/media/formats/audio_concepts/audio-waveform.svg`
- `translated-content/files/ko/web/media/formats/audio_concepts/audio-waveform-samples1.svg`
- `translated-content/files/ko/web/api/web_audio_api/audio-context_.png`
- `translated-content/files/ko/web/api/web_audio_api/advanced_techniques/sequencer.png`
- `translated-content/files/ko/web/api/web_audio_api/using_iir_filters/iir-filter-demo.png`
- `translated-content/files/ko/web/api/web_audio_api/visualizations_with_web_audio_api/wave.png`
- `translated-content/files/ko/web/api/web_audio_api/visualizations_with_web_audio_api/bar-graph.png`
- `translated-content/files/ko/web/api/web_audio_api/controlling_multiple_parameters_with_constantsourcenode/customsourcenode-as-splitter.svg`
- `translated-content/files/ko/web/api/web_audio_api/web_audio_spatialization_basics/web-audio-spatialization.png`
- `translated-content/files/ko/web/api/analysernode/fttaudiodata_en.svg`
- `translated-content/files/ko/web/api/element/getboundingclientrect/element-box-diagram.png`
- `translated-content/files/ko/web/web_components/using_shadow_dom/dom-screenshot.png`
- `translated-content/files/ko/web/web_components/using_shadow_dom/shadowdom.svg`
- `translated-content/files/ko/learn/css/first_steps/what_is_css/html-example.png`
- `translated-content/files/ko/learn/tools_and_testing/client-side_javascript_frameworks/react_todo_list_beginning/unstyled-app.png`
[...]

PR #5031, opened 2022-04-08
deleted duplicated mdn/content images/assets #4545

yechoi42:
#4545 에서 명기된 파일을 모두 삭제하였습니다


cw118:
Hi there, just a quick but important note that these files should stay until mdn/yari#5652 is fixed, otherwise Korean live samples will be broken (cf. French pages which have this problem :/). 

The assets fix had worked locally in dev but does not currently work in prod, so please don't merge this before the yari issue is resolved.

cw118:
> @cw118 Would it be okay to merge now?

Hello, as far as I know the issue is unfortunately still not resolved, so if this is merged then Korean live samples that use images from mdn/content will not work (cf. [the fr CSS filter page](https://developer.mozilla.org/fr/docs/Web/CSS/filter#blur) for examples).

hochan222:
Thank you for your reply. It's been quite some time for PR, so I asked once. 

Have a nice day :)

yin1999:
Hi @hochan222, we can go ahead now. The platform issue has been fixed by mdn/yari#7917. We can now safely remove duplicated images/assets. Please merge main branch and check the related content changes :)

hochan222 (approved):
LGTM :)

Issue #4675, opened 2022-03-17
Issue with "Faciliter la programmation asynchrone avec async e…": broken links to GitHub examples


StephaneRegnie:
MDN URL: https://developer.mozilla.org/fr/docs/Learn/JavaScript/Asynchronous/Promises

#### What information was incorrect, unhelpful, or incomplete?
As for the page "https://developer.mozilla.org/fr/docs/Learn/JavaScript/Asynchronous/Introducing", all links to examples files are broken (404 error).

#### Specific section or headline?
All sections.

#### What did you expect to see?
The files to check on GitHub or example pages as for other pages of this tutorial.

#### Did you test this? If so, how?
Yes for all links.

<details>
<summary>MDN Content page report details</summary>

* Folder: `fr/learn/javascript/asynchronous/promises`
* MDN URL: https://developer.mozilla.org/fr/docs/Learn/JavaScript/Asynchronous/Promises
* GitHub URL: https://github.com/mdn/translated-content/blob/main/files/fr/learn/javascript/asynchronous/promises/index.md
* Last commit: https://github.com/mdn/translated-content/commit/d53bae971e455859229bcb3246e29bcbc85179d2
* Document last modified: 2022-03-08T00:12:08.000Z

</details>

Issue #4647, opened 2022-03-16
Issue with "JavaScript asynchrone coopératif : délais et inter…": Broken links to GitHub examples


StephaneRegnie:
MDN URL: https://developer.mozilla.org/fr/docs/conflicting/Learn/JavaScript/Asynchronous_ae5a561b0ec11fc53c167201aa8af5df

#### What information was incorrect, unhelpful, or incomplete?
As for the page "https://developer.mozilla.org/fr/docs/Learn/JavaScript/Asynchronous/Introducing", all links to examples files are broken (404 error).

#### Specific section or headline?
All sections.

#### What did you expect to see?
The files to check on GitHub or example pages as for other pages of this tutorial.

#### Did you test this? If so, how?
Yes for all links.

<details>
<summary>MDN Content page report details</summary>

* Folder: `fr/conflicting/learn/javascript/asynchronous_ae5a561b0ec11fc53c167201aa8af5df`
* MDN URL: https://developer.mozilla.org/fr/docs/conflicting/Learn/JavaScript/Asynchronous_ae5a561b0ec11fc53c167201aa8af5df
* GitHub URL: https://github.com/mdn/translated-content/blob/main/files/fr/conflicting/learn/javascript/asynchronous_ae5a561b0ec11fc53c167201aa8af5df/index.md
* Last commit: https://github.com/mdn/translated-content/commit/d53bae971e455859229bcb3246e29bcbc85179d2
* Document last modified: 2022-03-08T00:12:08.000Z

</details>

cw118:
This is a conflicting page and has been deleted/redirected in #5037. Once that [PR is deployed](https://whatsdeployed.io/s/16d/mdn/translated-content), this issue should be fixed. We can reopen later on if needed.

StephaneRegnie:
OK @cw118, thanks 👍

PR #5037, opened 2022-04-09
JS conflicting asynchronous cleanup for fr

cw118:
There are several issues related to conflicting pages in this module (#4613, #4647, #4675), so I tried cleaning up the folder `conflicting/Learn/JavaScript/Asynchronous`. Let me know if there's anything to fix as I've never done it before :)

`fr` also seems to build with yari on my local machine.


SphinxKnight:
Merci ! tout m'a l'air bon :)

PR #5057, opened 2022-04-10
Fixed wrong class name in example

Jonatandb:
image


Jonatandb:
I have changed the CSS with developer tools and it shows correctly the black border if the outer div have the 'outer' class instead 'external'. I will make a new PR.

PR #5078, opened 2022-04-12
add Chinese version of structuredClone api

molang3739:
add Chinese version of structuredClone api


irvin:
I'm not sure why this is build failed, i guess it's the folder's upper case

PR #5101, opened 2022-04-14
remove `Tools` section from `pt-BR`

yin1999:
Part of #4822

Removed by running go script [here](https://github.com/mdn/translated-content/issues/4822#issuecomment-1079841949)


nathipg (approved):
Obrigada pela contribuição!

PR #5180, opened 2022-04-20
Remove two last occurrences of TODO macro

teoli2003:
This PR removes the last 2 occurrences of the `{{TODO}}` macro in the whole of mdn (mdn/content and mdn/translated-content).

After this is merged, we can actually delete the macro from mdn/yari !

### DIFF
--- a/files/de/web/css/align-items/index.html
+++ b/files/de/web/css/align-items/index.html
@@ -65,7 +65,7 @@ align-items: unset;
 
 <h2 id="Beispiele">Beispiele</h2>
 
-<p>{{ TODO() }}</p>
+<p>TBD</p>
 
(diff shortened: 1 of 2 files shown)


teoli2003:
cc/ @schalkneethling as I think the German area is unmaintained but it would be cool to land this!

caugner (approved):
LGTM

cw118:
It seems that the Korean file hasn't been translated, so this PR doesn't affect `ko` much and should be ready to merge — poking @mdn/yari-content-ko for review just in case :)

hochan222 (approved):
ko locale, LGTM :)

schalkneethling (approved):
Thanks, @teoli2003 🎉  🧹

PR #5223, opened 2022-04-23
Adjusts in portuguese translation

Grubba27:
In portuguese translation there are some strange code lost in the page: as this one that were removed //=> [1, 2, 3, 4, 5, 6, 7, 8, 9]
I don't think that was a important line because the value of these numbers are not related to the code above them.


josielrocha (approved):

josielrocha:
Obrigado pela contribuição @Grubba27!

PR #5237, opened 2022-04-24
Fix Japanese translation for WindowOrWorkerGlobalScope.fetch()

akiomik:
Original text:

> You can also use an object literal as headers in init.

https://github.com/mdn/content/blob/dad1bb01d2c72cec1bc6fe89fb044911c404c669/files/en-us/web/api/fetch/index.md?plain=1#L326-L327


mfuji09 (approved):
OKです。

Issue #3884, opened 2022-01-29
{{page}} macro removal for French pages - Web/JavaScript


SphinxKnight:
Fragment of #3881 in order to remove `{{page}}` calls from pages.
There are 44 pages under Web/JavaScript:

- [x] `web/javascript/data_structures/index.md`(#4921)
- [x] `web/javascript/guide/indexed_collections/index.md` (#3570)
- [x] `web/javascript/reference/index.md` (#4922)
- [x] `web/javascript/reference/global_objects/arraybuffer/index.md` (#5048)
- [x] `web/javascript/reference/global_objects/asyncfunction/index.md` (#5049)
- [x] `web/javascript/reference/global_objects/bigint/index.md` (#5133)
- [x] `web/javascript/reference/global_objects/boolean/index.md` (#5202)
- [x] `web/javascript/reference/global_objects/dataview/index.md` (#5203)
- [x] `web/javascript/reference/global_objects/date/tolocaledatestring/index.md` (#4151)
- [x] `web/javascript/reference/global_objects/date/tolocalestring/index.md` (#4151)
- [x] `web/javascript/reference/global_objects/error/index.md` (#5234)
- [x] `web/javascript/reference/global_objects/evalerror/index.md`(#5254)
- [x] `web/javascript/reference/global_objects/function/index.md` (#5255)
- [x] `web/javascript/reference/global_objects/generatorfunction/index.md` (#5256)
- [x] `web/javascript/reference/global_objects/internalerror/index.md` (#5258)
- [x] `web/javascript/reference/global_objects/intl/collator/index.md` (#5260)
- [x] `web/javascript/reference/global_objects/intl/datetimeformat/index.md` (#5261)
- [x] `web/javascript/reference/global_objects/intl/datetimeformat/formatrange/index.md` (#5262)
- [x] `web/javascript/reference/global_objects/intl/listformat/index.md` (#5264)
- [x] `web/javascript/reference/global_objects/intl/numberformat/index.md` (#5266)
- [x] `web/javascript/reference/global_objects/intl/pluralrules/index.md` (#5267)
- [x] `web/javascript/reference/global_objects/intl/relativetimeformat/index.md` (#5268)
- [x] `web/javascript/reference/global_objects/map/index.md` (#5276)
- [x] `web/javascript/reference/global_objects/number/tolocalestring/index.md` (#5277)
- [x] `web/javascript/reference/global_objects/object/index.md` (#5279)
[...]

PR #5279, opened 2022-04-26
Remove {{page}} from `Object`

SphinxKnight:
Part of #3884
The last section before Specification has been voluntarily removed in fr. Wording is unclear (if not ambiguous) and I struggle to find what it brings to this page.


cw118 (approved):
😃

PR #5324, opened 2022-04-28
Cleanup orphaned pages - Web API - A

SphinxKnight:
Well that's just 2 pages before I encountered a more complicated case, so here is a small PR :)


cw118 (approved):
Merci encore :)

PR #5327, opened 2022-04-28
Cleanup orphaned - E

SphinxKnight:
One new page and the rest of redirects.
I decided to go with a redirect for `orphaned/Web/API/ElementTraversal` as well (same argument as in #5325)


cw118 (approved):
Merci 😀

PR #5332, opened 2022-04-29
add String.prototype.at() zh-cn translation

liunnn1994:


asdjgfr:
@irvin Can we merge this pr? I have an article want to quote this, thx.

PR #5337, opened 2022-04-29
Cleanup orphaned - S

SphinxKnight:
I went with an update as well for `SVGElement`


cw118 (approved):
Merci beaucoup sphinx !

PR #5377, opened 2022-05-02
Conflicting cleanup - Learning Area

SphinxKnight:
Most (like in 95%) of the pages in conflicting were clearly lagging behind so I went with "simply" adding redirects here.
Updates on LA pages should be part of other usual / maintenance PRs.


cw118 (approved):
Merci beaucoup :)

PR #5409, opened 2022-05-04
Remove `{{exception}}` macro - fr

SphinxKnight:
This does the job of https://github.com/mdn/mdn-community/discussions/64 for `fr`.
I updated both pages vs their `en-US` counterparts and removed unnecessary macros (incl. but not limited to `{{exception}}`)


cw118 (approved):
Merci 🤗

SphinxKnight:
Merci pour toutes les revues @cw118 :pray: :bow:

PR #5501, opened 2022-05-09
convert to markdown and sync with english version for WebAssembly

yin1999:
This document does has any heading element, so the BCD is shown in wrong place. Fixes this by add heading element.

But I'm not very familiar with traditional Chinese, so I haven't translated the `API 參考` section. **Help wanted on this section.**


irvin:
> But I'm not very familiar with traditional Chinese, so I haven't translated the `API 參考` section. **Help wanted on this section.**

it's ok to leave it for other people to fix in future

PR #5581, opened 2022-05-11
Replace all remaining {{(m)anch}} calls with anchor tags

### LINKED ISSUE #4614: Replace {{anch}} macro with Markdown links
Over in https://github.com/mdn/content/pull/13802 , @queengooborg replaced all usages of the [`{{anch}}`](https://github.com/mdn/yari/blob/main/kumascript/macros/anch.ejs) macro with Markdown links.

If we could remove it from translated-content as well, we could remove the macro code from Yari.

@queengooborg used a script to do this which you can see at https://github.com/mdn/content/pull/13802 (I think this is the latest version isn't it?) and which I think should work with translated content?

queengooborg:
This PR fixes #4614 by removing all remaining calls to the now-deprecated `{{anch}}`  and `{{manch}}` macros and replacing them with anchor tags or Markdown links, using the script from https://github.com/mdn/content/pull/13802.

Note: this is a scripted change.  I'm more than happy to separate this into each individual language and/or category to make review easier!


yin1999:
I found there is one file that has been changed in this pr for `l10n-zh`, and the anch seems not match with `id` in this document. If that's ok to undo this change, I'd like to deal with it manually.

queengooborg:
Of course, @yin1999!  I've reverted the change to that file now!

queengooborg:
I decided that it would be better for reviewers and fellow contributors to keep the scope of this PR down to each language, so I'm going to close this PR and split it up!

yin1999:
> Of course, @yin1999! I've reverted the change to that file now!

I've created #5594 to resolve it.😘

PR #5584, opened 2022-05-12
Fix <a> inside last <h2> and remove invisible characters

carlossgabriel:
Inside the code, i found invisible characters that shouldn't be used, so my contribution just aims to replace them by normal spaces aside the minor fix in the last `<h2>`, where all the links should be inside the `<a>` tag


carlossgabriel:
Hi, @nathipg 
I've fixed some minor things on the page related to invisible characters and some misplaced `<a>` tags.

But as I was looking through the repository pull requests, I realised that almost all pull request with @josielrocha as reviewer are opened and without conclusion.

I wonder if you could change the reviewer to someone more active.

josielrocha (approved):

josielrocha:
Obrigado pela contribuição @carlossgabriel!

PR #5610, opened 2022-05-12
Remove {{obsolete_header}} from fr

cw118:
I excluded orphaned pages since those will be dealt with in the "cleanup orphaned" PRs (I will try to review them soon :)). For conflicting pages with this macro, I decided to create #5611.

Some pages were deleted because they don't exist in en-US, and I didn't find any redirects for them either.


SphinxKnight (approved):
Merci @cw118 :)

PR #5624, opened 2022-05-13
Remove `{{TemplateLink}}` from `fr`

SphinxKnight:
I updated the 4 pages accordingly and included specific information regarding either localization or trying to get rid of "link" macros.


cw118 (approved):
🙏

PR #5631, opened 2022-05-13
Remove {{obsolete_inline}} from fr

cw118:
This is mostly a simple "find and replace" PR since the macro appears so many times. Orphaned is also excluded here, I decided to deal with `conflicting` in another PR. `mdn/structures/macros/commonly-used_macros` is also excluded since #5624 updates the entire page.


SphinxKnight (approved):
:fire:

PR #5655, opened 2022-05-14
fix(cookies): grammatical errors and wrong name prefix

maxxborer:
**EN:**
- Fixed grammatical errors on the Cookies page.
- Replaced `__Host-` with `__Secure-` in the second case (in the description of prefixes), according to the [documentation in English](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent).

**RU:**
- Исправил грамматически ошибки на странице Cookies.
- Заменил `__Host-` на `__Secure-` во втором случае (в описании префиксов), согласно [документации на английском языке](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent).


myshov (approved):

myshov:
@mvbureev спасибо!

PR #5666, opened 2022-05-15
Remove `{{HTML5ArticleTOC}}` from `fr`

SphinxKnight:
This is a simple removal/redirection for the last (orphaned) page which contained the macro.


cw118 (approved):
Merci 🔥

PR #5705, opened 2022-05-17
Clean conflicting/javascript - removing `{{languages}}` as a side effect

SphinxKnight:
A bit overkill for just one page which had the macro, but still, now it's done


cw118 (approved):
😄

PR #5707, opened 2022-05-17
Clean conflicting/http - removing `{{languages}}` as a side effect

SphinxKnight:
One target page did not exist in `fr` so I went with translating it.


cw118 (approved):
Merci encore sphinx :)

PR #5732, opened 2022-05-19
Remove `{{HTMLVersionInline}}` from `fr`

SphinxKnight:


SphinxKnight:
I need to check the latest commits on en-US for the pages I've updated at the beginning, but at least, the macro has been removed.

cw118 (approved):
🔥 🔥

Issue #5436, opened 2022-05-06
Tracking removal of span and font tags in zh-{TW,CN}


yin1999:
It's used to track the removal of span and font tags in `l10n-zh`, as a part of preparation for `h2m` (#2474).

## sections

- [x] games (#5437)
- [x] glossary (#5435)
- [x] learn (#5585)
- [x] mdn (#5504)
- [x] mozilla (#5482)
- [x] web/api (#5734)
- [x] web/css (#5586)
- [x] web/events (#5558)
- [x] web/html (#5559)
- [x] web/http (#5505)
- [x] web/svg (#5737)
- [ ] web/other part (#5786)
- [x] webassembly (#5481)

## script

The script I've used is [here](https://gist.github.com/yin1999/7db7bea035aed5880f93ea8625a24522).

## exclude

I haven't working on those folders:

- conflicting
- orphaned

PR #5737, opened 2022-05-19
prepare for zh-{TW,CN}/web/svg h2m: removing span and font tag

yin1999:
It's a part of #5436.

There is no font tag in zh-TW.


yin1999:
checked all those modifications

PR #5782, opened 2022-05-23
replace standard-table with macros (Part 1)

yin1999:
It's a try to replace standard-table with `{{Specifications}}` macros. Using script [here](https://gist.github.com/yin1999/c718218df49661031280de7eca10ba13).

PR #5793, opened 2022-05-24
replace standard-table with macros (Part 2)

yin1999:
Continue #5782.


yin1999:
checked all those modifications.

PR #5794, opened 2022-05-24
replace standard-table with macros for zh-TW (Part 1)

yin1999:


yin1999:
Checked all those modifications.

PR #5859, opened 2022-05-27
chore: update CI to Node 16

nschonni:
Looks like Yari changed the default version so CI is complaining trying to Yarn install mdn/content


schalkneethling (approved):
Thanks, @nschonni 🎉

PR #5862, opened 2022-05-27
chore: add .markdownlint.json from mdn/content

nschonni:
I'm submitting some cleanups separately to align the content, but the CI job doesn't currently run across the repo right now


yin1999:
should we only let the workflow run on changed markdown files: https://github.com/mdn/translated-content/blob/82937239e0b147b6206bac8a7741f549635be61b/.github/workflows/markdown-lint.yml#L24-L26

.github/workflows/markdown-lint.yml
```diff
@@ -7,7 +7,7 @@ on:
     paths:
       - '*.md'
```
nschonni, line 8: @yin1999 right now this action only runs when a markdown file in the root is touched, not for any of the actual translated files

cw118:
Poking @mdn/core-yari-dev for a review :)

schalkneethling:
Hey @nschonni, thank you for opening this. Could we perhaps use the one here? https://github.com/mdn/workflows/blob/main/.github/workflows/markdown-lint.yml (https://github.com/mdn/workflows#markdown-lint)

If we need to make some changes to make it more customisable so it can be used here, that would be excellent.

nschonni:
@schalkneethling you could do the `uses:` thing like you have for some of the other actions, but that is separate from this change. Either way the current globs are incorrect and only touching the README file and none of the actual content files. That is the same issue being tracked in the discussion.
This only adds the same config file as used in the content repo, so the same rules are picked up for the current single file run, plus flags the issues for contributors that have a Markdownlint plugin in their IDE. This config file can't live in the shared repo for that reason.

schalkneethling:
> @schalkneethling you could do the `uses:` thing like you have for some of the other actions, but that is separate from this change. Either way the current globs are incorrect and only touching the README file and none of the actual content files. That is the same issue being tracked in the discussion.
> This only adds the same config file as used in the content repo, so the same rules are picked up for the current single file run, plus flags the issues for contributors that have a Markdownlint plugin in their IDE. This config file can't live in the shared repo for that reason.

💯 Thank you for the feedback.

PR #5986, opened 2022-05-29
remove summary and seoSummary from zh-CN

yin1999:
removing all summary and seoSummary classes from zh-CN. It's a preparation for markdown convertion (#2474).


yin1999:
checked all those modifications

PR #6027, opened 2022-05-30
chore: delete "conflicting" pages in fr

nschonni:
Maybe this isn't the right approach, but noticed these during the Markdownlint cleanup. Not sure if there is supposed to be some manual reconciliation then they get deleted. Feel free to close if that's the case.


nschonni:
NVM, looks like you manually review and move them back looking at the history in https://github.com/mdn/translated-content/commits/main/files/fr/conflicting

SphinxKnight:
Thanks @nschonni for bringing this up, yes we try to redirect those when possible. I'll try to tackle this in the coming days/weeks

nschonni:
No worries, I think I found a few stray ones in the regular content folders with hashes in the names. If I see those couple again, I'll open a smaller PR

PR #6040, opened 2022-05-31
Fix #5703 

SphinxKnight:
Fixes #5703 by duplicating the image from `en-US` to workaround mdn/yari#5652.


cw118 (approved):
Merci encore Sphinx !

PR #6048, opened 2022-06-01
fix: references to conflicting files in fr docs

nschonni:
Updating a few links that were hardcoded to "conflicting" copies. Should redirect to the conflicting files anyway till they are redirected


SphinxKnight (approved):
Thabks @nschonni 🙏

PR #6063, opened 2022-06-01
Update PR review companion to avoid error

cw118:
Currently after mdn/yari#6382, the PR review companion is exiting with error code 2 because it uses the option `--no-redirects`, which is now deleted. This PR removes said option from this workflow.


cw118:
@schalkneethling polite poke for a review (there's also mdn/content#16868) :)

caugner (approved):

PR #6081, opened 2022-06-03
Typofix: HTML entity in title

GKuetenang:
I saw there is a problem during translation process. The main title of the document is "Apprendre le HTML&nbsp;: guides et didacticiels" Instead of "Apprendre le HTML : guides et didacticiels". 
The problem is the &nbsp; (Space code in html).


SphinxKnight (approved):
Merci beaucoup @GustaveTsopmo ! 🙇

Issue #5879, opened 2022-05-28
AutoCorrect zh-CN TODO


huacnlee:
## SVG
- [x] https://github.com/mdn/translated-content/pull/5875
- [x] https://github.com/mdn/translated-content/pull/5962
- [x] https://github.com/mdn/translated-content/pull/5963

## JavaScript

- [x] https://github.com/mdn/translated-content/pull/5876
- [x] https://github.com/mdn/translated-content/pull/5968
- [x] https://github.com/mdn/translated-content/pull/5969
- [x] https://github.com/mdn/translated-content/pull/5970
- [x] https://github.com/mdn/translated-content/pull/5971
- [x] https://github.com/mdn/translated-content/pull/5972
- [x] https://github.com/mdn/translated-content/pull/5973
- [x] https://github.com/mdn/translated-content/pull/5974
- [x] https://github.com/mdn/translated-content/pull/5975
- [x] https://github.com/mdn/translated-content/pull/5976
- [x] https://github.com/mdn/translated-content/pull/5977
- [x] https://github.com/mdn/translated-content/pull/5978
- [x] https://github.com/mdn/translated-content/pull/5979
- [x] https://github.com/mdn/translated-content/pull/5877
- [x] https://github.com/mdn/translated-content/pull/5878

## HTTP

- [x] https://github.com/mdn/translated-content/pull/5880
- [x] https://github.com/mdn/translated-content/pull/5881
- [x] https://github.com/mdn/translated-content/pull/5882

## API

- [x] https://github.com/mdn/translated-content/pull/5883
- [x] https://github.com/mdn/translated-content/pull/5884
- [x] https://github.com/mdn/translated-content/pull/5885
- [x] https://github.com/mdn/translated-content/pull/5886
- [x] https://github.com/mdn/translated-content/pull/5887
- [x] https://github.com/mdn/translated-content/pull/5888
- [x] https://github.com/mdn/translated-content/pull/5889
- [x] https://github.com/mdn/translated-content/pull/5890
- [x] https://github.com/mdn/translated-content/pull/5891
- [x] https://github.com/mdn/translated-content/pull/5892
- [x] https://github.com/mdn/translated-content/pull/5893
- [x] https://github.com/mdn/translated-content/pull/5894
- [x] https://github.com/mdn/translated-content/pull/5895
- [x] https://github.com/mdn/translated-content/pull/5896
- [x] https://github.com/mdn/translated-content/pull/5897
[...]

PR #6097, opened 2022-06-04
AutoCorrect files/zh-cn/conflicting

huacnlee:
#5879


yin1999:
I think we are not necessary to modify those documents in `conflict`, those documents are conflicted with other documents (and will not be indexed by search engines)

yin1999:
I'm going to close this pr. If you have any other reason, we can reopen it

PR #6113, opened 2022-06-04
Update HTTP Cookie title

huacnlee:


files/zh-cn/web/http/cookies/index.html
```diff
@@ -1,5 +1,5 @@
 ---
-title: HTTP cookies
+title: HTTP Cookies
```
yin1999, line 2: I prefer `Cookie` not `Cookies`. On the other hand, the documentation needs to be updated.

yin1999 (changes requested):

yin1999 (approved):

PR #6149, opened 2022-06-06
modified the description of 1pc in CSS units

daheejo:
I found the description of 1pc is wrong and I modified it.

https://developer.mozilla.org/ko/docs/Learn/CSS/Building_blocks/Values_and_units


pje1740 (approved):
디테일한 부분 캐치 감사합니다! 고생하셨습니다 🔥

PR #6176, opened 2022-06-08
update the title for `Introduction`

awxiaoxian2020:
`介绍` is a generic expression in Chinese. We can say all the docs is "介绍". And almost all tech docs tend to translate to `简介`. See [vue docs](https://staging-cn.vuejs.org/guide/introduction.html) as an example.
We can consider `概述` too.

### DIFF
--- a/files/zh-cn/web/javascript/guide/introduction/index.html
+++ b/files/zh-cn/web/javascript/guide/introduction/index.html
@@ -1,5 +1,5 @@
 ---
-title: 介绍
+title: 简介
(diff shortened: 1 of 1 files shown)


yin1999:
I prefer `概述`, ref: <https://developer.mozilla.org/zh-CN/docs/Web/API/Document_Object_Model/Introduction>

awxiaoxian2020:
There are too many `介绍` for title or as a single word. And there are too many translations for `Introduction` in these docs. We can't replace them easily because it's hard to find all single words. close this PR.

PR #6220, opened 2022-06-11
keep example the same as renderd

yin1999:
Follow the changes I've made in mdn/content#17097.


yin1999:
Self approved for this simply modification

PR #6266, opened 2022-06-14
In line 50 , you write console.table() , i change it to console.log()

Ombre2:


SphinxKnight (changes requested):
Could you please indicate the reasons for this change?

SphinxKnight:
@Ombre2 is there any reason behind this change?

Ombre2:
Hello , i change it because console.table() give an error on the console if
it executed to the browser so the correct sintax is console.log()

Cordialement ,

Ombres

Le mar. 14 juin 2022 22:05, SphinxKnight ***@***.***> a
écrit :

SphinxKnight:
@Ombre2 thanks for the feedback, however the example works in my browser and `console.table()` seems to be fairly well supported: https://developer.mozilla.org/en-US/docs/Web/API/Console/table#browser_compatibility

Could you provide a screenshot of the error and/or some more details?

SphinxKnight:
For example, here's what I get in Firefox:
image

SphinxKnight:
Please let us know if you have details for reopening this PR.

PR #6292, opened 2022-06-15
Update index.md

Sanford284:
correct wrong word. `buttom` to `bottom`


yin1999 (approved):
Thanks a lot. Welcome aboard

PR #6320, opened 2022-06-17
suggest french translation for id:4270

ManuSquall:
Suggest a french translation for [id 4270](https://tristantheb.github.io/history-content/#4270).


SphinxKnight (approved):
Merci @ManuSquall 🥳

PR #6323, opened 2022-06-17
fix an wrong hyperlink (in zh-CN)

aixcyi:
更正一条由于语法错误导致显示不正常的超链接。


yin1999 (approved):
Thanks a lot. Welcome aboard

PR #6350, opened 2022-06-17
chore: Run Markdownlint rules on fr docs - Batch 25

nschonni:
Last batch of markdownlint cleanups. There is still an outstanding exception for the use of `<i lang="en">` and `<q lang="en">` that i'm ignoring for now


SphinxKnight:
Yes, `<i lang="en">` is used for "inline" English terms (in order to be typographically (and also semantically) correct). It's a convention we use.

SphinxKnight (approved):
Aaand done, thanks a lot @nschonni for this!

PR #6372, opened 2022-06-18
chore: Run Markdownlint rules on zh-cn docs

nschonni:
Cleaned up with .markdownlint.json config in repo


yin1999 (approved):
Thanks a lot

PR #6381, opened 2022-06-19
replace standard-table with macros for zh-TW (Part 10)

yin1999:
continue #6352.

It's the last part of this progress. Other `SpecName` and `Spec2` macros should be replaced by removing conflict documents and document sync.


irvin:
thanks

Issue #5603, opened 2022-05-12
Macro removal - Meta - deprecated macros and macro arguments


SphinxKnight:
Hello @mdn/localization-team-leads  With https://github.com/mdn/yari/pull/6252, the macros which are only used (because of legacy/history) in mdn/translated-content are marked as deprecated.

Here is a breakdown of each of them and their corresponding counts in this repo (as of `e296c15`). I hope this will make it easier for each locale's team to split the work and replace/update the pages accordingly.

Don't hesitate to ping me here or on Matrix (or to edit this directly if you have the corresponding rights) to update this list :)
N.B.

- [ ] `bug` - 745 occurrences (replace w/ `[Firefox bug $1](https://bugzil.la/$1)`)
  - [x] `es` - https://github.com/mdn/translated-content/pull/12057
  - [x] `fr` - #13605
  - [x] `ja` - All done!
  - [x] `ko` - #12007 
  - [ ] `pt-br` - 68 occurrences #14100
  - [ ] `ru` - 96 occurrences
  - [x] `zh-cn` - 1 occurrence #11892 #13257
  - [x] `zh-tw` - 39 occurrences #11893 #13257
- [x] `CanvasSidebar` - see #10516
- [ ] Arguments to `Compat` - see #11594
- [ ] `CSSAnimatedProperties` - 3 occurrences / 3 files
  - [ ] 1 file for `ja`
  - [ ] 1 file for `fr`
  - [ ] 1 file for `ko`
- [x] `CSSTutorialTOC` - 15 occurrences / 15 files
  - [x] 5 files for `de` (inactive locale) #6556
  - [x] 2 files for `es` #6557
  - [x] 1 file for `ja` #5762 
  - [x] 1 file for `ko` #5605
  - [x] 2 files for `pt-br`
  - [x] 1 file for `ru` #7340
  - [x] 3 files for `zh-cn` (#5825)
- [x] `DeprecatedGeneric` - 3 occurrences
  - [x] 1 occurrence for `es` #11585
  - [x] 2 occurrences for `ko` #11588
- [x] `DOM0` - 97 occurrences / 97 files
  - [x] 2 files for `de` (inactive locale)
  - [x] 1 file for `es` #6385
  - [x] 24 files for `fr` #5668 
  - [x] 17 files for `ja` #5770 
  - [x] 1 file for `ko` #5605
  - [x] 25 files for `pl` (inactive locale)
  - [x] 2 files for `pt-br` #8000
[...]

PR #6396, opened 2022-06-19
remove orphaned document to remove unused macro

yin1999:
Part of #5603. Remove `gecko_callout_heading` macro from zh-TW.


irvin:
dup of https://github.com/mdn/translated-content/blob/main/files/zh-tw/mozilla/firefox/releases/4/index.html

PR #6404, opened 2022-06-19
Remove disinformation about code 405

henry701:
The text says GET and HEAD cannot ever return code 405, but there is no such limitation cited on the RFC, and the full 405 page itself uses GET as an example of a disallowed method.


josielrocha (approved):

josielrocha:
Obrigado pela contribuição @henry701!

PR #6421, opened 2022-06-21
Update the chinese translation of `Set`

jynxio:
## 概述
修改了 `Set.prototype.delete(value)` 的汉语翻译，原因是：
- 原翻译有些晦涩
- 原翻译与英语原文不一致

## Overview
Update the chinese translation of `Set.prototype.delete(value)`, the reason is:
- The old translation was a bit obscure.
- The old translation does not mean the same as the English original


yin1999 (approved):
Thanks a lot.

PR #6426, opened 2022-06-21
Remove `{{anch}}` and its mentions for `es`

cw118:
Part of #4614, this is a small PR to just quickly remove the macro :)


cw118:
Politely poking @mdn/yari-content-es, almost all `{{anch}}`s are gone so we are hoping to remove these soon, then the macro can be deleted :)

Graywolf9 (approved):
Thank you very much, it must have escaped me XD

cw118:
Thanks for the review @Graywolf9 :)

PR #6427, opened 2022-06-21
Remove `{{anch}}` macro for `ja`

cw118:
Part of #4614 (pinging @mdn/yari-content-ja so that we can remove all `{{anch}}`s fast and soon :) )


mfuji09 (approved):
Ok, thank you!

PR #6428, opened 2022-06-21
Remove `{{anch}}` for korean locale

cw118:
Part of #4614, poking @mdn/yari-content-ko so that we can merge this soon and remove all `{{anch}}`s from `translated-content` :)


hochan222 (approved):
LGTM :)

PR #6469, opened 2022-06-24
Fix #6467

### LINKED ISSUE #6467: Content bug: <menu> element page is outdated in French
## What page(s) did you find the problem on?

https://developer.mozilla.org/fr/docs/Web/HTML/Element/menu

## Specific page section or heading?

Entire page

## What is the problem?

The content of the page doesn't match with the English version at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu, it is probably outdated

## What did you expect to see?

French translation of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu.

SphinxKnight:
An update to match the `en-US` state of things

### DIFF
--- a/files/fr/web/html/element/menu/index.md
+++ b/files/fr/web/html/element/menu/index.md
@@ -1,128 +1,73 @@
 ---
-title: <menu>
+title: "<menu> : l'élément de menu"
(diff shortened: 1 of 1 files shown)


cw118 (approved):
💯

PR #6470, opened 2022-06-24
Fix Chinese translation typo

jishen027:
Here is a Chinese Typo, which is not "gamepadconnected world(世界)", should be "gamepadconnected event(事件)".


yin1999 (approved):
Thanks a lot. Welcome aboard

PR #6569, opened 2022-06-28
fix: cache with specific python version

### LINKED ISSUE #6568: PR review companion failed to run
Recently, there are many `PR review companion` workflows which are failed to run.

It seems to be a `cache` problem. The `actions/cache` will cache `.local` with a specific key. But the `key` will be the same when the python version may be different. Consider those two action jobs:

- https://github.com/mdn/translated-content/runs/7088612871?check_suite_focus=true#step:5:6 (python 3.8.12)
- https://github.com/mdn/translated-content/runs/7091308837?check_suite_focus=true#step:5:6 (python 3.8.13)

With different python version but the `actions/cache` will restore the cache used in `python 3.8.12`.

And see snok/install-poetry#82. I think we should add `python version` to the `actions/cache` action's key to avoid this.

yin1999:
Fixes #6568

### DIFF
--- a/.github/workflows/pr-review-companion.yml
+++ b/.github/workflows/pr-review-companion.yml
@@ -77,7 +77,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/.local
-          key: dotlocal-${{ runner.os }}-${{ hashFiles('.github/workflows/pr-review-companion.yml') }}-${{ steps.setup-python.outputs.python-version }}
+          # the trailing number is used to increase for getting
+          # a different cache key when this file changes
+          key: dotlocal-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-0
 
       - name: Install Python poetry
         uses: snok/install-poetry@v1.3
--- a/.github/workflows/pr-review-companion.yml
+++ b/.github/workflows/pr-review-companion.yml
@@ -25,7 +25,7 @@ jobs:
 
     steps:
       - name: 'Download artifact'
-        uses: actions/github-script@v5
+        uses: actions/github-script@v6
         with:
           script: |
             var artifacts = await github.rest.actions.listWorkflowRunArtifacts({
@@ -63,11 +63,12 @@ jobs:
 
       - uses: actions/checkout@v3
         with:
-           repository: mdn/yari
-           path: yari
+          repository: mdn/yari
+          path: yari
 
       - name: Install Python
-        uses: actions/setup-python@v3
+        id: setup-python
+        uses: actions/setup-python@v4
         with:
           python-version: "3.8"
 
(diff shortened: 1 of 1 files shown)


yin1999:
Tested: https://github.com/yin1999/translated-content/runs/7103501129?check_suite_focus=true#step:6:4

The cache key will be specified with python version, and `snok/install-poetry` could find `python` now.

.github/workflows/pr-review-companion.yml
```diff
@@ -76,10 +77,10 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/.local
-          key: dotlocal-${{ runner.os }}-${{ hashFiles('.github/workflows/pr-review-companion.yml') }}
+          key: dotlocal-${{ runner.os }}-${{ hashFiles('.github/workflows/pr-review-companion.yml') }}-${{ steps.setup-python.outputs.python-version }}
```
yin1999, line 80: I'm not sure about `hashFiles('.github/workflows/pr-review-companion.yml')` it seems to be an empty string.

yin1999:
pin @mdn/core-dev for a review

yin1999:
Hi @schalkneethling, could you have a review on this. This problem is a bit annoying.

.github/workflows/pr-review-companion.yml
LeoMcA: Yeah, we don't checkout the repo in this workflow, so there's no file to cache - I think that's ok, we can increment a number at the end of the cache key if we make any breaking changes. So, can you remove this part?

LeoMcA: We can remove it from the venv cache key below, as well. Perhaps we also want to include the python version in there, I don't think the poetry lockfile includes it.

LeoMcA (approved):
looks good if we remove those empty hashes, thanks!

we should make the same changes to this file in the `content` repo - would you like to open a similar PR there, or should I with the magic of [`Co-authored-by`](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)?

.github/workflows/pr-review-companion.yml
yin1999: thanks for answering this, I've add a trailing `0` with a comment to explain it.

yin1999:
> looks good if we remove those empty hashes, thanks!
> 
> we should make the same changes to this file in the `content` repo - would you like to open a similar PR there, or should I with the magic of [`Co-authored-by`](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)?

Ok, I'll make the same changes there.

LeoMcA (approved):

LeoMcA:
Thank you!

PR #6574, opened 2022-06-29
Remove RTL direction

donaldxdonald:


yin1999 (approved):
Thanks a lot. Welcome aboard

PR #6583, opened 2022-06-29
Update core dev team name

cw118:
The `core-yari-dev` team no longer exists, so this PR updates the codeowners. 

N.B. Github says there are errors because it seems that `core-dev` and `core-yari-content` aren't explicitly given access to the `translated-content` repo (cc. @schalkneethling / @Rumyra)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,7 +9,7 @@
 # ----------------------------------------------------------------------------
 # DEFAULT OWNERS
 # ----------------------------------------------------------------------------
-*    @mdn/core-yari-dev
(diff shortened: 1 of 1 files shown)


schalkneethling:
> The `core-yari-dev` team no longer exists, so this PR updates the codeowners.
> 
> N.B. Github says there are errors because it seems that `core-dev` and `core-yari-content` aren't explicitly given access to the `translated-content` repo (cc. @schalkneethling / @Rumyra)

Thanks for doing this @cw118, I added the two teams to `translated-content`.

LeoMcA (approved):
lgtm, thanks!

PR #6602, opened 2022-06-30
Fix an issue with <a> tags

remyhuang03:
When <a>tags are used to link to pages for something like <iframe>, <canvas>, etc., there are some redundant words enclosed in them.


yin1999 (approved):

yin1999:
Thanks a lot. Welcome aboard

PR #6622, opened 2022-07-01
update wrong translation of ES2015 Accessors

KevynTD:
The last portuguese translation forgot to remove the 'function' part in the code, this syntax is wrong in JS:

old:
 >   **get function()** { return 10; },
 >   **set function(value)** { console.log('setting `o.bar` to', value); }

new corrected:
>    **get()** { return 10; },
>    **set(value)** { console.log('setting `o.bar` to', value); }


nathipg (approved):
Thank you, nice job!

PR #6705, opened 2022-07-06
(ru) Fix code example in Web/JavaScript/Guide/Working_with_Objects

vtivadze:
Car owner is an object itself.
So to find out owner name we have to get 'name' property of object 'owner' and not simply object 'owner' itself.


sashasushko (approved):

sashasushko:
@vtivadze спасибо!

PR #6720, opened 2022-07-06
Remove conflicting PWAs pages

SphinxKnight:
I updated the (short) landing page in the process.
The [introduction page](https://developer.mozilla.org/fr/docs/web/progressive_web_apps/introduction) would benefit an update as well, but I'll leave this for another PR.


cw118 (approved):
🚀

PR #6733, opened 2022-07-07
Remove conflicting Web API pages - W

SphinxKnight:
Given the number of pages, no update for target pages on this one.
I should file an issue to update the various `_event` pages of  `Window`


cw118 (approved):
🤗

PR #6736, opened 2022-07-07
Translation of the keyword list of the position property.

jeremarques:
In the **Syntax** section, I have translated the keyword list and their descriptions of the `position` property.


josielrocha (approved):

josielrocha:
Obrigado pela contribuição @jeremarques!

PR #6737, opened 2022-07-07
Fixing image paths

TedMartins:
Someone translated the images path, this can be confusing for those just starting to learn. This fix fixes that.


nathipg (approved):
Good one!

PR #6781, opened 2022-07-11
Remove Web/API conflicting pages as of 2022-07-11

SphinxKnight:
Just removing/redirecting here. Same comment as earlier, Web/API event pages would benefit from a larger update :/


cw118 (approved):
Merci Sphinx :)

PR #6805, opened 2022-07-12
cache venv with specific python version

### LINKED ISSUE #6804: `poetry venv` in PR review companion may not be correctly cached
## problem

The `poetry venv` in `PR review companion` is also cached by `actions/cache@v3` in path `yari/deployer/.venv`. But now it's cached with a key without specific python version which may broken the cached `poetry venv` (when the python version is updated).

This is the same as the previous question: #6568

This can be confirmed by those actions:

- https://github.com/mdn/translated-content/runs/7231479302?check_suite_focus=true#step:10:11 (with python version changed and the restored venv is broken)
- https://github.com/mdn/translated-content/runs/7232127850?check_suite_focus=true#step:8:11 (lock file changed so cache with new key)
- https://github.com/mdn/translated-content/runs/7233198476?check_suite_focus=true#step:10:13 (cached and skip `poetry install`)

## resolve

We should add python version to the `cache key`

yin1999:
Fixes #6804.

When python version changed, the cached venv may be broken.

### DIFF
--- a/.github/workflows/pr-review-companion.yml
+++ b/.github/workflows/pr-review-companion.yml
@@ -92,7 +92,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: yari/deployer/.venv
-          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}-${{ hashFiles('.github/workflows/pr-review-companion.yml') }}
+          # the trailing number is used to increase for getting
+          # a different cache key when this file changes
+          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}-${{ steps.setup-python.outputs.python-version }}-0
 
       - name: Install poetry dependencies
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'


.github/workflows/pr-review-companion.yml
```diff
@@ -92,7 +92,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: yari/deployer/.venv
-          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}-${{ hashFiles('.github/workflows/pr-review-companion.yml') }}
```
yin1999, line 95: remove `${{ hashFiles('.github/workflows/pr-review-companion.yml') }}` just follow #6569

yin1999:
poke @LeoMcA for a review. If it's ok, I'll make the same change on `mdn/content`

awxiaoxian2020:
> poke @LeoMcA for a review. If it's ok, I'll make the same change on `mdn/content`

I want to know whether it fixes https://github.com/mdn/content/issues/18412 (if you do the same change in it)

yin1999:
> I want to know whether it fixes [mdn/content#18412](https://github.com/mdn/content/issues/18412) (if you do the same change in it)

not at all

LeoMcA (approved):
looks great

> If it's ok, I'll make the same change on mdn/content

thank you!

PR #6821, opened 2022-07-13
fix a display problem

remyhuang03:
Some `<code>` tags were misused in the original html file. (Below is a screenshot for the original page)
image


yin1999 (approved):
Thanks a lot

PR #6889, opened 2022-07-15
Rename the file Glossary/Safe/HTTP

linlinzzo:
https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP didn't have Chinese translations, so I renamed the file.

according to https://github.com/mdn/content/blob/main/files/en-us/glossary/safe/http/index.md


yin1999:
we cannot just rename the file, if you want to move a document, please using command `yarn content move` (<https://github.com/mdn/content#fundamental-concepts>)

yin1999:
I've created #6898 to resolve this, thanks for raising this.

PR #6898, opened 2022-07-16
zh-CN glossary item move to follow en-US

yin1999:
Reference: #6889


yin1999:
`Rebase and merge` ony

PR #6915, opened 2022-07-17
replace standard-table with macros for zh-CN (Part 30)

yin1999:
continue #6899


files/zh-cn/web/css/flex-basis/index.html
```diff
@@ -178,22 +178,7 @@ <h3 id="结果">结果</h3>
 
 <h2 id="规格">规格</h2>
```
yin1999, line 179:
```suggestion
<h2 id="规范">规范</h2>
```

PR #6944, opened 2022-07-18
remove redundant li tags

nanvon:


yin1999 (approved):
Thanks a lot. Welcome aboard

PR #6985, opened 2022-07-19
fix: wrong code example

abitwhy:
Because of wrong punctuation.

You can check it on [regex101](https://regex101.com/r/0boKWf/1).


yin1999 (approved):
Thanks

PR #7029, opened 2022-07-20
Remove extra styles for dash

xel23:


Saionaro:
@xel23 Hello! Thank you for your contribution!

PR #7043, opened 2022-07-21
Fix embeded example in pt_br Dialog element

rodgco:
The ID of the Advanced Example was translated, but the reference to embed it wasn't.


josielrocha (approved):

josielrocha:
Obrigao pela contribuição @rodgco!

PR #7050, opened 2022-07-21
zh-cn/conflicting, replace unicode space to ascii space

fwqaaq:


yin1999:
Hi @Jack-Zhang-1314, the conflicting folder is used to place conflict document (by github action). We should remove those documents if it's ok. So it's not necessary to convert those documents to markdown.

I'm going to close this pr.

Jack-Zhang-1314:
Sorry, you did right

PR #7051, opened 2022-07-21
zh-cn/learn, replace unicode space to ascii space

fwqaaq:
The `.html` file in the zh-cn have all replaced unicode space with ascii space.


yin1999 (approved):
checked locally

Jack-Zhang-1314:
There should be no more in the zh-cn.

Jack-Zhang-1314:
Previous translation issues need to be addressed. I'll improve it.
<https://github.com/mdn/content/issues/17799#event-7037357843>

Jack-Zhang-1314:
But english documnet has not been changed, do we need to change it in advance?

yin1999:
> But english documnet has not been changed, do we need to change it in advance?

[`mdn/yari`](https://github.com/mdn/yari/actions/workflows/prod-build.yml) builds those documents and serving them on aws s3 at 00:00 UTC everyday. So the changes may not be applyed immediately, just follow the [raw document](https://github.com/vvswps/content/blob/main/files/en-us/web/api/textdecoderstream/index.md) in mdn/content

Jack-Zhang-1314:
Oh, I see. I will improve it.

PR #7079, opened 2022-07-23
ci: use built-in cache

yin1999:
Follow mdn/content#12313, setup-node will cache global cache. So the yarn cache can be reused between different Node.js version.


yin1999:
poke @mdn/core-dev for a review

yin1999:
Close in favor of #7709

PR #7130, opened 2022-07-25
Remove unnecessary words

koh1project:
Remove unnecessary word that is not used in the original sentence.


potappo (approved):
修正ありがとうございます。OK です。

PR #7144, opened 2022-07-26
Remove `conflicting` pages as of 20220726 w/ an update

cw118:
This PR removes some conflicting pages that were recently created, and updates `load_event`


SphinxKnight:
Merci @cw118 :)

PR #7330, opened 2022-08-03
remove all `{{Spec2}}` macros from zh-CN

yin1999:


files/zh-cn/web/api/mouseevent/initmouseevent/index.html
```diff
@@ -93,29 +93,13 @@ <h3 id="JavaScript">JavaScript</h3>
 
 <h2 id="标准">标准</h2>
```
yin1999, line 94:
```suggestion
<h2 id="规范">规范</h2>
```

PR #7334, opened 2022-08-03
Remove `{{css_key_concepts}}` from `fr`

SphinxKnight:
I didn't check the status of this macro in mdn/content and mdn/yari but I'm pretty sure it's not worth it for `fr` for a bunch of links in two pages.


cw118 (approved):
Et voilà une macro de moins 🙏

PR #7340, opened 2022-08-03
Remove last use of {{CSSTutorialTOC}} macro

queengooborg:
This PR removes the very last use of the `{{CSSTutorialTOC}}` macro from MDN content.


queengooborg:
I'm going to self-merge this because it is the very last instance of this macro in all of MDN and I've confirmed simple removal is appropriate based upon analyzing other translations of this document.  Happy to follow up on this if requested!

PR #7341, opened 2022-08-03
Remove last use of {{outdated}} macro

queengooborg:
This PR removes the very last use of the `{{outdated}}` macro from MDN content.


SphinxKnight:
Thanks @queengooborg :) let's get rid of this one quickly :)

Issue #7319, opened 2022-08-03
FR: broken external links


OnkarRuikar:
All external broken links in `mdn/content` have been fixed: https://github.com/mdn/content/pull/19085

While working on the PRs I've found following broken links in `fr` locale:

- [ ] https://anlexn.github.io/shared-worker-mdn/
- [ ] www.affirmationsforpositivethinking.com
- [ ] https://github.com/mdn/panner-node
- [ ] https://mail.mozilla.org/listinfo/dev-addons
- [ ] https://mdn.mozillademos.org/
- [ ] https://dxr.mozilla.org/mozilla-central/source/content/xslt/public/nsIXSLTProcessor.idl
- [ ] {{Source("content/xml/tests/load/", "Voir aussi l'extrait de chargement")}}
- [ ] https://dxr.mozilla.org/mozilla-central/source/netwerk/protocol/about/public/nsIAboutModule.idl
- [ ] http://devedge.netscape.com/2002/de
- [ ] http://docs.activestate.com/komodo/8.5/
- [ ] https://developer.android.com/sdk/index.html
- [ ] https://discuss.atom.io/categories
- [ ] https://jinja.pocoo.org/docs/2.9/
- [ ] modern.ie
- [ ] http://smush.it/
- [ ] https://opendevicelab.com/
- [ ] thimble.mozilla.org
- [ ] www.aha-soft.com/free-icons/free-retina-icon-set
- [ ] www.ucweb.com/ucbrowser
- [ ] www.weba11y.com/blog/2016/04/22/screen-reader-support-for-new-html5-section-elements
- [ ] www.xmlpitstop.com
- [ ] http://webglstats.com
- [ ] http://weblogs.mozillazine.org/roc/archives/2006/06/the_future_is_now.html
- [ ] http://weblogs.mozillazine.org/roc/archives/2010/08/mozrequestanima.html

In the process we've removed all `Source`[[ref](https://github.com/mdn/content/pull/18872)] and `Interwiki` macros.
***
**Tips:**
- To find files for the links without using IDE use GitHub Code search feature, e.g.
 [`https://github.com/mdn/translated-content/search?q=modern.ie`](https://github.com/mdn/translated-content/search?q=modern.ie)
- For links that are only part of the discussion put them in code fences(e.g. `www.example.com`) to avoid creating actual hyper links for them on the site. [[ref](https://github.com/mdn/content/pull/18917)]

SphinxKnight:
[...]

PR #7358, opened 2022-08-04
Remove dead links (part 2 of #7319)

SphinxKnight:
Updated the existing page in the process
Part of #7319


cw118 (approved):
Merci encore :)

PR #7362, opened 2022-08-04
fetch: remove unrelevant section

RomainLanz:
Hey there! 👋🏻 

The `Fetch API` is no longer in experimental for years, we can safely remove this section (which is not in the English version).


SphinxKnight (approved):
Yess please :) Thanks @RomainLanz! 🚀

PR #7369, opened 2022-08-04
Replace 0xA0 (non-breaking space) characters with normal spaces (zh-*)

queengooborg:
This PR replaces 0xA0 (non-breaking space) characters with normal spaces throughout the Chinese locale, both Simplified and Traditional.


Rumyra (approved):

yin1999:
Thanks, we have replaced a lot of Unicode space in l10n-zh, but we may ignore some paths.

PR #7372, opened 2022-08-04
Convert /games folder to Markdown (es)

queengooborg:
This PR converts the documents in the `/games` folder from HTML to Markdown for the Spanish locale.  Three files were skipped as they have no translations in them and will be removed in another PR.


JuanVqz (approved):
@queengooborg so nice! Thank you for contributing

JuanVqz:
I've noticed some minor issues, we are not adding the tags section anymore because it's not used by the translated languages, and our URL paths should start with `/es` not with `/en-US` but we can check that in the future.

Again thank you 🚀

queengooborg:
> we are not adding the tags section anymore

Oh I didn't realize that, thanks for the head's up!  I'd be happy to put together a script to remove the tags frontmatter keys from translated content, unless there's already one ready to go?

> our URL paths should start with `/es` not with `/en-US`

Oops, were there some that I had missed?

I'd be happy to open up some PRs to fix these, or I can open an issue to track this for all locales in case there are other instances lurking around, whichever is easier?

JuanVqz:
> > we are not adding the tags section anymore
> 
> 
> 
> Oh I didn't realize that, thanks for the head's up!  I'd be happy to put together a script to remove the tags frontmatter keys from translated content, unless there's already one ready to go?


There is not, open it please.


> 
> 
> 
> > our URL paths should start with `/es` not with `/en-US`
> 
> 
> 
> Oops, were there some that I had missed?
> 
> 
> 
> I'd be happy to open up some PRs to fix these, or I can open an issue to track this for all locales in case there are other instances lurking around, whichever is more preferable?

I'm speaking up about Spanish files start with `/es/path/to/file` but French files start with `/for/path/to/file`, it means this apply only for Spanish files.

Also, are you aware that we are introducing a new metadata, here you have more info https://github.com/orgs/mdn/discussions/129

Any further question I'm happy to help 🚀

PR #7415, opened 2022-08-07
fix redirects for l10n-fr

yin1999:
see redirects error in fr: https://github.com/mdn/translated-content/runs/7710006136?check_suite_focus=true#step:7:22

For the document has been created by #7305


SphinxKnight (approved):
Thanks @yin1999, sorry for the mishap here. I need to check why I hadn't seen that on the original PR :(

PR #7376, opened 2022-08-05
Convert /glossary folder to Markdown (es)

queengooborg:
This PR converts part of the `/glossary` folder to Markdown for the Spanish locale.  Note: some files were removed, either because they were deleted upstream or are too far out of sync with upstream.

_Disclaimer: I am an English content writer rather than a localization expert, and as such I strive to keep my changes purely editorial._

PR #7436, opened 2022-08-08
Restore accidentally-removed #Quick_links sections (es)

queengooborg:
This PR is a follow-up to #7376 that restores the `#Quick_links` `<section>` elements.  During Markdown conversion, I didn't realize these sections were still used for the sidebars.


Graywolf9:
Hi! this is not going to be necessary I reverted the changes to can do a correct manual review of the content as we talk before about larger PRs with significant changes without human supervision

Thank you so much

PR #7490, opened 2022-08-10
Update index.html

EduardooPV:
Removes two repeated words (same same) in "Packing with prototypes and static methods" session


josielrocha (approved):

josielrocha:
Obrigado pela contribuição @EduardooPV!

Issue #5599, opened 2022-05-12
Macro removal - `{{Draft}}` deprecation


SphinxKnight:
Hello @mdn/localization-team-leads, I'm opening this one to follow the deprecation on https://github.com/mdn/yari/pull/6219.

There is currently https://github.com/mdn/translated-content/pull/3085 which is opened but recent work on macro removal show that localization teams were blazing fast when tackling this kind of task in separate PRs.

I assume the macro call can simply be replaced by nothing, you may want to delete the calling page if its state is too `drafty`.

SphinxKnight:
https://github.com/mdn/translated-content/pull/5598 for `fr`

cw118:
#5598 PR is merged, `fr` taken care of by @SphinxKnight 🎉

yin1999:
`l10n-zh` resolved in #5600 and #5601.

mfuji09:
Created PR #5693 for 'ja'.

yin1999:
use `rg "\{\{\s*draft" -il .` and find those remaining usage:

- [ ] files/ko/web/api/audioparammap/index.md
- [ ] files/ru/web/security/securing_your_site/index.html
- [ ] files/es/web/security/securing_your_site/index.html
- [ ] files/es/web/css/_doublecolon_-webkit-scrollbar/index.html
- [ ] files/es/games/tutorials/html5_gamedev_phaser_device_orientation/index.md
- [ ] files/pt-br/web/api/webvr_api/index.html
- [ ] files/pt-br/web/security/securing_your_site/index.html
- [ ] files/ru/web/api/document/registerelement/index.html
- [ ] files/es/web/api/document/registerelement/index.html
- [ ] files/es/web/api/websockets_api/writing_websocket_client_applications/index.html
- [ ] files/pl/mozilla/firefox/releases/2/updating_extensions/index.html
- [ ] files/ko/web/accessibility/aria/roles/dialog_role/index.html

SphinxKnight:
ouch, nice find

queengooborg:
I've just merged #7302, now only one file remains!

PR #7518, opened 2022-08-10
remove last `{{Draft}}` macro from l10n-pl

yin1999:
Part of #5599


queengooborg (approved):
Thanks, this is LGTM!

Issue #7566, opened 2022-08-12
`{{Link}}` macro removal


yin1999:
The `{{Link}}` macro which is [going to be depreciated](https://github.com/mdn/yari/pull/6865) and should be [removed from content](https://github.com/orgs/mdn/discussions/187).

I've used `rg` to find all reference:

- de
  - [x] files/de/glossary/javascript/index.html #7567

- pl
  - [x] files/pl/glossary/javascript/index.html #7567

- ja #8292
  - [x] files/ja/glossary/json/index.html
  - [x] files/ja/glossary/javascript/index.html
  - [x] files/ja/glossary/localization/index.html

- ko
  - [x] files/ko/glossary/json/index.html https://github.com/mdn/translated-content/pull/7577

- fr - https://github.com/mdn/translated-content/pull/7593
  - [x] files/fr/glossary/json/index.md
  - [x] files/fr/glossary/javascript/index.md
  - [x] files/fr/web/api/document/stylesheets/index.md
  - [x] files/fr/glossary/localization/index.md

- ru #7671
  - [x] files/ru/glossary/json/index.html
  - [x] files/ru/glossary/javascript/index.html

- es
  - [x] files/es/glossary/primitive/index.html #8322
  - [x] files/es/glossary/javascript/index.html #8322
  - [x] files/es/web/javascript/typed_arrays/index.html #8611

- zh
  - [x] files/zh-tw/glossary/json/index.md #7565
  - [x] files/zh-tw/glossary/javascript/index.md #7565
  - [x] files/zh-cn/glossary/json/index.md #7565
  - [x] files/zh-cn/glossary/javascript/index.md #7565

- pt-br
  - [x] files/pt-br/glossary/javascript/index.html #7595

MrBrain295:
Should this be added to #5603?

yin1999:
> Should this be added to #5603?

Let's leave this alone, or we should re-add some locale tags to that issue again.

MrBrain295:
All occurrences seem to have been removed.

yin1999:
> All occurrences seem to have been removed.

Thanks for notifying this, let's remove it from mdn/yari.

PR #7567, opened 2022-08-12
remove `{{Link}}` macro from frozen locale

yin1999:
Part of #7566


queengooborg (approved):
Thanks, this is LGTM!

PR #7597, opened 2022-08-12
Portuguese pronominal mistake fix

vferraz7:
It was wrote "Um promise" twice and all other times "Uma promise".


josielrocha (approved):

josielrocha:
Obrigado pela contribuição @victor-ferraz7!

PR #7601, opened 2022-08-13
convert to markdown for web/http zh-CN

yin1999:


queengooborg:
Thanks for your hard work on this!  Just to check, are you using the conversion script from https://github.com/mdn/markdown with the latest commit?  I happened to notice that some of the `en-US` locale links remained, which the updated script is supposed to handle, and the outdated script from Yari (or even before my commits were merged) can cause data loss around `<dt>` elements!

yin1999:
> Thanks for your hard work on this! Just to check, are you using the conversion script from https://github.com/mdn/markdown with the latest commit? I happened to notice that some of the `en-US` locale links remained, which the updated script is supposed to handle, and the outdated script from Yari (or even before my commits were merged) can cause data loss around `<dt>` elements!

Thanks, I'll have a try.

PR #7639, opened 2022-08-13
remove html file

hochan222:
I deleted the html file because the html and md files existed at the same time.


hochan222:
CC: @queengooborg

caugner:
The duplicate HTML file has meanwhile been deleted in 30c03a31da7e52bbe697ab090e42bf6edcc8d161.

PR #7673, opened 2022-08-15
chore: remove link titles in FR

nschonni:
These were removed on the English side since they were mostly just duplicates of the link text and came from the old wiki inline values


cw118 (approved):
Thanks very much @nschonni !

PR #7713, opened 2022-08-16
translate WebExtensions/user_interface/Notifications (zh-cn)

daidr:


yin1999 (changes requested):

yin1999:
Please remove the image file, it could fallback to en-US programly

Issue #7683, opened 2022-08-15
Remove duplicated frontmatter keys from zh-{CN, TW}


yin1999:
Part of #7412

sections to remove from:

- [x] games #7682
- [x] glossary #7706
- [x] learn #7729 #7721
- [x] mdn #7732 #7744
- [x] mozilla #7761 #7744
- [x] web #7760 #7832 #7831 #7830 #7829
- [x] webassembly #7729 #7720

Jack-Zhang-1314:
#7760

yin1999:
Done.

PR #7829, opened 2022-08-20
remove duplicated frontmatter keys from web/api zh-TW

yin1999:
Part of #7683


yin1999 (commented):
Checked all those files

PR #7910, opened 2022-08-22
update links for zh-CN: `en-US` -> `zh-CN`

awxiaoxian2020:
use `/en-US` -> `/zh-CN`


yin1999:
Hi @awxiaoxian2020, could you provide the script you have used to process this?

awxiaoxian2020:
> Hi @awxiaoxian2020, could you provide the script you have used to process this?

I just use a replacement `/en-US` -> `/zh-CN` and make it case senstive in vscode...

yin1999:
> I just use a replacement `/en-US` -> `/zh-CN` and make it case senstive in vscode...

This is not reliable, some `/en-US` may not appear in the link. And there are too many files, it's a bit difficult to review all of those changes.

Could you try to parse markdown file and replace string only in links?

awxiaoxian2020:
> > I just use a replacement `/en-US` -> `/zh-CN` and make it case senstive in vscode...
> 
> This is not reliable, some `/en-US` may not appear in the link. And there are too many files, it's a bit difficult to review all of those changes.
> 
> Could you try to parse markdown file and replace string only in links?

I have no experience about it. 

But I use a different way to replace(`](/en-US`) and git diff and get the following result:


Now we have no reasons to remain any `/en-US` in our docs on my side, which potential errors can be corrected easily.
If you think it is not reliable, I can just close it :)

yin1999:
In some live sample, we have used some image from en-US, so the uri will be `/en-us/docs/...`, the changes you have made will break this

yin1999:
Please replace `](/en-US/` and `](</en-US/` and do not replace widely

yin1999:
> Please replace `](/en-US/` and `](</en-US/` and do not replace widely

after that, you can replace `href="/en-US/`. There won't be too many files.

yin1999:
Hi @awxiaoxian2020, if you want to help with this, please try [this yari branch](https://github.com/yin1999/yari/tree/update-to-locale). You can run `yarn tool fix-flaws --locale zh-CN` to replace `/en-US/` and `/en/` with `/zh-CN/` only for links.

Note that, please spread the fixes into multiple PRs (see the number of changed files, less than 200).

PR #7920, opened 2022-08-22
Remove dead links - Part 4 of #7319

SphinxKnight:
This one is for the good ol' http://devedge.netscape.com/2002/de
(DevEdge was the first name of MDN)


cw118 (approved):
Merci beaucoup 🎉

PR #7924, opened 2022-08-22
Typofix for `<image>`

SphinxKnight:
A typo for `cross-fade`/`cross-face` was spotted earlier and reported over Matrix.
This is an update for the whole page.


cw118 (approved):
💯

PR #7960, opened 2022-08-23
Add Content-Length for Simple Response Header

rwv:
Content-Length was not part of the original set of safelisted response headers
See Also https://github.com/whatwg/fetch/pull/626


yin1999 (approved):
Hi @rwv, Thanks for your help.

And we are actually translating the english documents only. But this one in english are now a redirect definition: <https://developer.mozilla.org/en-US/docs/Glossary/Simple_response_header>

We should try to sync the changes. But for the small change, we can let it go for first

Issue #5618, opened 2022-05-13
Replace old-style compatibility/specification tables with new macro


queengooborg:
In mid-2019, we have migrated all of the compatibility table information over to the [BCD project](https://github.com/mdn/browser-compat-data).  Recently, we have also transitioned to using BCD to generate the specifications table.

When I went and marked the old `{{CompatXXX}}` macros as deprecated in Yari, I happened to notice that there are a LOT of translations that are still using the old-style compatibility tables, which means readers may be getting outdated compatibility information.  We should go through and replace all of these tables with ones generated by the new `{{Compat}}` macro.  While we're at it, we should also replace the specifications tables with the `{{Specifications}}` macro as well.

Steps to fix pages:
- Find pages that use any of the following macros: ~`CompatAndroid`~, ~`CompatChrome`~, ~`CompatGeckoDesktop`~, ~`CompatGeckoMobile`~, ~`CompatibilityTable`~, ~`CompatIE`~, ~`CompatNo`~, ~`CompatOpera`~, ~`CompatOperaMobile`~, ~`CompatSafari`~, ~`CompatUnknown`~, ~`CompatVersionUnknown`~, `spec`, `spec2`, `SpecName`
- Replace all content in the "Specifications" section with `{{Specifications}}`
- Replace all content in the "Browser Compatibility" section with `{{Compat}}`

P.S. BCD has an open issue regarding how to handle localization for notes.  We're open to suggestions, see https://github.com/mdn/browser-compat-data/issues/114 for more!

yin1999:
For replacing the specifications tables, I've using `<table class="standard-table">(?:[\s\S\r]*?)</table>` to match and replace with `{{Specifications}}` if existed in `en-US`. The script I've used is [here](https://gist.github.com/yin1999/c718218df49661031280de7eca10ba13) (in Golang, really simple, may not deal with all the cases).

And please check the modifications made by this script, because of the usage of `standard-table` in other tables.

yin1999:
Hi @queengooborg, I found that yari will merge metadata from en-US when render a macro in l10n documents, so whether it's ok to remove `browser-compat` metadata from l10n documents (is there any other usage of this metadata)?

see: https://github.com/mdn/yari/blob/5d8d88976164b6f653c3f2de688f3cd76f8537bb/kumascript/index.js#L56-L69

queengooborg:
[...]

PR #7973, opened 2022-08-24
Remove old compat macros from `fr`

SphinxKnight:
Only one page was having such macros left. This is an update of it.
(I went with removing some Fx only stuff when there was no doc linked)

see #5618


cw118 (approved):
🤗

PR #7974, opened 2022-08-24
Update Glossary pages with old spec macros - `fr` - Part of #5618

SphinxKnight:
This is an "update" PR for Glossary pages which contained macros targeted by #5618


cw118 (approved):
Merci pour tout ce travail Sphinx :)

PR #7976, opened 2022-08-24
Remove old spec macros - Web/API/A* - Part of #5618

SphinxKnight:
Part of #5618. This is not an update PR but a "dumb" replace


cw118 (approved):
Parfait, merci :)

PR #7989, opened 2022-08-25
Create index.md for Regexp Source and fixing grammar errors

B-e-sa:


cw118 (changes requested):
Hi @B-e-sa, could you remove the changes made to `l10n-fr`, `l10n-ja` and `l10n-zh` files from this PR please? If possible, it would be best to create a branch on your fork before modifying pages instead of editing on the `main` branch.

Let me know if help is needed :)

PR #7999, opened 2022-08-25
Remove `{{DOM0}}` from `pt-BR`

SphinxKnight:
see https://github.com/mdn/translated-content/issues/5603


SphinxKnight:
that's what happens when you don't switch back to `main` :/

Issue #4579, opened 2021-03-18
Remove unwanted macros


escattone:
We'd like to remove the following macros from all documents (English, non-English, and archived):

- [x] `fx_minversion_header`
- [x] `fx_minversion_inline`
- [ ] `gecko_minversion_header`
- [ ] `gecko_minversion_inline`
- [x] https://github.com/mdn/translated-content/issues/5170
- [x] `languages`

SphinxKnight:
Tagging all active locales "for reach", if macros are already absent from any, please let me know so that I can remove the right tags.

yin1999:
@SphinxKnight removed all those macros in `l10n-zh` via #4728 #4729

SphinxKnight:
Tbh, #5603 arrived later than this one (2021 ;))

mfuji09:
The all macros have been removed in the `ja` locale.

yin1999:
Done for `ko` in #8096.

Graywolf9:
I think this is done in `es`

doing: 
````
rg -i "\{\{\s?(fx_minversion_header|fx_minversion_inline|gecko_minversion_header|gecko_minversion_inline|IncludeSubnav|languages)"```
throws 0 results

PR #8040, opened 2022-08-27
remove `{{languages}}` macro from de

yin1999:
Part of #4579.

This macro just do nothing now, we can just remove it, see: https://github.com/mdn/yari/blob/27678e9d95e6bfdce42496b13b7592a6eb5d9709/kumascript/macros/languages.ejs


SphinxKnight (approved):
👌

PR #8086, opened 2022-08-28
remove `{{languages}}` macro from pl (Part 2)

yin1999:
Part of #4579.


yin1999:
Close in favor of #8107.

PR #8096, opened 2022-08-29
Remove some unwanted macros from ko

yin1999:
Part of #4579.

Those are now `no-op` macros, we can just remove them without any replacement.


hochan222 (approved):
LGTM :)

PR #8103, opened 2022-08-29
fix: learn/forms navigation link and missing part

jasonren0403:
fix navigation links and add back 'See also' part at the end of the document


jasonren0403:
temporaily convert to temp, because `Advanced_form_styling` was also changed in my previous commit #8030

PR #8111, opened 2022-08-30
fix: Markdownlint cleanup for ZH-CN

nschonni:


files/zh-cn/learn/css/building_blocks/cascade_and_inheritance/index.md
```diff
@@ -165,9 +165,9 @@ CSS 的简写属性 [`all`](/zh-CN/docs/Web/CSS/all) 可以用于同时将这些
 
 一个选择器的优先级可以说是由三个不同的值（或分量）相加，可以认为是百（ID）十（类）个（元素）——三位数的三个位数：
 
-2. **ID**：选择器中包含 ID 选择器则百位得一分。
-3. **类**：选择器中包含类选择器、属性选择器或者伪类则十位得一分。
-4. **元素**：选择器中包含元素、伪元素选择器则个位得一分。
+- **Identifiers**：选择器中包含 ID 选择器则百位得一分。
```
yin1999, line 168:
```suggestion
- **ID**：选择器中包含 ID 选择器则百位得一分。
```

nschonni: I can change it back, but they have it fully spelled out in the English version, which is why I changed it

yin1999: In zh-CN, we are used to use `ID` for element Identifiers. It's ok to change it back 😊

See also: line 166

yin1999 (approved):

yin1999:
By the way, should we consider to add a markdown lint workflow for documents as we did in mdn/content?

nschonni:
Yeah, the PR selective workflow that @OnkarRuikar figured out in the other repo would probably work here, but some changes are needed since there isn't a direct yarn.lock/package.json here

OnkarRuikar:
> some changes are needed since there isn't a direct yarn.lock/package.json here

We don't have to have the package mangers configured in the repo to do this:
```yml
 # Code refered from https://github.com/mdn/content/blob/main/.github/workflows/pr-check_markdownlint.yml#L24

     - name: Setup Node.js environment
        uses: actions/setup-node@v3
        with:
          node-version: "16"

      - name: Install required packages
        run: |
          ## make sure to install the specific version
          npm install -g markdownlint-cli@0.32.2

      - name: Lint markdown files
        run: |
          echo "::add-matcher::.github/workflows/markdownlint-problem-matcher.json"
          markdownlint ${{ steps.changed-files.outputs.all_changed_files }}
```

yin1999:
> ```bash
> npm install -g markdownlint-cli@0.32.2
> ```

Use command to install the cli may be a bit difficult to maintain (we can't use [Dependabot](https://github.com/dependabot) to update dependency)

An idea: if it's ok to use `curl`/`wget` to download the yarn.lock/package.json from `mdn/content` (checkout the whole mdn/content repo may be a bit heavy)

OnkarRuikar:
We can use following command to get the version: :thinking: 
```
> curl https://raw.githubusercontent.com/mdn/content/main/package.json 2>&1 | grep -Po '(?<=markdownlint-cli": \")[\d\.]+'
0.32.2
```

***

> (we can't use [Dependabot](https://github.com/dependabot) to update dependency)

For linters we do not want to update the version automatically.
We do it manually because new linter version often comes with new rules enabled by default. And they suddenly start flagging a lot of files or a lot of content in a file which may not be relevant to the updates in the PR.
So before updating to new linter version we clean the entire repo first.
The markdownlint hasn't updated that frequently.

PR #8149, opened 2022-08-31
Change to "Voir aussi" pt 1

cw118:
This is just a "find and replace", there were many files here so I will change some more in another PR (I didn't modify the files that will be edited by the recent `fr` pull requests to avoid conflicts)


SphinxKnight (approved):
Voyons ça rapidement avant les PR pour les Spec / Compat sur SVG ^^ 
Merci @cw118

PR #8152, opened 2022-08-31
fix(typo): la instead of le

H3r3zy:
Fix a typo where a "la" is in place of a "le"


SphinxKnight (approved):
Merci pour cette correction @H3r3zy :)

PR #8175, opened 2022-09-01
update slug in learn area

yin1999:


yin1999:
self approved for only updating slug

Issue #7615, opened 2022-08-13
Remove deprecated `{{Tag}}` macro


yin1999:
The `{{Tag}}` macro [has been deprecated](https://github.com/mdn/yari/pull/6269), and can be simply replaced by `<code>$0</code>` (regexp: `\{\{\s*tag\s*\((.*?)\)\s*\}\}` may have a outer `<code>` tag)

This macro is only used in `MDN/Writing_guidelines/Howto/Tag` or `MDN/Writing_guidelines/Experimental_deprecated_obsolete`.

- [x] de #7653
- [x] es #7823
- [x] ko https://github.com/mdn/translated-content/pull/7651
- [x] pt-br #7672

PR #8198, opened 2022-09-01
remove the last `{{tag}}` macro

yin1999:
Part of #7615.


Graywolf9 (approved):
Thank you so much for notice and fix it!

PR #8212, opened 2022-09-01
remove all {{tag}} macro from translated-content

yin1999:
Continue #8198.

Sorry for I've ignored the second `tag` macro in this file. 😢


JuanVqz (approved):
🚀thank you

PR #8218, opened 2022-09-02
ci: sync yml linting from mdn/content

nschonni:
Will run when any YML file is changed. Ran `prettier -w **/*.yml` to cleanup the current files


nschonni:
Now that the `package.json` has landed separately, I've changed this to use the scripts and a pinned version of Prettier in this repo instead of copying the files from mdn/content

.gitignore
```diff
@@ -1,3 +1,6 @@
 _githistory.json
 .DS_Store
 mdn/content
+node_modules/
```
caugner, line 4:
```suggestion
```

caugner (approved):

PR #8181, opened 2022-09-01
ci: sync Markdownlint setup from mdn/content

nschonni:
Copied the setup from the upstream repo, but added a step to copy over the package.json and yarn.lock to use the same version of the tools

PR #8219, opened 2022-09-02
ci: sync JSON linting from mdn/content

nschonni:
No current issues, and the Yari commandlind now syncs formats with Prettier, so the wikihistory.json files shouldn't have any issues unless they are manually changed and don't match the format


nschonni:
This might change if the package.json file lands in #8181

nschonni:
Added scripts and dependencies now that they have landed in this repo

caugner (approved):

PR #8257, opened 2022-09-03
Run `sync-translated-content` on `mdn/translated-content` only

cw118:
This workflow should only run on `mdn/translated-content`, not on forks.


caugner (approved):

cw118:
@caugner Thank you for the review :) Would it be OK to merge or is this PR waiting on something?

PR #8263, opened 2022-09-03
fix: Markdownlint cleanup for JA

nschonni:
Not sure if these should be skipped, so I separated them out from the other cleanups


mfuji09 (approved):
LGTM!

PR #8299, opened 2022-09-04

nschonni:
Left the orphans/conflicts separate since they might just be deleted


Graywolf9 (approved):
Great! Thank you so much for your time !

After the markdown conversion I'm going to check tasks like orphaned/conflicting to remove or order later, thanks!

PR #8333, opened 2022-09-06
Adding a space to successfully apply a syntax

Yu1us:
The first line should be added a space to successfully apply the "bold text" syntax.


yin1999 (approved):
Thanks. Welcome aboard!

PR #8334, opened 2022-09-06
Correct the errors of translation expression about zh-CN

yuemeng200:
原文的意思是`可枚举且自有`，这里翻译成`可枚举的自有属性`较为合适，`可枚举和自有`在中文语境里具有歧义。
The original meaning is 'enumerable and own', this has translated into 'enumerable own properties' is more appropriate, 'enumerable, own' have the ambiguity in the Chinese.


yin1999 (approved):
Thanks a lot. Welcome aboard!

PR #8360, opened 2022-09-07
Fix MDN broken link

karnaryen:
Replace broken url by working url from English version of this article (see https://developer.mozilla.org/en-US/docs/Web/CSS/content#element_replacement)

### DIFF
--- a/files/ru/web/css/content/index.md
+++ b/files/ru/web/css/content/index.md
@@ -243,7 +243,7 @@ li {
 
 ```css
 #replaced {
-  content: url("https://mdn.mozillademos.org/files/12668/MDN.svg");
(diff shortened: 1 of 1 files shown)


sashasushko:
@nat-k-dev let's wait for #8359

cw118:
@sashasushko hello, images used in live samples (`{{EmbedLiveSample}}`s) are affected by mdn/yari#5652, so for now the solution is to add the image files in the folders of locales as well (this may create many duplicates :/ but hopefully a fix will soon be found/merged in yari).

Let me know if help and/or clarifications are needed :)

sashasushko:
@cw118 thanks for notice. I didn't know.

PR #8378, opened 2022-09-08
ci: Sync mdn-content through PRs

nschonni:
Creates a PR per language instead of pushing a larger sync commit to main.
Pretty sure the current approach is intended, so feel free to close.
This would also allow marking `main` as protected against non-PR pushes

### DIFF
--- a/.github/workflows/sync-translated-content.yml
+++ b/.github/workflows/sync-translated-content.yml
@@ -60,7 +60,7 @@ jobs:
         uses: peter-evans/create-pull-request@v4
         with:
           commit-message: "${{ matrix.lang }}: sync translated content"
-          branch: ${{ matrix.lang }}-content-sync
-          title: "${{ matrix.lang }}: sync translated content"
+          branch: content-sync-${{ matrix.lang }}
+          title: "[${{ matrix.lang }}] sync translated content"
           author: mdn-bot <108879845+mdn-bot@users.noreply.github.com>
           body: Yari generated sync
--- a/.github/workflows/sync-translated-content.yml
+++ b/.github/workflows/sync-translated-content.yml
@@ -62,5 +62,5 @@ jobs:
           commit-message: "${{ matrix.lang }}: sync translated content"
           branch: content-sync-${{ matrix.lang }}
           title: "[${{ matrix.lang }}] sync translated content"
-          author: mdn-bot <108879845+mdn-bot@users.noreply.github.com>
+          author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
           body: Yari generated sync
--- a/.github/workflows/sync-translated-content.yml
+++ b/.github/workflows/sync-translated-content.yml
@@ -16,6 +16,20 @@ jobs:
     if: github.repository == 'mdn/translated-content'
     runs-on: ubuntu-latest
 
+    strategy:
+      fail-fast: false
+
+      matrix:
+        lang:
+          - es
+          - fr
+          - ja
+          - ko
+          - pt-br
+          - ru
+          - zh-cn
+          - zh-tw
+
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


yin1999:
The problem is, some locales are not active enough. Some old PRs which are not resolved may lead to documents conflict. (IMO, may be not right.)

nschonni:
This would just keep on updating/recreating each language PR until it landed, so there wouldn't be an issue with conflicts

nschonni:
But your right that then the less active locales might leave some stale redirects till it was merged I guess

yin1999:
An example:

In an old PR (not resolved):

> the bot move `slug/to/old` to `slug/to/new`

And in mdn/content we are going to move(or delete and redirect) the `slug/to/new` to `slug/to/new/new`

I'm not sure whether the bot will going to  create a new PR to move(or delete and redirect) the `slug/to/old` to `slug/to/new/new`. (If so, we can only try to (or notice to) close the old PR)

nschonni:
The job will continue to overwrite the commit in the branch and the same PR is used till it is closed https://github.com/peter-evans/create-pull-request#action-behaviour

.github/workflows/sync-translated-content.yml
```diff
@@ -33,25 +47,20 @@ jobs:
...
-          git commit -a -m "[CRON] sync translated content" && \
-          git pull --rebase upstream main && \
-          git push upstream main
+        run: yarn content sync-translated-content ${{ matrix.lang }}
+
+      - name: Create PR with sync for ${{ matrix.lang }}
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: "${{ matrix.lang }}: sync translated content"
+          branch: ${{ matrix.lang }}-content-sync
```
caugner, line 63: Use language as postfix, to keep branches together.

```suggestion
          branch: content-sync-${{ matrix.lang }}
```

.github/workflows/sync-translated-content.yml
```diff
@@ -33,25 +47,20 @@ jobs:
...
+          title: "${{ matrix.lang }}: sync translated content"
```
caugner, line 64: I have a slight preference for the `[locale]` pattern used by Korean contributors, see: https://github.com/mdn/translated-content/pulls?q=is%3Apr+%22%5Bko%5D%22+

```suggestion
          title: "[${{ matrix.lang }}] sync translated content"
```

caugner (approved):
One a half nits, otherwise LGTM.

.github/workflows/sync-translated-content.yml
```diff
@@ -33,25 +47,20 @@ jobs:
...
-          git push upstream main
+        run: yarn content sync-translated-content ${{ matrix.lang }}
+
+      - name: Create PR with sync for ${{ matrix.lang }}
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: "${{ matrix.lang }}: sync translated content"
+          branch: content-sync-${{ matrix.lang }}
+          title: "[${{ matrix.lang }}] sync translated content"
+          author: mdn-bot <108879845+mdn-bot@users.noreply.github.com>
```
yin1999, line 65: Should PR's `author` be the same as the `committer`? (Or when we use `Squash and merge`, the commit to main branch will be [`Co-authored`](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors))

And we have used `github-actions[bot]` in [previous content-sync commit](https://github.com/mdn/translated-content/commits?author=github-actions%5Bbot%5D). Maybe use `github-actions[bot] github-actions[bot]@users.noreply.github.com` is better:

```suggestion
          commit-message: "${{ matrix.lang }}: sync translated content"
          committer: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
          branch: content-sync-${{ matrix.lang }}
          title: "[${{ matrix.lang }}] sync translated content"
          author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
```

@caugner What's your opinion?

caugner: Yes, for consistency, it probably makes sense to use `github-actions[bot] <github-actions[bot]@users.noreply.github.com>`, but `author` should suffice (like [here](https://github.com/mdn/content/blob/e47cebe4a6db41979dcb37ac7115b38513f8b873/.github/workflows/interfacedata-updater.yml#L35-L45)).

```suggestion
          commit-message: "${{ matrix.lang }}: sync translated content"
          branch: content-sync-${{ matrix.lang }}
          title: "[${{ matrix.lang }}] sync translated content"
          author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
```

yin1999 (approved):
LGTM. Thanks @caugner, @nschonni!

yin1999:
Let's try this

yin1999:
It seems that, the PR created by workflows will not trigger other actions. This has been reported here: peter-evans/create-pull-request#48.

So those PR will not be labeled and tested by actions.

nschonni:
Good catch @yin1999 . It seems like labels could probably be set by the PR action manually, but the getting the part of getting it to start running the CI checks would require using a token.
@caugner any thoughts on using a new token? I could try and piggy back on the automerge one `${{ secrets.AUTOMERGE_TOKEN }}` but that doesn't seem like a great idea 😆 

Overall, it seems to be working (2 PRs created today) and it was a direct push without CI before

PR #8445, opened 2022-09-12
zh-CN: update AbortController example code

janyin:
The `zh-CN` version is not align with English version, and there is a Reference Error in `reports.textContent = 'Download error: ' + e.message;`. `reports` is not defined.

I think it should align with English version.


yin1999 (approved):
Thanks!

PR #8472, opened 2022-09-13
Markdown conversion for `pt-BR` - Cleanup - Guide section

SphinxKnight:
No conversion report as there is no error left


josielrocha (approved):
Thank you for your contribution @SphinxKnight!

PR #8498, opened 2022-09-14
fix Chinese typo

ellieot:
Here it should be for “Unoptimized images”， which means "未优化", the original here is wrong


yin1999 (approved):
Thanks! Welcome aboard~

PR #8542, opened 2022-09-16
fix code example

Adophlidu:
A subclass must call super when its constructor is declared. Sometimes, an empty constructor for the subclass are misleading


jasonren0403:
According to [EN document](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super), the empty `constructor() {}` is removed from the code example.

yin1999 (changes requested):
Please follow @jasonren0403's suggestion. If you want to update the code example, please update the english document first.

yin1999 (approved):

PR #8548, opened 2022-09-16
Web/SVG/Element/radialGradient を更新

mfuji09:


SphinxKnight (changes requested):
~@mfuji09 seems like there's a `_wikihistory.json` file here. Please let me know if you think it should be part of this PR, otherwise, it should be removed :) ~

Edit: ohmy you're fast :)

SphinxKnight (approved):
Nothing to add regarding the update in Japanese and OK since the extra file was removed.

Thank you @mfuji09

mfuji09:
@SphinxKnight I have failed to rebase so a conflict was caused, but I have corrected (to remove the file of `fr`). Thank you!

PR #8594, opened 2022-09-18
fix src assignment error in loadImage function

haicg:


yin1999 (approved):
Thanks a lot. Welcome aboard

PR #8636, opened 2022-09-19
Conflicting and orphaned cleanup as of 20220919

SphinxKnight:
Updated the glossary page for Firefox OS along the way.

Note for self (and maybe others): see https://github.com/mdn/yari/issues/7196 and checkout https://github.com/mdn/yari/commit/79f3eb93bfa4a3c1a6c9138fe5c350af4538d3f5 in mdn/yari while it's opened.


cw118 (approved):
💯

PR #8667, opened 2022-09-20
fix default value of scrollIntoViewOptions.block

ryuran:
fix default value of scrollIntoViewOptions.block in french translation it should be `'start'`


cw118 (approved):
Merci pour cette correction @ryuran !

PR #8674, opened 2022-09-20
fix: metadata spacing games and learn

nschonni:


SphinxKnight (approved):
Thanks @nschonni, I've sampled the review across the set. Seems Ok :)

PR #8675, opened 2022-09-20
fix: metadata spacing mdn and mozilla

nschonni:


SphinxKnight (approved):
👌

PR #8676, opened 2022-09-20
fix: metadata spacing web/a*

nschonni:


SphinxKnight (approved):
Sampled through ~10%: OK

Thanks again @nschonni for this.

PR #8692, opened 2022-09-21
Deleted a sentence that was repeated twice.

LucileGB:


SphinxKnight (approved):
Merci beaucoup @LucileGB :)

PR #8728, opened 2022-09-22
fix: remove space

btea:


yin1999:
Hi @btea, we've now converted all those documents into markdown. Please sync with upstream and then go with your contribution.

btea:
Ok.

PR #8760, opened 2022-09-24
zh-cn(fix): update about Compare  macro

fwqaaq:
I wanna update these Compare macro, they no longer need these params. Let me know if it's not necessay


yin1999:
Yes, we can remove the params, but please check whether these are Compat macro call in the reflect en-US document.

yin1999 (approved):

Jack-Zhang-1314:
> Yes, we can remove the params, but please check whether these are Compat macro call in the reflect en-US document.

Alright, I will do.

PR #8834, opened 2022-09-26
fix: add missing images and change paths in md-files

BamblooV:


SphinxKnight:
adding images here shouldn't be necessary. Using the links that are used in the English documents should suffice. When the resource does not exist in the tree, it fallbacks on mdn/content (en-US).

In other words, just fixing the links without adding the images should be enough (of course this applies if the image does not contain localized content and is the same as in en-US)

sashasushko (approved):

sashasushko:
@BamblooV спасибо 🙏

PR #8844, opened 2022-09-26
Use double quotes to remain consistent with other content

Klemart3D:
Use double quotes for `::after` pseudo-element to remain consistent with other content like for `::before`


SphinxKnight (approved):
Merci @Klemart3D :)

PR #8875, opened 2022-09-28
fix bold mark

YoungSx:


yin1999 (approved):
Thanks a lot. Welcome aboard

PR #8879, opened 2022-09-28
fix: periods inside code blocks

nschonni:
Took a look around for `` 。` `` to find where the code spans had been closed after the periods


nschonni:
@huacnlee noticed some of these looking at your tools PRs. Not sure if this is a case you're looking at for the autofix

yin1999 (approved):
Good catch, thanks @nschonni

PR #8948, opened 2022-10-01
Remove old specification macros

SphinxKnight:
Final part of #5618 for `fr`


cw118 (approved):
merci encore sphinx 😃

PR #8952, opened 2022-10-01
ci(pr-test): use GitHub API to get changed files + diff

yin1999:
### Description

get diff file and changed files' name from GitHub API is much faster.

### Motivation

as `gh` is an built-in GitHub cli in Action runner, we can use this tool to get changed files, which is much faster than `git diff` command (must fetch the history commits).

#### using `gh` cli

Only used 0 seconds.

https://github.com/yin1999/translated-content/actions/runs/3580409184/jobs/6022509363#step:6:4

image

#### using `git diff` (used by `technote-space/get-diff-action`)

used 40 seconds.

https://github.com/mdn/translated-content/actions/runs/3163384214/jobs/5150888786#step:7:1

image

#### conclusion

Using `gh` cli could speed up the PR test workflow. But when we use this, we must get the diff file from `https://github.com/ower/repo/compare/{base}...{head}.diff` and this diff file just show the diff in changed files (which may differ from `git diff` which will collected the changes in the main branch).


.github/workflows/pr-test.yml
```diff
@@ -39,32 +39,12 @@ jobs:
...
-          #
-          # But after you set `core.quotePath` you get:
-          #
-          #    ▶ git status
-          #    Changes not staged for commit:
-          #    ...
-          #   modified:   "files/en-us/glossary/bézier_curve/index.html"
-          #
-          # Now, this gets used by the `git diff ...` inside get-diff-action.
-          git config --global core.quotePath false
```
yin1999, line 61: not necessary for `gh` cli:

image

.github/workflows/pr-test.yml
```diff
@@ -39,32 +39,12 @@ jobs:
...
-
-      - uses: technote-space/get-diff-action@v6.1.0
-        id: git_diff_content
-        with:
-          PATTERNS: files/**/*.+(html|md)
-          SET_ENV_NAME: GIT_DIFF_CONTENT
+          DIFF_DOCUMENTS=$(gh pr view ${{ github.event.number }} --json files --jq ".files.[].path" | egrep -i "files/.*/.*\.(html|md)$" | tr "\n" " ")
+          echo "GIT_DIFF_CONTENT=$DIFF_DOCUMENTS" >> $GITHUB_ENV
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
yin1999, line 47: `GITHUB_TOKEN` is needed for `gh` cli

.github/workflows/pr-test.yml
```diff
@@ -132,7 +112,7 @@ jobs:
           # directory.
           # The purpose of this is for the PR Review Companion to later
           # be able to use this raw diff file for the benefit of analyzing.
-          git diff --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} > $BUILD_OUT_ROOT/DIFF
+          wget "https://github.com/${{ github.repository }}/pull/${{ github.event.number }}.diff" -O $BUILD_OUT_ROOT/DIFF
```
yin1999, line 115: download diff file from github

yin1999:
If this is ok, I'd like to reflect the changes in mdn/content

.github/workflows/pr-test.yml
caugner: Could we use GitHub CLI for this?

```suggestion
          gh pr diff > $BUILD_OUT_ROOT/DIFF
```

.github/workflows/pr-test.yml
```diff
@@ -39,32 +39,12 @@ jobs:
...
-          #
-          # Now, this gets used by the `git diff ...` inside get-diff-action.
-          git config --global core.quotePath false
-
-      - uses: technote-space/get-diff-action@v6.1.0
-        id: git_diff_content
-        with:
-          PATTERNS: files/**/*.+(html|md)
-          SET_ENV_NAME: GIT_DIFF_CONTENT
+          DIFF_DOCUMENTS=$(gh pr view ${{ github.event.number }} --json files --jq ".files.[].path" | egrep -i "^files/.*/.*\.(html|md)$" | tr "\n" " ")
```
caugner, line 44: Do we really need to specify the PR number, given that GitHub CLI is run in the context of the PR?

```suggestion
          DIFF_DOCUMENTS=$(gh pr view --json files --jq ".files.[].path" | egrep -i "^files/.*/.*\.(html|md)$" | tr "\n" " ")
```

.github/workflows/pr-test.yml
```diff
@@ -161,19 +141,21 @@ jobs:
...
 
-      - uses: technote-space/get-diff-action@v6.1.0
-        with:
-          PATTERNS: files/**/*.+(png|jpeg|jpg|gif|svg|webp)
-          ABSOLUTE: true
-          SET_ENV_NAME: GIT_DIFF_FILES
+      - name: Get changed images
+        if: ${{ env.GIT_DIFF_CONTENT }}
+        run: |
+          DIFF_IMAGES=$(gh pr view ${{ github.event.number }} --json files --jq ".files.[].path" | egrep -i "^files/.*/.*\.(png|jpeg|jpg|gif|svg|webp)$" | tr "\n" " ")
```
caugner, line 147: Could we retrieve **all** changed files in `files/` (`DIFF_FILES`) **once** (above near line 42) rather than twice, and then derive `DIFF_DOCUMENTS` and `DIFF_IMAGES` from it?

caugner (changes requested):
LGTM and promising, thanks! 🎉 

Just some comments:

.github/workflows/pr-test.yml
yin1999: seems that we should specific the pr number:

https://github.com/mdn/translated-content/actions/runs/3289036253/jobs/5420089835#step:6:10

.github/workflows/pr-test.yml
```diff
@@ -122,17 +105,18 @@ jobs:
...
           # This must match what we set in the "Build changed content" step above
           BUILD_OUT_ROOT: /tmp/build
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           # Save the raw diff blob and store that inside the ./build/
           # directory.
           # The purpose of this is for the PR Review Companion to later
           # be able to use this raw diff file for the benefit of analyzing.
-          git diff --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} > $BUILD_OUT_ROOT/DIFF
+          gh pr diff ${{ github.event.number }} > $BUILD_OUT_ROOT/DIFF
```
yin1999, line 119: https://cli.github.com/manual/gh_pr_diff

yin1999: tested locally

image

.github/workflows/pr-test.yml
caugner: TIL, thanks for checking! 🙂

.github/workflows/pr-test.yml
```diff
@@ -161,19 +145,20 @@ jobs:
...
-        with:
-          PATTERNS: files/**/*.+(png|jpeg|jpg|gif|svg|webp)
-          ABSOLUTE: true
-          SET_ENV_NAME: GIT_DIFF_FILES
+      - name: Get changed images
+        if: ${{ env.GIT_DIFF_CONTENT }}
+        # note that we should get the absolute path of the changed images
+        run: |
+          DIFF_IMAGES=$(cat ${{ env.CHANGED_FILES }} | egrep -i "^files/.*/.*\.(png|jpeg|jpg|gif|svg|webp)$" | xargs readlink -e | tr "\n" " ")
+          echo "GIT_DIFF_IMAGES=$DIFF_IMAGES" >> $GITHUB_ENV
```
yin1999, line 153: My bad, I've ignored to export absolute path here, see testing pr (yin1999/translated-content#8):

- failed with image (unused): https://github.com/yin1999/translated-content/actions/runs/3293923349
- passed with image (used): https://github.com/yin1999/translated-content/actions/runs/3293947316/jobs/5430964574

yin1999:
marked as draft until I figure out the workflow run (<https://github.com/yin1999/translated-content/actions/runs/3293826528/jobs/5430706523>). It seems to be not right.

yin1999:
Hi @caugner, I found that there is a [cli restrictions](https://github.com/cli/cli/issues/5368) for `gh pr view`. And I've tried to use the [solution](https://github.com/cli/cli/issues/5368#issuecomment-1087515074), but it can't work properly (for the [rate limation of GitHub GraphQL API](https://docs.github.com/en/graphql/overview/resource-limitations)), the output filename may only counts to 100 for some PR like #9534.

So, I try to find another way to filter modified and existed filenames. The idea listed below:

1. use `gh pr diff` or `wget https://url/to/pr.diff` to get diff file and save it to file:

    ```bash
    gh pr diff ${{ github.event.number }} > $BUILD_OUT_ROOT/DIFF
    ```

2. install `diffstat` and then get modified and existed filenames by the following command.

    ```bash
    diffstat -l $BUILD_OUT_ROOT/DIFF | grep "^b/" | sed 's?^b/??' > out
    ```

    **It seems that the modified and existed file would be in `b/` directory. I'm not really sure about this.**

Is this a proper solution?

yin1999:
> So, I try to find another way to filter modified and existed filenames.

I've testes this on a "large" PR (#9513).

image

.github/workflows/pr-test.yml
```diff
@@ -39,32 +43,20 @@ jobs:
...
-      - uses: technote-space/get-diff-action@v6
-        id: git_diff_content
-        with:
-          PATTERNS: files/**/*.+(html|md)
-          SET_ENV_NAME: GIT_DIFF_CONTENT
+          # Use the GitHub API to get the list of changed files
+          # documenation: https://docs.github.com/rest/commits/commits#compare-two-commits
+          gh api repos/{owner}/{repo}/compare/${{ env.BASE_SHA }}...${{ env.HEAD_SHA }} \
+            --paginate \
+            --jq '.files | .[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) | .filename' > ${{ env.CHANGED_FILES }}
```
yin1999, line 52: For there maybe too many files has been changed in a single PR, the `gh pr view` command used before can only fetch up to 100 changed files. But for almost all the PR, the commit in it would be less than `30` commits. So we can use github rest API to get changed files. (See <https://docs.github.com/rest/commits/commits?apiVersion=2022-11-28#compare-two-commits>). This API will limit 30 commits per page. So for those must few PRs, we can use `--paginate` (see options in <https://cli.github.com/manual/gh_api>) to fetch all the commits.

Addition: The [API limitation](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#requests-from-github-actions) from GitHub Actions with `GitHub Token` would be 1000 requests per hour. This should be enough for most of cases.

.github/workflows/pr-test.yml
```diff
@@ -132,7 +124,7 @@ jobs:
           # directory.
           # The purpose of this is for the PR Review Companion to later
           # be able to use this raw diff file for the benefit of analyzing.
-          git diff --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} > $BUILD_OUT_ROOT/DIFF
+          wget https://github.com/${{ github.repository }}/compare/${{ env.BASE_SHA }}...${{ env.HEAD_SHA }}.diff -O $BUILD_OUT_ROOT/DIFF
```
yin1999, line 127: We can see the diff url in https://docs.github.com/rest/commits/commits#compare-two-commits

This may avoid an additional api call (making it harder to reach the limitation)

yin1999 (commented):
There is a test I've made before

https://github.com/yin1999/translated-content/actions/runs/3580409184/jobs/6022509363#step:8:6

.github/workflows/pr-test.yml
yin1999: We can check this PR as an example: https://github.com/mdn/translated-content/pull/9513

PR info: https://api.github.com/repos/mdn/translated-content/pulls/9513
Compare info (all the 183 files in in the single rest response, see field `files`): https://api.github.com/repos/mdn/translated-content/compare/a9df62bb4885e5d9cdccd3834d5a24c921ef7cfa...16f5fc0db4ab1f5cb3858ab7878c428c19e9a79e

yin1999: If this is ok. Should we try it in [Markdown lint (PR)](https://github.com/mdn/translated-content/actions/workflows/pr-check_markdownlint.yml) first? There are some failed workflows (cannot get changed files), and it is less important than the current CI.

caugner:
@yin1999 This has probably been discussed before, but why can we actually not use `git ls-files`? My expectation would be that the workflow run in the context of the PR has the history of all commits between `BASE_SHA` and `HEAD_SHA`. Is this not so?

yin1999:
> My expectation would be that the workflow run in the context of the PR has the history of all commits between `BASE_SHA` and `HEAD_SHA`. Is this not so?

Actually not (see: actions/checkout#520), the `actions/checkout@v3` action could only fetch specified `n` depth (with `fetch-depth` input). By default, the param would be `1`, so the commit history will not be included in the workflow context.

image

If we need to fetch commits that are in a single PR (branch), we need to use `git fetch` with enough depth ([technote-space/get-diff-action's implement](https://github.com/technote-space/get-diff-action/blob/200391037d31f93b5b39b4b6f329a966c7ac29cf/src/utils/command.ts#L88-L92)) which may take too much time.

And I've created [another PR](https://github.com/mdn/translated-content/pull/10390) for markdown-lint (PR), to deal with the workflow failure. If the solution seems ok, could we try on markdown-lint. Because this workflow may be more important for us.

caugner:
@yin1999 Understood, thanks for clarifying!

Could you update this PR to align with the changes in https://github.com/mdn/translated-content/pull/10390 (i.e. without the temporary file)? And then it should be good to go. 🙏

.github/workflows/pr-test.yml
```diff
@@ -161,19 +156,13 @@ jobs:
...
 
-      - uses: technote-space/get-diff-action@v6
-        with:
-          PATTERNS: files/**/*.+(png|jpeg|jpg|gif|svg|webp)
-          ABSOLUTE: true
-          SET_ENV_NAME: GIT_DIFF_FILES
-
       - name: Check changed files
-        if: ${{ env.GIT_DIFF_FILES }}
+        if: ${{ env.GIT_DIFF_IMAGES }}
```
caugner, line 160: Can we rename this back to `GIT_DIFF_FILES`? Soon the `filecheck` will also check non-image file attachments, such as audio, video and font files (see https://github.com/mdn/yari/pull/7605).

caugner (approved):

PR #8957, opened 2022-10-01
fix: replace remaining Chrome licesnes

nschonni:
### Description

Most of the ones in french were converted, this cleans up the remaining, and a few that we accidentally translated


nathipg (approved):
Approved by @mdn/yari-content-pt-br

PR #8961, opened 2022-10-01
fix: remove developer domain for MDN links

nschonni:


nschonni:
Going to combine this with the other Markdownlint changes, but in some smaller PRs

PR #9048, opened 2022-10-05
chore(zh-CN): adjust example code layout

YunYouJun:
### Description

> https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Data

zh-CN docs: adjust example code layout


yin1999 (approved):
👍

Issue #8918, opened 2022-09-30
Localized issue template (`page-report.yml`)


cw118:
@JuanVqz asked if `page-report.yml` can be translated in each locale now that the new English issue template is being used on `translated-content`, so I decided to create this issue for discussion :)

I was thinking of creating a `page-report-<locale>.yml` file for each locale (e.g. `page-report-fr.yml`) in a PR, all in English first. I will update the implementation on `yari`, then locale teams can:
  - translate them on that PR with suggestions/commits
  - *or* it could be merged first, and each team creates a PR after to translate the template for their locale

Poking @mdn/yari-content-ru @mdn/yari-content-ja @mdn/yari-content-pt-br @mdn/yari-content-zh (apologies for pinging so many times 😅) who have yet not approved

**Edit:** mdn/yari#7277 should take care of the "Report on GitHub" link

nschonni:
What about stubbing each of the locales using the English template to start, then the change it Yari can land, with the translated templates to follow

Jack-Zhang-1314:
> What about stubbing each of the locales using the English template to start, then the change it Yari can land, with the translated templates to follow

This is a good feature for localization. Tbh, I'm looking forward.

cw118:
@nschonni as I wrote in the issue, 

> I was thinking of creating a page-report-<locale>.yml file for each locale all in English first.

The modifications are almost ready and I will create the PR very soon :)

JuanVqz:
I like the second option:

or it could be merged first, and each team creates a PR after to translate the template for their locale

Graywolf9:
Done for `es` in #9505

Thank you so much for all!

mfuji09:
Done for `ja` in #10857.

hochan222:
Done for ko in https://github.com/mdn/translated-content/pull/11536.

PR #9073, opened 2022-10-07
Create separate `page-report` for locales

cw118:
Create a `page-report` file for each locale, all in English first (w/ updates to CODEOWNERS) so that the yari PR can be merged sooner. See #8918 and mdn/yari#7277. The idea is that locale teams can translate templates after this and the yari update are merged.

Politely poking @mdn/mdn-community-engagement for a review :)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -57,7 +57,10 @@
 # ----------------------------------------------------------------------------
 /docs/        @mdn/core-yari-content
 /docs/es/     @mdn/yari-content-es
+/docs/fr/     @mdn/yari-content-fr
+/docs/ja/     @mdn/yari-content-ja
 /docs/ko/     @mdn/yari-content-ko
+/docs/pt-br/  @mdn/yari-content-pt-br
 /docs/ru/     @mdn/yari-content-ru
 /docs/zh-cn/  @mdn/yari-content-zh
 /docs/zh-tw/  @mdn/yari-content-zh
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -56,10 +56,23 @@
 # TRANSLATION GUIDE OWNER(S)
 # ----------------------------------------------------------------------------
 /docs/        @mdn/core-yari-content
+/docs/es/     @mdn/yari-content-es
 /docs/ko/     @mdn/yari-content-ko
 /docs/ru/     @mdn/yari-content-ru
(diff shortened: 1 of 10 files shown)


.github/CODEOWNERS
```diff
@@ -56,11 +56,24 @@
 # TRANSLATION GUIDE OWNER(S)
 # ----------------------------------------------------------------------------
 /docs/        @mdn/core-yari-content
+/docs/es/     @mdn/yari-content-es
 /docs/ko/     @mdn/yari-content-ko
 /docs/ru/     @mdn/yari-content-ru
```
nschonni, line 61:
```suggestion
/docs/pt-br/  @mdn/yari-content-pt-br
/docs/ru/     @mdn/yari-content-ru
```

nschonni (commented):
~~Not really related, just noticed the other change you made~~ missed the difference between the `/docs/` and the existing `/files/`

cw118:
> ~Not really related, just noticed the other change you made~ missed the difference between the `/docs/` and the existing `/files/`

I didn't see the edit of the comment and added `docs` owners for other locales in case translation guides are created for those, I guess we will see what `community-engagement` says. Thank you for the review and credit goes to @yin1999 for first noticing that CODEOWNERS needed an update :)

nschonni:
No problem, we just were updating at the same time. I don't think that the extra entries will cause any issue, I had only realized afterwards that the files didn't exist yet

JuanVqz:
Why is this failing?

cw118:
> Why is this failing?

It's just the system file change workflow, which always fails no matter the changes brought to system files (e.g. `.github` files)

Politely poking @mdn/mdn-community-engagement again for a review :)

JuanVqz:
Thank you for explaining, this is pretty need, I can’t wait to see it merged 🚀 thanks!

caugner (approved):
LGTM, although everyone will need to be careful not to change the field names (or this could be ensured via a workflow).

nschonni (approved):
I think the field thing should be fine, since this should still require the `@mdn/core-dev` approval for changes to these files

caugner:
@JuanVqz Good catch! Looks like GitHub doesn't like that all page report templates have the same first line:

```
name: "[via MDN pages only]"
```

The best would probably be to add the locale to it, e.g.:

```
name: "[via MDN pages only // es]"
```

caugner:
I'll draft a PR right away, as this breaks the links.

PR #9083, opened 2022-10-07
fix: Markdownlint cleanup for KO

nschonni:


nschonni:
@pje1740 would you like me to split this into smaller parts to make it easier to review?

pje1740:
Thank you for the suggestion! No need for splitting, I will review this as soon as possible!

pje1740 (approved):
Thank you so much for contributing to ko locale! I was going to check all the files, but if the number of files are kept below 30 (as Hochan mentioned), I think it would be better for future reviews 😊

PR #9092, opened 2022-10-08
fix: Remove Google Translate Links ES

nschonni:


Graywolf9 (approved):
great! thank you so much!

Issue #7178, opened 2022-07-28
`{{Source}}` macro removal in translated-content


yin1999:
The `{{Source}}` macro has been deprecated and [removed from `en-US`](https://github.com/mdn/content/pull/18872), we should remove it from l10n documents.

I've used rg to find all the usage: `rg "\{\{\s*source" -i -l`:

- ru
  - files/ru/web/api/selection/index.html
  - files/ru/web/api/event/comparison_of_event_targets/index.html

For `l10n-zh`, I've created #7177 to fix this.

mfuji09:
Done for `ja` by #7290

yin1999:
There is no `{{source}}` macro in translated-content. Let's close this

PR #9108, opened 2022-10-08
chore: remove the last source macro in ja

yin1999:
### Description

chore: remove the last source macro in ja

Part of #7178


mfuji09 (approved):
@yin1999 Thank you!

PR #9118, opened 2022-10-08
fix: Markdownlint cleanup for pt-br

nschonni:
### Description

Cleanup some bad escaping from the convesion, and some old punctuation that was stuck inside code blocks.
A few sections didn't exist in the en-US and hadn't been translated, so I just swapped with the current text


josielrocha (approved):
LGTM ;)

PR #9180, opened 2022-10-11
fix: remove dash for code fences

nschonni:


SphinxKnight (approved):
Thanks again @nschonni :)

PR #9182, opened 2022-10-11
fix: OL fences

nschonni:


SphinxKnight (commented):
@nschonni I'm not sure of the rationale/added value for this one since it's a list of steps for the reader to follow. To me, it's clearer to have the list item so that the process can be followed in order (from the simplest to the most complex snippet).

That being written, I won't argue too much for one single page but if you have any pointer, I'd be glad to read them :)

nschonni:
These were cleaned up upstream already. I can see there might be some benefit in some cases, but they mostly represented errors from IDEs auto-inserting them on formatting

PR #9211, opened 2022-10-12
Remove dead link and update/improve What are WebExtensions

SphinxKnight:
Remove https://mail.mozilla.org/listinfo/dev-addons from Mozilla/Add-ons/WebExtensions/What_are_WebExtensions
See https://github.com/mdn/translated-content/issues/7319 for rationale

I updated the page along the way and went with a more explicit structure than English.


cw118 (approved):
💯

PR #9215, opened 2022-10-12
remove {{languages}} macro from pt-br

yin1999:
### Description

remove {{languages}} macro from pt-br

### Motivation

This macro is used to generate links to other locale for the same-slug document. But now, the link are generated by mdn/yari. We can just remove the [no-op macro](https://github.com/mdn/yari/blob/5849ef0a1d3f544292725321d20be1f84242aaca/kumascript/macros/languages.ejs) now.

image

Part of #4579.


josielrocha (approved):
LGTM ;)

PR #9217, opened 2022-10-12
Remove dead link and update XSLT JS transformation

SphinxKnight:
Part of #7319 : removed https://dxr.mozilla.org/mozilla-central/source/content/xslt/public/nsIXSLTProcessor.idl and updated the page (going further would have required work on the English document first...)


cw118 (approved):
merci 🤗

PR #9235, opened 2022-10-13
set the correct image format

rosald:
'image/jpg' will ignore, and generate image/png (default) the currect is 'image/jpeg'


yin1999 (approved):
Thanks, welcome aboard

PR #9248, opened 2022-10-14
The results of Universal selectors are not displayed properly

isolcat:
### Description
The results of Universal selectors cannot be displayed properly, and comparing with the English document, I found that the Chinese document seems to be translated directly from the original text and written directly,I have modified it according to the English document

### Motivation
I hope the MDN Chinese documentation can demonstrate the effect of Universal selectors properly

### Additional details
THE PROBLEMS happend at the doc:https://developer.mozilla.org/zh-CN/docs/Web/CSS/Universal_selectors

THE CORRECT reference at the doc:https://developer.mozilla.org/en-US/docs/Web/CSS/Universal_selectors

### DIFF
--- a/files/zh-cn/web/css/universal_selectors/index.md
+++ b/files/zh-cn/web/css/universal_selectors/index.md
@@ -36,7 +36,7 @@ slug: Web/CSS/Universal_selectors
 
 则会产生这样的效果：
 
-{{EmbedLiveSample('Examples')}}
+{{EmbedLiveSample('示例')}}
 
 > **备注：** 笔者不推荐使用通配选择器，因为它是[性能最低的一个 CSS 选择器](http://www.stevesouders.com/blog/2009/06/18/simplifying-css-selectors/).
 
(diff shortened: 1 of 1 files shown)


files/zh-cn/web/css/universal_selectors/index.md
```diff
@@ -36,9 +36,7 @@ slug: Web/CSS/Universal_selectors
 
 则会产生这样的效果：
 
-A green span in a red paragraph.
-
-A red span in a green paragraph (with a border.)
+{{EmbedLiveSample('Examples')}}
```
yin1999, line 39: 其与标题一一对应（英文标题 example 被翻译为了“示例”），不能仅仅只是拷贝英文的内容：

```suggestion
{{EmbedLiveSample('示例')}}
```

isolcat: 感谢

yin1999 (approved):

PR #9265, opened 2022-10-14
[translate to japanese] CSP:font-src

neko-kawaii-nyan:
Translate CSP:font-src page to Japanese.
related issue : https://github.com/mozilla-japan/translation/issues/648


mfuji09 (approved):
良いと思います。

PR #9269, opened 2022-10-15
chore: synchronize upstream `.gitignore`

YangFong:
### Description

Synchronize upstream `.gitignore`.

### Motivation

I need to ignore `.idea` very much.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,4 @@
 /node_modules
-/build
-.env
-yarn-debug.log*
-yarn-error.log*
 _githistory.json
 .DS_Store
 .idea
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,3 +1,9 @@
+/node_modules
+/build
+.env
+yarn-debug.log*
+yarn-error.log*
```
caugner, line 5: These entries are irrelevant for translated-content.

```suggestion
```

caugner (changes requested):

.gitignore
YangFong: Okay, let me make an adjustment.

caugner (approved):

PR #9333, opened 2022-10-18
fix: remove CODE from macros FR

nschonni:


SphinxKnight (approved):
wat ^^ 

thanks @nschonni

PR #9343, opened 2022-10-18
Manually editing redirects to remove link with macros

SphinxKnight:
This removes two redirects which included a macro name.
I assume any consequences for those redirects being lost (even though cool URLs shouldn't change, I'm not sur those one were cool to begin with)


SphinxKnight:
well check_redirects looks happy, I'm going to merge this one as is.

PR #9362, opened 2022-10-19
Remove `{{event}}` from `fr`

SphinxKnight:
See https://github.com/orgs/mdn/discussions/263

Most of the changes are "replace with the links from the English page"
Sometimes, it involved deleting content or removing the macro and leaving the keyword on the page for the old `on<event>` pages which are not yet updated in `fr` to match `en-US` (where the page describes the event instead of the event handler)


cw118 (approved):
🔥🔥

PR #9390, opened 2022-10-19
fix: Update $edit links

nschonni:
### Description

These were broken from the transition and were pointing at the same page. In many cases the headings have been translated, so I'll need some suggestions to fix them from whoever reviews


hochan222 (approved):
LGTM. Thank you for your contribution :)

PR #9410, opened 2022-10-20
Changing 'var' variable to 'const', in pt-Br translate

Pedro0505:
### Description

Replacing variable declarations from 'var' to 'const'

### Motivation

The pt-BR translation is using obsolete variable declarations ('var') in some parts of the documentation, so it's always a good idea to update this documentation


josielrocha (approved):

josielrocha:
Obrigado pela ajuda @Pedro0505!

PR #9448, opened 2022-10-22
fix: CODE spans stuck to words

nschonni:


files/fr/web/html/date_and_time_formats/index.md
```diff
@@ -296,7 +296,7 @@ Une chaîne de caractères représentant une heure peut décrire différentes pr
 
 Voici quelques règles supplémentaires :
 
-- L'heure est toujours exprimée selon une horloge sur 24 heures où `00` correspond à minuit et où `23` correspond à onze heures du soir. Aucune valeur en dehors de l'intervalle `00`-`23` n'est autorisée.
+- L'heure est toujours exprimée selon une horloge sur 24 heures où `00` correspond à minuit et où `23` correspond à onze heures du soir. Aucune valeur en dehors de l'intervalle `00` - `23` n'est autorisée.
```
cw118, line 299: I'm not sure about this one as it's a range of numbers, and the [en-US file](https://github.com/mdn/content/blob/main/files/en-us/web/html/date_and_time_formats/index.md?plain=1#L290) is the same

nschonni: This one I'm doing an upstream PR for

cw118: If it's OK, I would prefer to wait for the upstream PR. It seems to me that a range of numbers should be written `num1`-`num2` without spaces between the numbers and the dash, so perhaps the way to fix this would be num1-num2, but I would like to see what the en-US reviewers say/do :)

SphinxKnight (approved):
Thanks once again @nschonni 
Edit: just saw @cw118 review, I'll leave this one pending, feel free to merge when you think it's OK @cw118 :)

cw118:
Let's merge this (for now at least), thank you @nschonni !

PR #9468, opened 2022-10-23
Move `WebAssembly` to new place (which is obstructing pull requests of other languages)

mfuji09:
### Description

Move documents of `WebAssembly` object to new place.

### Motivation

They have been already moved, but converting to md had run in parallel, so they were accidentally re-created in the old place. It causes a redirect error, and is obstructing pull request in the all other languages which has correcting redirects.


mfuji09:
I will merge by myself to repair redirects.

PR #9478, opened 2022-10-24
[zh-tw] Svelte variables props 部分新翻譯

cdcd72:
新增翻譯下列區段：

* 前言
* 文末連結

其它未翻譯部分，待後續有時間會再處理。


yin1999 (changes requested):
Hi @cdcd72, could you remove the image file in this PR. [MDN/yari] which is used to build raw docs can fallback to refer the image file in english documents.

PR #9389, opened 2022-10-19
fix: Remove escaped astrisks

nschonni:

PR #9512, opened 2022-10-25
fix: character escaping from conversion KO

nschonni:


nschonni:
@hochan222  or one of the other `ko` maintainers, would you like me to split this up a bit, since i think it is a little larger than your recommended PR size?

nschonni:
@pje1740 I gave this one a rebase because there was a bunch of overlap with #9389

hochan222 (approved):

hochan222:
@nschonni Sorry for the late reply. You've already done the job!

At least for the PR you gave me, if it's a markup syntax correction error, it's okay to commit more than 100 files as long as Github's loading time isn't too late.

Have a nice day!

PR #9520, opened 2022-10-26
fix: stuck DD characters

nschonni:
### Description

These DDs with a non-breaking space don't actually exist upstream, but added the extra space for consistency, since characters stuck to the colon in DDs usually break the rendering


hochan222:
> These DDs with a non-breaking space don't actually exist upstream, but added the extra space for consistency, since characters stuck to the colon in DDs usually break the rendering

@nschonni  Do you have an issue number that I can refer to?

hochan222 (approved):
Thanks for your contribution!

PR #9573, opened 2022-10-28

nschonni:


files/es/web/javascript/reference/global_objects/string/charcodeat/index.md
```diff
@@ -51,7 +51,7 @@ El siguiente ejemplo devuelve 83.
 "AaSdas".charCodeAt(2) // returns 83
 ```
 
-teniendo en cuenta que 2 es la posicion de la letra. Si \`S\` fuera minuscula, el Unicode es diferente
+teniendo en cuenta que 2 es la posicion de la letra. Si `S` fuera minuscula, el Unicode es diferente
```
Graywolf9, line 54:
```suggestion
teniendo en cuenta que 2 es la posicion de la letra. Si "S" fuera minuscula, el Unicode es diferente
```

Graywolf9 (approved):
Thank you so much @nschonni . I have a doubt, there is a kind of rule about having italics, bold or link in titles? we sometimes use italics in titles for words that could be better don't translate

nschonni:
> Thank you so much @nschonni . I have a doubt, there is a kind of rule about having italics, bold or link in titles? we sometimes use italics in titles for words that could be better don't translate

I don't really like it, but you can see the discussion here https://github.com/orgs/mdn/discussions/271. I kind of agree with the italics/strong, but I think code has a place

Graywolf9:
Great! Thansk you! I'm not sure but in most cases the emphasis is for code related words, thanks again

PR #9578, opened 2022-10-29
fix(zh-CN): uniform title name

NewVo1d:
### Description
The titles of each page of Array.prototype are not uniform, most of them are "示例", and some are "例子".

#### BeforeChange:
例子
#### AfterChange:
示例


yin1999 (approved):
Thanks

PR #9641, opened 2022-11-02

nschonni:
### Description

Mostly minor spacing from the deprecated macro cleanup


marcelozarate (approved):
Excellent! Thanks for your enormous work @nschonni

PR #9699, opened 2022-11-04
fix pic src in normal_flow

ZxBing0066:
### Description

The image in the normal_flow page has been broken.


yin1999 (approved):
Hi @ZxBing0066, thanks for raising this. Because of an issue in [mdn/yari](https://github.com/mdn/yari/issues/5652), we should refer the image with absolute path to en-US

PR #9702, opened 2022-11-04
Remove an old document which is already marked as a redirect

mfuji09:
### Description

Remove an old document which is already marked as a redirect.

### Motivation

This issue is obstructing PRs in other languages which have changing redirects.


mfuji09:
I have self-merged because this issue is obstructing PRs in other languages.

PR #9792, opened 2022-11-08
zh-cn(update): sync tostring about English Version

fwqaaq:


yin1999 (changes requested):
links to `zh-CN` locale

yin1999 (approved):

PR #9810, opened 2022-11-09
Fix some translation errors

Script-Xia:
The logical operator only should not be translated into Chinese


yin1999 (approved):

yin1999:
Thanks @Script-Xia and @Jack-Zhang-1314 👍

PR #9845, opened 2022-11-11
Fixed a wrong word

chufeng-official:


yin1999 (approved):
Thanks @chufeng-official

PR #9945, opened 2022-11-16
stroke-dasharray: add "space at start" example

homersimpsons:
### Description

Add "space at start" example for `stroke-dasharray`.

Notice that this is for the `fr` version.

### Motivation

It is not clear that we can start our "pattern" using a space. Doing that clarify that this is possible and show a simple way to achieve our result.

### Additional details

N/A

I did not open an issue for this.

### DIFF
--- a/files/fr/web/svg/attribute/stroke-dasharray/index.md
+++ b/files/fr/web/svg/attribute/stroke-dasharray/index.md
@@ -33,7 +33,7 @@ html,body,svg { height:100% }
 ```
 
 ```html
-<svg viewBox="0 0 30 10" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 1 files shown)


cw118 (changes requested):
Thank you for the PR @homersimpsons. However, French content should reflect its English counterpart, and the [en-US page has slightly different examples](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray#example). 
(But I would be happy to merge this if these examples are modified/added to the [English page on the `mdn/content` repo](https://github.com/mdn/content) :))

homersimpsons:
Hello @cw118 thank you, just opened https://github.com/mdn/content/pull/22372

cw118 (approved):
Merci beaucoup @homersimpsons, félicitations pour cette première contribution (et celle sur mdn/content) 🚀

PR #9948, opened 2022-11-16
Fix markup for the type of orientation

mikecat:
### Description

Use `<code>String</code>` instead of `` `String` `` like other pages.

### Motivation

The type is displayed as `` `String` `` and it looked weird for me.

The proposing markup `<code>String</code>` is used in other pages. For example, it is used here:

https://github.com/mdn/translated-content/blob/62b2b82989340ed4ca079f197c6ce7a20c2816fe/files/ja/web/manifest/name/index.md?plain=1#L12


mfuji09 (approved):
OKです。

PR #9950, opened 2022-11-16
Improve translation about waitUntil()

mikecat:
### Description

Improve translation of the note about `waitUntil()` in the Service Worker API page.

### Motivation

Current translation looks like the `waitUntil()` method calls `install` or `activate` and hands promises.

This looks weird because, according to the English version, the `waitUntil()` method is actually called from `install` or `activate` and handed promises.


mfuji09 (approved):
OKです。

PR #10040, opened 2022-11-21
[zh-cn]Fix docs: translation mistake

jynxio:
### Description

padding is wrongly translated as `外边距`, it should be translated as `内边距`.


yin1999 (approved):
Thanks @jynxio 👍

PR #10043, opened 2022-11-21
Fix translate error about [Optimizing for CRP]

kisnows:
### Description

The old ZH_CN translate about `[Optimizing for CRP](https://developer.mozilla.org/en-US/docs/Web/Performance/Critical_rendering_path#optimizing_for_crp)` has a mistake in understanding.


yin1999 (approved):
Thanks @kisnows, welcome aboard :)

PR #10049, opened 2022-11-21
Fix #9635

### LINKED ISSUE #9635: Mauvais liens
Capture d’écran 2022-11-01 à 20 23 43

Les liens pour le deuxième article correspond à la page du quatrième article.

SphinxKnight:
This is an update of https://developer.mozilla.org/fr/docs/Learn/JavaScript/Objects and a new translation for https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object-oriented_programming (which is a child page of the first one).

This should be a fix for #9635 (links aligned vs en-US and missing subpage now translated)

### DIFF
--- a/files/fr/learn/javascript/objects/index.md
+++ b/files/fr/learn/javascript/objects/index.md
@@ -1,51 +1,51 @@
 ---
-title: Introduction aux objets JS
+title: Introduction aux objets en JavaScript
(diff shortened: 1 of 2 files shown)


cw118 (approved):
Merci pour tout ce travail 🙏

PR #10193, opened 2022-11-26
chore: remove mention of deprecated macro

yin1999:
### Description

chore: remove mention of deprecated macro.

Note that: the current document should be the old translation of `https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Compatibility_tables` (for the sync content workflow has mismove the document to the current slug). And this part has been removed from the reflecting document.

See: mdn/content#20775

https://github.com/orgs/mdn/discussions/263


josielrocha (approved):
LGTM ;)

PR #10208, opened 2022-11-26
fix: broken definition list FR

nschonni:


cw118 (approved):
Looks good, thanks so much 😃

PR #10224, opened 2022-11-26
fix: Double spaces FR - Part 3

nschonni:


cw118 (approved):
Thanks again @nschonni !

PR #10245, opened 2022-11-27
fix: MD050 and links FR

nschonni:


SphinxKnight (commented):
Oh my, I guess I should go and update every page here but I'll gladly accept this patch in the meantime :)

SphinxKnight (approved):

PR #10268, opened 2022-11-28
Fix markdown code bold not work

Hacker-C:
### Description

Markdown code bold is not working because the incorrect spaces.

image

### Motivation

Make the docs more readable.

### Additional details

Here is the docs link: https://developer.mozilla.org/zh-CN/docs/Web/API/Window/DOMContentLoaded_event


yin1999 (approved):
Thanks @Hacker-C. Welcome aboard

PR #10280, opened 2022-11-28
fix: add newline after metadata block

nschonni:


SphinxKnight (approved):
I don't know (yet) why the workflow is failing but the change is OK for me.
Let's merge once the CI has been figured out

SphinxKnight:
OK so the CI is complaining about no shared commit ancestor. I'm going to give it a rebase and we'll see how it goes.
@nschonni I hope you won't have to deal with that again locally but if needed, I guess an update of your local state will be necessary :/

SphinxKnight:
let's goo

nschonni:
@SphinxKnight that job is having a little bit of an issue (sometimes) right now, but it's safe to ignore https://github.com/mdn/translated-content/pull/8181#issuecomment-1328532328

PR #10282, opened 2022-11-28
add zh-CN translation for a subtitle

Congb19:
### Description

String.prototype.fixed() 中有一个小标题没有翻译成中文。
A subtitle in String.prototype.fixed() has not been translated into zh-CN.


yin1999 (approved):
Thanks @Congb19. Welcome aboard :)

PR #10336, opened 2022-11-30
Fix doctype declaration in the French language

esnard:
### Description

I removed a space in the doctype declaration description in the French language.

### Motivation

The previous doctype declaration was invalid.


cw118 (approved):
Merci pour cette correction @esnard :)

PR #10344, opened 2022-11-30
Fix error in Web/API/setInterval

DreamOfIce:
Just fix a translate error.


yin1999 (approved):
Although this does not correspond to the original text, it is better for Chinese-speaking readers.

Thanks @DreamOfIce

PR #9513, opened 2022-10-25
Convert /api/i-r folder to Markdown (es)

Graywolf9:
### Description

Converting html to md from api/i-r folders

### Motivation

the chore of conversion

### Related issues and pull requests

Relates to https://github.com/mdn/translated-content/issues/8635

PR #10390, opened 2022-12-02
ci(markdownlint): use GitHub API to get changed files

yin1999:
### Description

fix(ci): use GitHub API to get changed files

### Motivation

The [tj-actions/changed-files@v34 action in Markdownlint (PR files)](https://github.com/mdn/translated-content/actions/workflows/pr-check_markdownlint.yml) may failed to get changed files. And as a test for mdn/translated-content#8952.

I've made a test [before](https://github.com/yin1999/translated-content/actions/runs/3293826528/jobs/5430706523) to get changed via GitHub API. It works, but I'm not sure whether it's always ok. I just find out a limitation in `gh` cli as [here](https://github.com/mdn/translated-content/pull/8952#issuecomment-1293201541) before, so I've tried another way to get changed files (by [commit compare REST API](https://docs.github.com/rest/commits/commits?apiVersion=2022-11-28#compare-two-commits)).

And for the API I used for now, it could list 30 commits per REST API request, so the files limitation maybe resolved, see below:

> PR info: https://api.github.com/repos/mdn/translated-content/pulls/9513
Compare info (all the 183 files in in the single rest response, see field files): https://api.github.com/repos/mdn/translated-content/compare/a9df62bb4885e5d9cdccd3834d5a24c921ef7cfa...16f5fc0db4ab1f5cb3858ab7878c428c19e9a79e

Is there any other things I've ignored? @nschonni @OnkarRuikar

more information can be found in those PR:

- #8952
- yin1999/translated-content#8

### DIFF
--- a/.github/workflows/pr-check_markdownlint.yml
+++ b/.github/workflows/pr-check_markdownlint.yml
@@ -17,17 +17,16 @@ jobs:
       - name: Get Changed files
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          CHANGED_FILES: /tmp/changed-files.txt
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
         run: |
           # Use the GitHub API to get the list of changed files
           # documenation: https://docs.github.com/rest/commits/commits#compare-two-commits
-          gh api repos/{owner}/{repo}/compare/${{ env.BASE_SHA }}...${{ env.HEAD_SHA }} \
+          DIFF_DOCUMENTS=$(gh api repos/{owner}/{repo}/compare/${{ env.BASE_SHA }}...${{ env.HEAD_SHA }} \
             --paginate \
-            --jq '.files | .[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) | .filename' > ${{ env.CHANGED_FILES }}
+            --jq '.files | .[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) | .filename')
           # filter out files that are not markdown
-          DIFF_DOCUMENTS=$(egrep -i "^files/.*\.md$" ${{ env.CHANGED_FILES }} | xargs)
+          DIFF_DOCUMENTS=$(echo "${DIFF_DOCUMENTS}" | egrep -i "^files/.*\.md$" | xargs)
           echo "DIFF_DOCUMENTS=${DIFF_DOCUMENTS}" >> $GITHUB_ENV
 
       - name: Setup Node.js environment
--- a/.github/workflows/pr-check_markdownlint.yml
+++ b/.github/workflows/pr-check_markdownlint.yml
@@ -14,12 +14,21 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Get changed files
-        id: changed-files
-        uses: tj-actions/changed-files@v34
-        with:
-          files: |
-            **/*.md
+      - name: Get Changed files
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CHANGED_FILES: /tmp/changed-files.txt
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          # Use the GitHub API to get the list of changed files
+          # documenation: https://docs.github.com/rest/commits/commits#compare-two-commits
+          gh api repos/{owner}/{repo}/compare/${{ env.BASE_SHA }}...${{ env.HEAD_SHA }} \
+            --paginate \
+            --jq '.files | .[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) | .filename' > ${{ env.CHANGED_FILES }}
+          # filter out files that are not markdown
+          DIFF_DOCUMENTS=$(cat ${{ env.CHANGED_FILES }} | egrep -i "^files/.*/.*\.md$" | tr "\n" " ")
+          echo "DIFF_DOCUMENTS=${DIFF_DOCUMENTS}" >> $GITHUB_ENV
 
       - name: Setup Node.js environment
         uses: actions/setup-node@v3
@@ -34,5 +43,4 @@ jobs:
       - name: Lint markdown files
         run: |
           echo "::add-matcher::.github/workflows/markdownlint-problem-matcher.json"
-          files_to_lint="${{ steps.changed-files.outputs.all_changed_files }}"
-          yarn markdownlint-cli2 ${files_to_lint}
+          yarn markdownlint-cli2 ${{ env.DIFF_DOCUMENTS }}
--- a/.github/workflows/pr-check_markdownlint.yml
+++ b/.github/workflows/pr-check_markdownlint.yml
@@ -27,7 +27,7 @@ jobs:
             --paginate \
             --jq '.files | .[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) | .filename' > ${{ env.CHANGED_FILES }}
           # filter out files that are not markdown
-          DIFF_DOCUMENTS=$(cat ${{ env.CHANGED_FILES }} | egrep -i "^files/.*/.*\.md$" | tr "\n" " ")
+          DIFF_DOCUMENTS=$(egrep -i "^files/.*\.md$" ${{ env.CHANGED_FILES }} | xargs)
           echo "DIFF_DOCUMENTS=${DIFF_DOCUMENTS}" >> $GITHUB_ENV
 
       - name: Setup Node.js environment
--- a/.github/workflows/pr-check_markdownlint.yml
+++ b/.github/workflows/pr-check_markdownlint.yml
@@ -43,4 +43,5 @@ jobs:
       - name: Lint markdown files
         run: |
           echo "::add-matcher::.github/workflows/markdownlint-problem-matcher.json"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-check_markdownlint.yml
```diff
@@ -34,5 +43,4 @@ jobs:
       - name: Lint markdown files
         run: |
           echo "::add-matcher::.github/workflows/markdownlint-problem-matcher.json"
-          files_to_lint="${{ steps.changed-files.outputs.all_changed_files }}"
-          yarn markdownlint-cli2 ${files_to_lint}
+          yarn markdownlint-cli2 ${{ env.DIFF_DOCUMENTS }}
```
OnkarRuikar, line 46: It'll throw syntax error when there are files with `(` in their path. Refer https://github.com/mdn/content/pull/20466
The indirection was used to avoid the syntax error.

For example,
```
# this doesn't work
touch file(css).md

# this does
touch "file(css).md"
```

yin1999:
```suggestion
          files_to_lint="${{ env.DIFF_DOCUMENTS }}"
          yarn markdownlint-cli2 ${files_to_lint}
```

yin1999: Thanks. Fixed in 9395ebad4a180e94913f6631326aa016978f4432

.github/workflows/pr-check_markdownlint.yml
```diff
@@ -14,12 +14,21 @@ jobs:
...
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          # Use the GitHub API to get the list of changed files
+          # documenation: https://docs.github.com/rest/commits/commits#compare-two-commits
+          gh api repos/{owner}/{repo}/compare/${{ env.BASE_SHA }}...${{ env.HEAD_SHA }} \
+            --paginate \
+            --jq '.files | .[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) | .filename' > ${{ env.CHANGED_FILES }}
+          # filter out files that are not markdown
+          DIFF_DOCUMENTS=$(cat ${{ env.CHANGED_FILES }} | egrep -i "^files/.*/.*\.md$" | tr "\n" " ")
```
OnkarRuikar, line 30: nitpick
```suggestion
          DIFF_DOCUMENTS=$(egrep -i "^files/.*\.md$" ${{ env.CHANGED_FILES }} | xargs)
```

yin1999: Thanks

OnkarRuikar:
This will be efficient cos unlike the third party action this uses GitHub APIs instead of git(requires to fetch a lot of history).
This also reduces dependency on third party action(which has to handle a lot of use cases, our requirement is simple).
***
@yin1999 can we use the [List pull requests files API](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files) instead of the diff API? It will make the logic simple and reduce a few lines of code.
```
gh api /repos/OWNER/REPO/pulls/PULL_NUMBER/files
```
***
For testing you'll have to do the same change in https://github.com/yin1999/translated-content/pull/8 . After that wait till there are some commits to the main branch then manually rerun the worflows(without rebasing).
Better if you create a new test PR for this in your fork.
***
Nitpick, we can work without using `/tmp/changed-files.txt`:
```
DIFF_DOCUMENTS=$(gh api ...)
DIFF_DOCUMENTS=$(echo ${DIFF_DOCUMENTS} | egrep -i "^files/.*\.md$" | xargs)
...
```

yin1999:
> This will be efficient cos unlike the third party action this uses GitHub APIs instead of git(requires to fetch a lot of history). This also reduces dependency on third party action(which has to handle a lot of use cases, our requirement is simple).
> 
> @yin1999 can we use the [List pull requests files API](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files) instead of the diff API? It will make the logic simple and reduce a few lines of code.
> 
> ```
> gh api /repos/OWNER/REPO/pulls/PULL_NUMBER/files
> ```

Hi @OnkarRuikar. I've tried this before. But [there](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files) is a limitation of 30 files per_page. So for some large PR. We may failed to fetch all the changed files (use `gh api` with `--paginate`). (You can try it locally, for this PR #9513, sometimes we can only fetch 100 changed files, I just could not figure out the reason.)

And if we use `compare` api, we can make sure to only list out changed file for the single workflow (that means, we may push a new commit, so the files listed by `pulls` api may be different with the corresponding workflow run (`head` and `base` sha may changed))

> Nitpick, we can work without using `/tmp/changed-files.txt`:
> 
> ```
> DIFF_DOCUMENTS=$(gh api ...)
> DIFF_DOCUMENTS=$(echo ${DIFF_DOCUMENTS} | egrep -i "^files/.*\.md$" | xargs)
> ...
> ```

Thanks for the suggestion, I'll apply this.

yin1999:
also /cc @SphinxKnight for any suggestions :)

nschonni:
Cool, if you've found a better solution, it might be worth upstreaming to that tj action, or create it as an action in mdn/workflows later for easier maintenance

yin1999:
> Cool, if you've found a better solution, it might be worth upstreaming to that tj action, or create it as an action in mdn/workflows later for easier maintenance

Thanks for suggestions, [tj action](https://github.com/tj-actions/changed-files) metioned `No extra API calls.`, so this maybe not a good idea to upstreaming this. And I'm not sure if this always works well, so this might be a good idea to have a test on this workflow first.

OnkarRuikar:
@yin1999 
> But [there](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files) is a limitation of 100 files per_page. So for some large PR. We may failed to fetch all the changed files (use gh api with `--paginate`).

The [Compare two commits](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#compare-two-commits) API also has the same limitations. The `--pageinate` option works perfectly with both the APIs.

> You can try it locally, for this PR https://github.com/mdn/translated-content/pull/9513, sometimes we can only fetch 100 changed files, I just could not figure out the reason.

It works in my environment:
```bash
gh api /repos/mdn/translated-content/pulls/9513/files --paginate \
  --jq '.[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) |.filename' | wc -l

93
```
I tried it multiple times and it gave me 93 count every time which is correct: 93(added) + 90(deleted) = 183. Note, we are excluding deleted files in the query.
So we can use the [List pull request files API](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files) without any issues.


> And if we use compare api, we can make sure to only list out changed file for the single workflow (that means, we may push a new commit, so the files listed by pulls api may be different with the corresponding workflow run (head and base sha may changed))

Head always points to the latest commit and the base remains the same every time. So with the current compare API code we always get files modified from first commit to the latest commit. So how is it different from the list PR files api?

***
@nschonni 
> Cool, if you've found a better solution, it might be worth upstreaming to that tj action, or create it as an action in mdn/workflows later for easier maintenance

There is already another third party action available that uses the same strategy (of using the Github compare commits API): https://github.com/jitterbit/get-changed-files/blob/d06c756e3609dd3dd5d302dde8d1339af3f790f2/src/main.ts#L62-L67
We could simply replace tj-action with that one or have our own in-house solution(mdn/workflow) using this PR's code.

Third party workflows provide a lot of features that we don't use. I am in favor of having our own small, simple, customized solution to avoid any dependency.

yin1999:
> So how is it different from the list PR files api?

image

But we can make sure that it's ok for the final workflow run.

OnkarRuikar:
We need to merge this solution quickly and do this in `mdn/content` as well, because the PR checks are failing frequently now:
- https://github.com/mdn/translated-content/actions/workflows/pr-check_markdownlint.yml?query=is%3Afailure
- https://github.com/mdn/content/actions/workflows/pr-check_markdownlint.yml?query=is%3Afailure

Ping @caugner

caugner (approved):

yin1999:
Thanks @caugner, @OnkarRuikar and @nschonni. Let's see would it works fine. So we can go ahead for mdn/content and [another workflow](https://github.com/mdn/translated-content/actions/workflows/pr-test.yml) :)

OnkarRuikar:
@yin1999 to test this quickly, rerun workflows on older PR without rebasing it.

yin1999:
> @yin1999 to test this quickly, rerun workflows on an old PR without rebasing it.

Yes, we have some workflow runs here: https://github.com/mdn/translated-content/actions/workflows/pr-check_markdownlint.yml (with a merge commit in this PR: #10508)

OnkarRuikar:
No, we want base branch few commits ahead of the feature branch. That means we need to let the branch sit idle till main has more commits after that without updating the branch by any means do `Re-run jobs`:
guidepages

yin1999:
> No, we want base branch few commits ahead of the feature branch. That means we need to let the branch sit idle till main has more commits. After that without updating the branch by any means do `Re-run jobs` in Actions tab: guidepages
> 
> I am sure this won't fail but because of this test case we got rid of the tj-action.

Sorry for the mis-understand. For the old PR would run with the old markdown-lint workflow. So I could not test it out. See: https://github.com/mdn/translated-content/actions/runs/3618623238/jobs/6125024148#step:3:1

OnkarRuikar:
There is a huge performance improvement.
Execution time went down from ~150sec to ~40sec!

PR #10395, opened 2022-12-02
[zh-CN] Fixed a typo in introducing workers

ghost:
Fix `delicated` to `dedicated` L221


yin1999 (approved):
Thanks @lodisy. Welcome aboard 🎉

PR #10402, opened 2022-12-02
fix(images): replace mdn.mozillademos.org references with local files

caugner:
### Description

Replaces all occurrences of **images** referencing https://mdn.mozillademos.org/files/ with the filename, and imports the file from a backup if the file does not yet exist in mdn/content or mdn/translated-content.

### Motivation

The https://mdn.mozillademos.org/ site is no longer working since MDN's backend migration from mdn/kuma to mdn/rumba.

### Related issues and pull requests

**Fixes https://github.com/mdn/translated-content/issues/10400.**

Fixes https://github.com/mdn/translated-content/issues/10324.
Fixes https://github.com/mdn/translated-content/issues/10323.
Fixes https://github.com/mdn/translated-content/issues/10322.

Issue #10400, opened 2022-12-02
[Meta] Replace https://mdn.mozillademos.org references


caugner:
## Background

Originally, images in pages on MDN were served from https://mdn.mozillademos.org/files/, and this has since been **deprecated** in favor of putting those images next to the page content.

## Problem

While all occurrences in mdn/content have already been replaced by local images (mostly in [these PRs](https://github.com/mdn/content/pulls?q=is%3Apr+is%3Aclosed+%22mdn.mozillademos.org%22+merged%3A%3C2021-06-30), e.g. [this one](https://github.com/mdn/content/pull/6084/files)), mdn/translated-content still has 2941 occurrences in 941 files referencing that domain.

As we plan to decommission the domain as part of an upcoming move from AWS to GCP, these usages will need to be updated.

### 1. Replace images that exist in content

- https://github.com/mdn/translated-content/pull/10474
- https://github.com/mdn/translated-content/pull/10475
- https://github.com/mdn/translated-content/pull/10476
- https://github.com/mdn/translated-content/pull/10477
- https://github.com/mdn/translated-content/pull/10478
- https://github.com/mdn/translated-content/pull/10479, #10508

### 2. Import missing images

- https://github.com/mdn/translated-content/pull/10480
- https://github.com/mdn/translated-content/pull/10481
- https://github.com/mdn/translated-content/pull/10482
- https://github.com/mdn/translated-content/pull/10483
- https://github.com/mdn/translated-content/pull/10484
- https://github.com/mdn/translated-content/pull/10485

### 3. Other manual tasks

- https://github.com/mdn/translated-content/pull/10404

## Scope

This affects the following locales (according to `rg mdn.mozillademos.org/files/`):

- [x] es (181 pages)
- [x] fr (0 pages)
- [x] ja (48 pages)
- [x] ko (167 pages)
- [x] pt-br (160 pages)
- [ ] ru (267 pages)
- [x] zh-cn (99 pages)
- [x] zh-tw (11 pages)

caugner:
FWIW https://github.com/mdn/translated-content/pull/10402 reduces the number of occurrences from 2941 to 2112 (-829).

yin1999:
@SphinxKnight just wrote a [script](https://gist.github.com/SphinxKnight/60d766b8d6cd970b42eaf967b8dac3ff) before to check duplicated assets before. I'm not sure would it be helpful.

[...]

PR #10479, opened 2022-12-05
[zh] Replace mdn.mozillademos.org images that exist in content

caugner:
### Description

Replaces occurrences of images that reference https://mdn.mozillademos.org/files/ with the (normalized) filename, but only those that exist in mdn/content (at the same slug location).

(Missing images are added in https://github.com/mdn/translated-content/pull/10485.)

### Motivation

The domain no longer serves images (it used to redirect).

### Additional details

**Note**: Unfortunately, the "Preview URLs" don't include images from content, so the only way to test that this works is to check out this PR and visit the URLs locally.

Extracted from https://github.com/mdn/translated-content/pull/10402.
Part of https://github.com/mdn/translated-content/issues/10400.

### DIFF
--- a/files/zh-cn/learn/css/building_blocks/advanced_styling_effects/index.md
+++ b/files/zh-cn/learn/css/building_blocks/advanced_styling_effects/index.md
@@ -242,7 +242,7 @@ div {
   padding: 10px;
   margin: 10px;
   display: inline-block;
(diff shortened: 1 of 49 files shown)


yin1999 (approved):
I've applied some changes. And LGTM now. We still need to resolve mdn/yari#5652. Or we could not see the expected live sample in production.

PR #10495, opened 2022-12-06
updated js code to english version

borwi4ok:
Code which using in code insertion is old.


saionaro (approved):
@borwi4ok thank you for the update!

PR #10521, opened 2022-12-07
remove CanvasSidebar from l10n-zh

yin1999:
### Description

remove CanvasSidebar from l10n-zh

mdn/yari#7760


Graywolf9:
I was looking to take yours as an example and notice that there is extra `}}` in `{{DefaultAPISidebar("Canvas API")}}}}` hope don't mind the observation

PR #10522, opened 2022-12-07
[zh] fix <meter> Markdown format error

zhoushengdao:
### Description

fix <meter> Markdown format error


yin1999 (approved):
Thanks @zhoushengdao

PR #10617, opened 2022-12-12
fix(ci): do not paginate for fetching file changes

yin1999:
### Description

do not paginate for fetching file changes. This is a reflect of mdn/content#22898. (We don't need to add `--paginate` for all the file changes are in the first API call)

poke @caugner for a review


caugner (approved):

caugner:
Thank you @yin1999 once again! 🎉

PR #10678, opened 2022-12-17
docs: correct some translations in autocomplete

jasonren0403:


files/zh-cn/web/html/attributes/autocomplete/index.md
```diff
@@ -201,9 +201,9 @@ TN99 8ZZ
 
 ## 另请详见
```
yin1999, line 202:
```suggestion
## 参见
```

yin1999 (approved):

PR #10693, opened 2022-12-17
fix: replace hidden PR license blocks JA

nschonni:
### Description

Was cleaned upstream since it sometimes caused Markdownlint parsing issues


mfuji09 (approved):
LGTM.

PR #10694, opened 2022-12-17
fix: replace hidden PR license blocks KO

nschonni:
### Description

Sync with other locale cleanup of these blocks


hochan222 (approved):
LGRM :)

PR #10807, opened 2022-12-28
Update padding page [es]

Mangostar1:
### Description

i change the comment line that describe the padding prosition

### Motivation

because old vercion its confused to read


Graywolf9 (dismissed):

Unfortunately we took the `en-US` version as a template, we only translate it, and the original version looks more like the `en-US`, also notice that this article is a little outdated, if you want you can update the whole article.

Also to made change to the document structure you can suggest it first to the English version: https://developer.mozilla.org/en-US/docs/web/css/padding, https://github.com/mdn/content/blob/main/files/en-us/web/css/padding/index.md?plain=1, once aproved, we can merge it here.

Graywolf9:
Hello! @Mangostar1 

Hope you are well, do you have any comments or updates?

PR #10837, opened 2022-12-31
fix: some markdown render issue and link flaws

jasonren0403:
### Description

There are some render issue and flaws in advanced_text_formatting zh-cn document, here are some fixes.


jasonren0403:
I also noticed some text that should be transformed into EmbedLiveSample calls, let me do it with next commit.

yin1999 (approved):

PR #10962, opened 2023-01-06
Error in end of sentence

Massoud114:
The sentence ends with "à", which is not normal.

### Description

complete the sentence to make sense of it, because the end of the sentence has been mistranslated

### Motivation

These changes will ensure that the reference at the end of the sentence is not misunderstood


SphinxKnight (approved):
Merci pour cette correction @Massoud114 :)

PR #10998, opened 2023-01-08
fix(ru): article __defineGetter__

scffs:
var -> let


sashasushko:
@scffs thank you. Let's use const as in [original article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__defineGetter__)

scffs:
Got it, thanks!
next time i'll check the orig. article before changing smth

saionaro (approved):
@scffs спасибо за правки!

PR #10999, opened 2023-01-08
fix(ru): article __defineSetter__

scffs:
var -> let


sashasushko:
@scffs thank you. Let's use `const` as in [original article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__defineSetter__)

sashasushko:
@scffs please, also check other your RP 🙏  We should sync translation version with the original one, with English.

scffs:
oh, sry I didn't think about it
got it now tnaks!

sashasushko:
Source of the original version you can find here – https://github.com/mdn/content/ – the file structure is the same

scffs:
I think that's done

saionaro (approved):
@scffs спасибо за правки!

PR #11000, opened 2023-01-08
fix(ja): fix typo in Element.previousElementSibling

megos:
### Description

`Element.previousElementSibling` 内の誤字を修正しました

https://developer.mozilla.org/en-US/docs/Web/API/Element/previousElementSibling

原文：

> This example outputs the following into the page when it loads:

### Motivation
N/A

### Additional details
N/A

N/A


mfuji09 (approved):
OKです。

PR #11072, opened 2023-01-12
fix(ru): article Global_Objects/Object

scffs:
let -> const as in orig article


saionaro (approved):

Saionaro:
@scffs thank you for contributing to MDN!

PR #11085, opened 2023-01-14
fix:(ru) translation of "JavaScript/Howto" article

iampopovich:
previous translation meaned naming for method
current translation means exactly calling of method


sashasushko (approved):

sashasushko:
@iampopovich спасибо 🙏

PR #11201, opened 2023-01-21
[ko] remove event page for legacy

hochan222:
### Description

en-us 버전과 많이 차이가 나서 삭제합니다.

---

https://github.com/orgs/mdn/discussions/263#discussioncomment-4743530


SphinxKnight:
@hochan222 it seems the deletion caused an error for redirects but I can't see any job about checking redirects here? Do you have any idea?

OnkarRuikar:
> @hochan222 it seems the deletion caused an error for redirects but I can't see any job about checking redirects here? Do you have any idea?

Because the redirect check workflow doesn't work on `.md` files at the moment:
https://github.com/mdn/translated-content/blob/76fe3d280db2c739652a929dcb8eae4f3d18bc58/.github/workflows/pr-check_redirects.yml#L1-L12

On `mdn/content` repo it does run on all files:
```yml
name: Check Redirects

on:
  pull_request:
    branches:
      - main
    paths:
      - files/**
      - .github/workflows/pr-check_redirects.yml

jobs:
```

Ping @nschonni

SphinxKnight:
Thanks @OnkarRuikar I did not even thought of that first :x 
Here is https://github.com/mdn/translated-content/pull/11218

hochan222:
@SphinxKnight First of all, I'm sorry for the problem, and thank you for resolving the problem.

Could it be that the problem is occurring in a different locale? Clearly, all git actions of PRs since then, including the current PR, have passed in ko-locale.

I thought that was strange too... So I was waiting for the mdn bot to reflect only one file in the PR and renew the redirect.

Previously, there was an experience in which the redirect action was applied to the md file, but when I thought about it, it was not the case that only the md file was present because it was changed from html to md extension.

SphinxKnight:
No need to be sorry. As I understand it now, the failure only occurred for PRs which touched something else than Markdown files (ex. "content-sync-<locale>" ones, touching history and redirects). So if you had PRs only editing pages, no error would be raised since the workflow is not triggered.

Anyway, let's try to reproduce and check what happens after #11218 has been merged :)

PR #11221, opened 2023-01-23
fixed incorrect reference links

kokic:
Change "...类似于 Array.prototype.indexOf()，但只是..." to "...类似于 Array.prototype.findIndex()，但只是..."

This is a very obvious problem caused by carelessness (in the translation process), and in the English document (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) it is "...similar to findIndex(), but...".


yin1999 (approved):

yin1999:
Thank you @kokic. Welcome aboard :tada:

PR #11248, opened 2023-01-25
fix: numComponents

liyifan2002:
### Description

For this 2d vertex buffer, the numComponents value should be 2.

### Motivation
When I was learning and testing my code with the guide on this page, I found it didn't work.
After comparing it with the source of [demo page](https://mdn.github.io/dom-examples/webgl-examples/tutorial/sample3/), I noticed that the value of numComponents is different. It works well when the variable equals 2.


yin1999 (approved):
Thank you @liyifan2002. Welcome aboard :tada: :)

PR #11263, opened 2023-01-26
Translate text from the English 'How the web works' article

ruurtjan:
### Description

For reference, this is the original article:
https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/How_the_Web_works

The English text is:
> You can use a DNS lookup tool to find the IP address of a website.


SphinxKnight (approved):
Thanks @ruurtjan, I've reworded a bit and followed the convention about web/Web from https://github.com/mozfr/besogne/wiki/Guide-stylistique-pour-la-traduction#divers

SphinxKnight:
Forgot to merge earlier :x

PR #11285, opened 2023-01-28
fix(ru): translation mistake

GusarevKirill:


captainspring:
@GusarevKirill, thank you for the correction!

PR #11291, opened 2023-01-29
fix: service-worker lifecycle image invalid

lemoxcc:
### Description
service worker lifecycle image invalid

### Motivation
When reading this docs of service worker，I found this picture of 'service worker lifecycle'. Therefore, I referred to the primitive english docs and would like to make a change to this.


yin1999 (approved):
Thanks

PR #11302, opened 2023-01-30
Remove `{{fx_minversion_note}}`

SphinxKnight:
Regarding the "Using XHR" page, I removed the large examples using some mix of PHP and mentions of IE9. As far as I can tell, they don't follow current "MDN" editorial practices for examples and should be replaced by Fetch equivalent.


cw118 (approved):
Merci beaucoup pour tout ce travail Sphinx 🙏

PR #11358, opened 2023-02-02
chore: add `In this module` translation to translation-guide

yin1999:
### Description

chore: add `In this module` translation to translation-guide

Also sort items by lexicographical order

### DIFF
--- a/docs/zh-tw/translation-guide.md
+++ b/docs/zh-tw/translation-guide.md
@@ -129,11 +129,11 @@ For example, consider the [JavaScript](/en-US/docs/Web/JavaScript) guide, which 
 
 | 原文 | 翻譯 |
 | --- | --- |
-| Try It | 試試看 |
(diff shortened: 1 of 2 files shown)


docs/zh-tw/translation-guide.md
```diff
@@ -129,11 +129,11 @@ For example, consider the [JavaScript](/en-US/docs/Web/JavaScript) guide, which
 
 | 原文 | 翻譯 |
 | --- | --- |
-| Try It | 試試看 |
```
yin1999, line 132: used in [`EmbedInteractiveExample`](https://github.com/mdn/yari/blob/d89ab8028f93c8ae13b1316a1fc2ee252d91eb4c/kumascript/macros/EmbedInteractiveExample.ejs#L54-L55). So remove this from here (for we would not need to translate this)

PR #11379, opened 2023-02-03
fix typo

taquanjianimei:
There is a spelling error in this article,which is RSTP. I changed it to RTSP


yin1999 (approved):
Thank you @taquanjianimei. Welcome aboard :tada:

PR #11413, opened 2023-02-05
pt-BR: Fix removal of {{ARIARole}} macro

queengooborg:
This PR fixes the changes made in #11341.  The regex search-and-replace had an issue that caused the links and text to break, thus unfortunately breaking the localizations.


queengooborg:
Haven't heard from the pt-BR team but I don't think there's any changes in this PR that require locale-specific review, so I'm going to merge this.

Issue #11430, opened 2023-02-06
Update sidebar under `webassembly` folder


yin1999:
The `WebAssembly` pages are now having their own sidebar: `{{WebAssemblySidebar}}`. Let's replace the old `{{JSRef}}` sidebar with the new one.

- [x] l10n-zh #11431
- [x] l10n-es #11474
- [x] l10n-fr #11472
- [x] l10n-ja #11475
- [ ] l10n-ko #11481
- [x] l10n-ru #11480

PR #11431, opened 2023-02-06
chore: replace sidebar under `WebAssembly` section

yin1999:
### Description

chore: replace sidebar under `WebAssembly` section

#11430


yin1999 (commented):
confirmed that all those pages are using webassemblysidebar in en-US

PR #11437, opened 2023-02-06
fix definition list

yin1999:
### Description

fix definition list


yin1999 (commented):
This is a really simple fix, let's merge this.

before

image

after

image

Issue #11284, opened 2023-01-28
remove {{gecko}} macro from translated-content


yin1999:
The macro `{{gecko}}` has been removed from content repo and has been [deprecated](https://github.com/mdn/yari/pull/8071). We should also start to remove this:

- [x] es #11352
- [x] fr #11526
- [x] ja #11397
- [x] ko #11329
- [ ] pt-BR #11527
- [x] ru #11530
- [x] zh-cn #11283
- [x] zh-tw #11283

## process

use regexp `\{\{\s*Gecko\s*\(\s*['"]([^'"]+)['"]\s*\)\s*\}\}` to find and replaced with `Gecko $1`

PR #11530, opened 2023-02-09
ru: Remove {{Gecko}} macro

queengooborg:
This PR removes the `{{Gecko}}` macro from the Russian locale as a part of #11284.


saionaro (approved):
LGTM, thank you!

PR #11533, opened 2023-02-09
es: Remove Web/HTTP/CORS/Errors

queengooborg:
This PR removes the document for `Web/HTTP/CORS/Errors` from the Spanish locale.  This file was just an old copy of the English locale and had no translation.


Graywolf9:
Thank you so much! @queengooborg 

We usually prefer to keep the documents to help the SEO of `es` and make an issue to invite more people to help in the proyect

queengooborg:
Personally, I think that keeping around old, untranslated content is not a good idea.

As far as I know, it actually creates a _negative_ SEO impact when you offer content that's in a different language than reported by the website.  If we describe a page to be in Spanish with `lang` tags and the likes, but it's actually in English, search engines consider the page to be broken/non-inclusive.

Additionally, I don't think that it's helpful for retaining readers.  If a reader comes to MDN web docs (or really, any site) and requests Korean, they're going to want to read the content in Korean and Korean alone, most likely out of necessity.  If, all of the sudden, that reader comes across a page that is labeled as "Korean", but it is actually in English, they're going to think that something's broken with the site.  This may very well cause them to click away and search for another resource instead.

Furthermore, I think that this poses a heavy risk of a new contributor attempting to localize the version of the doc that is _currently_ in said locale, instead of taking the effort to search upstream for the latest version and use that as the original content for translation, a contributor may just use the version that's already in front of them.  (I know I would, if I was doing localization.)  This poses risks of retaining outdated content, discouraging new contributors, or most likely both.

I've asked @Rumyra to chime in on this topic as well.

Rumyra:
If we remove it is there any signal to the `es` group it is then missing and would need to be translated & added back?

Whilst I agree with @queengooborg 's points about SEO, I'm also concerned this might open a can of worms, how many docs are not translated?

Graywolf9:
Thank you so much again @queengooborg and @Rumyra !

About the signal of remove there's not, that's why we rather open a bug so we can fix it, instead of simply deleting the file

`es` is about a year ago unfrozen so there's almost 80% of the content outdated or with English content, we already have a meta issue for this last case, so we are currently working on it

queengooborg:
Wouldn't the fact that there is no corresponding file in the repository count as a signal already, though?  I mean, if a locale is missing its own copy, isn't that a clear indication that a translation is missing and needs to be created?

queengooborg:
I see that this document was just translated in #11635, so closing!

PR #11665, opened 2023-02-16
correct the output of example

yurj26:


yin1999 (approved):
感谢，因为英文文档已经更新了此部分的内容，所以我在此基础上做来一定的修改

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/expm1#using_math.expm1

Issue #11594, opened 2023-02-12
Update all browser compat/spec sections


queengooborg:
In many translated pages, old-style spec/compat data sections are used, from more recent formats such as calling `{{Compat}}` with an argument to as far back as wiki tables and using the `{{spec}}` macro.  We should check all of the pages for these older-style sections and replace them with a translated version of the following:

```md
### Specifications

{{Specifications}}

### Browser compatibility

{{Compat}}
```

This project involves the following tasks:
- Replace compatibility data tables (markdown and HTML) with `{{Compat}}`
- Remove arguments to `{{Compat}}` calls - #16873
- Remove arguments to `{{Specifications}}` calls - complete
- Replace sections with `{{spec}}`, `{{spec2}}` and `{{SpecName}}` with `{{Specifications}}`
- Replace any Specifications sections stating "there is no spec" or similar with `{{Specifications}}` (if the same applies upstream)
- Remove any notes regarding specific browser versions (especially of SpiderMonkey)

Locales:
- [ ] es
  - #11609
  - #11779
  - #11780
  - #11781
  - #11782
  - #11783
  - #11784
  - #11785
  - #11786
  - #11787
  - #11788
  - #16897
- [x] fr
  - #11791
  - #11793
  - #11794
  - #11795
  - #11796
  - #11797
  - #11798
  - #11799
  - #11800
- [x] ja
- [x] ko #11691
- [x] pt-BR
  - #15989
  - #15991
  - #16377
  - #16379
  - #16453
  - #16454
  - #16610
  - #16582
- [x] ru
  - #16503
  - #16504
  - #16505
  - #16674
- [x] zh-cn
- [x] zh-tw

---

Some examples of old-style compat tables and spec tables include:

```md
{{compat("api.Document.pointerLock")}}

---

| Browser           | Suporte         | Notas                                                          |
| ----------------- | --------------- | -------------------------------------------------------------- |
| Internet Explorer | 8               | (IE8) apenas selectores CSS 2.1                                |
| Firefox (Gecko)   | **3.5** (1.9.1) |                                                                |
| Opera             | 10              |                                                                |
| Chrome            | 1               |                                                                |
| Safari (webkit)   | 3.2 (525.3)     | [webk.it/16587](https://bugs.webkit.org/show_bug.cgi?id=16587) |

--- 

## Especificação

- {{spec("https://www.w3.org/TR/selectors-api/","Selectors API Level 1","rec")}}
[...]

PR #11691, opened 2023-02-17
ko replace old spec and compat macro

SnowMarble:
#11594 거의 다 확인한것 같습니다. Specifications 또는 명세서 헤드라인을 거의 다 확인 해보았지만, 한 번 더 검토가 있었으면 좋겠습니다.


wisedog:
오! 대단합니다! 이걸 해내셨군요!

hochan222 (approved):
고생 많으셨습니다. 이상 없는 듯 합니다! 

좋은 하루 보내세요 :)

files/ko/web/api/clipboard_api/index.md
```diff
@@ -38,15 +38,15 @@ navigator.clipboard.readText().then(
...
 
 ### ClipboardEvent
 
-{{Compat("api.ClipboardEvent")}}
+{{Compat}}
 
 ### ClipboardItem
 
-{{Compat("api.ClipboardItem")}}
+{{Compat}}
```
queengooborg, line 49: A little after-merge review: sections like this with multiple interfaces should actually just be replaced with a single `{{Compat}}` call -- the macro (well, "macro") has been enhanced to do this for us if multiple BCD keys are given in the frontmatter keys!

SnowMarble: What a nice catch! Why didn't I think there would be more than one Compat on a page? I think it would be better to just update the entire page. Thanks again😸

queengooborg: There's a lot of little details like this left over from the old wiki days, and it's hard to catch them all -- I had just found out that some `{{Specifications}}` macro calls can also have arguments that need to be removed, apparently, and I hadn't even thought about that until stumbling across it in the Spanish locale!

PR #11852, opened 2023-02-22
fr: remove SeeCompatTable macro for scrollIntoView

ledenis:
### Description

Remove SeeCompatTable macro for french scrollIntoView article

### Motivation

The SeeCompatTable macro is not present in the english article, it should be removed in the french translation

### Additional details

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView


SphinxKnight (approved):
Thanks @ledenis :)

PR #11853, opened 2023-02-23
[zh-cn]: format selector

btea:
### Description

`lint-staged` will automatically format and remove trailing spaces, so I commented it out when committing


yin1999:
We just don't use tailing spaces to present a new paragraph. Please use a `new-line` instead.

yin1999 (approved):

btea:
> We just don't use tailing spaces to present a new paragraph. Please use a `new-line` instead.

I see, thank you.

PR #11855, opened 2023-02-23
Fix typo in `Promise.all`

lucaszhu2zgf:


yin1999 (approved):
Good catch. Welcome aboard :tada: :+1:

PR #11903, opened 2023-02-26
[zh-cn] Regression in ID selectors

jasonren0403:
### Description

I did the update on #11897, but I didn't notice the regression in cssxref macro.

### Motivation

PR #11897 
image
Now
image

### DIFF
--- a/files/zh-cn/web/css/id_selectors/index.md
+++ b/files/zh-cn/web/css/id_selectors/index.md
@@ -20,7 +20,7 @@ CSS **ID 选择器**会根据该元素的 [`id`](/zh-CN/docs/Web/HTML/Global_att
 #id 属性值 { 样式声明 }
 ```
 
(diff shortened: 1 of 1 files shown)


yin1999 (commented):
It's the same, we could just left it as it is.

jasonren0403:
PR #11897 
image
Now
image

yin1999:
> PR #11897 image Now image

Thanks for this. Actually, [the macro in yari](https://github.com/mdn/yari/blob/87c525144b27a3b0d9961b7df4ac5ffda4399ffb/kumascript/macros/cssxref.ejs#L37) should have the same behavior with [other macros](https://github.com/mdn/yari/blob/87c525144b27a3b0d9961b7df4ac5ffda4399ffb/kumascript/macros/Glossary.ejs#L12).

But as for now, we should change the param to avoiding this ):

yin1999 (approved):

PR #11969, opened 2023-03-02
fix incorrect word translation

linshehai:
### Description
correct the wrong word's translation

### Motivation
when I'm reading the document I find the mistake.
use correct word can help people understand the doc.


yin1999 (approved):
Good catch :+1:. Thanks.

PR #11974, opened 2023-03-02
es: Sync HardwareConcurrency with en-US Slug

Graywolf9:
### Description

Fix the slug for the page according to `en-US`

### Motivation

Sync according to https://github.com/mdn/content/pull/6637


JuanVqz (approved):
:1st_place_medal:

Issue #11953, opened 2023-03-01
Web/HTML/Element: All translated pages are missing table content


manjuu-eater:
### MDN URL

https://developer.mozilla.org/docs/Web/HTML/Element

### What specific section or headline is this issue about?

all sections

### What information was incorrect, unhelpful, or incomplete?

Table contents are empty in all translated Web/HTML/Element pages.

This problem is caused by the removed macro `HTMLRefTable`.
Related: https://github.com/mdn/content/pull/24515

This issue is a part of https://github.com/mdn/translated-content/issues/5603 .

### What did you expect to see?

table content

### Do you have any supporting links, references, or citations?

https://developer.mozilla.org/es/docs/Web/HTML/Element
https://developer.mozilla.org/fr/docs/Web/HTML/Element
https://developer.mozilla.org/ja/docs/Web/HTML/Element
https://developer.mozilla.org/ko/docs/Web/HTML/Element
https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element
https://developer.mozilla.org/ru/docs/Web/HTML/Element
https://developer.mozilla.org/zh-CN/docs/Web/HTML/Element
https://developer.mozilla.org/zh-TW/docs/Web/HTML/Element

### Do you have anything more you want to share?

I made PRs for each page.

- [x] #12015
- [x] #12016
- [x] #11993
- [x] #12017
- [x] #12018
- [ ] #12019
- [x] #12020
- [x] #12021

yin1999:
as a part of: #5603

manjuu-eater:
Where is the translated contents referenced from {{HTMLRefTable("...")}} located?
I want to fix this, but I couldn't find translated contents.

I can use [WayBackMachine](https://web.archive.org/web/20230203235743/https://developer.mozilla.org/ja/docs/Web/HTML/Element) content, but I want to use translated source if exists.

yin1999:
> Where is the translated contents referenced from {{HTMLRefTable("...")}} located? I want to fix this, but I couldn't find translated contents.

See mdn/content#24515. We should replace them with markdown tables

manjuu-eater:
Thank you for comment.

Sorry for the ambiguous text.
I just want to know where the already translated text (e.g. Japanese translated text) is placed.

I think translated Japanese text exists (or existed) somewhere because [WayBackMachine](https://web.archive.org/web/20230203235743/https://developer.mozilla.org/ja/docs/Web/HTML/Element) displays translated Japanese.

[...]

PR #12016, opened 2023-03-04
fr: Replaced `{{HTMLRefTable}}` with markdown table

manjuu-eater:
### Description

The `{{HTMLRefTable}}` macro became obsolete, and the entire table content was gone. So, I converted the macros as described below:

1. Get the markdown corresponding to the "Element" cell of each table from the English version
2. Set the markdowns to the "Element" cells in the translated version
3. For each element (=HTML tag), get the first line of the translated version of the page
(specifically, the first line that is not a front-matter, heading, or Lv1 list)
4. Set the retrieved line to the "Description" cell
5. If "Description" cell was empty, fill with en-US description (`{{HTMLRefTable}}` did not seem to have this process)

I am not entirely sure what the `{{HTMLRefTable}}` macro was actually doing.
I assumed it probably did something like the above.
A PR in the Japanese version (#11993) was determined that there was no problem.

I can't understand what the translated text means.
So, I would like the reviewer to do the following tasks:

1. Since I did the conversion semi-mechanically, there may be some strange things in the text. Please check it lightly.
2. Please compare with [WayBackMachine page](http://web.archive.org/web/20230214020459/https://developer.mozilla.org/fr/docs/Web/HTML/Element) and check if there are any significant changes.

I am writing this English text with using a translator. Sorry if there are parts that are difficult to read.

Thank you in advance.

### Additional details

WayBackMachine page: http://web.archive.org/web/20230214020459/https://developer.mozilla.org/fr/docs/Web/HTML/Element
current page: https://developer.mozilla.org/fr/docs/Web/HTML/Element

Relates to #11953


SphinxKnight:
Sorry it took so long @manjuu-eater and thanks again for the PR. I'll be back at it in the coming days (for real :x )

PR #12050, opened 2023-03-07
ko: fix flatmap example

username1103:
### Description

flatmap example is incorrect. so fix it.

### Motivation

This is an incorrect example result.


pje1740 (approved):
고생하셨습니다!

PR #12051, opened 2023-03-07
Typofix

cestfredy:
fix typo


MrDeveloper01010:
The problem was very simple!
it was just that instead of: "Un tableau" it was rather written "Une tableau".

SphinxKnight (approved):
Merci @MrDeveloper01010

PR #12103, opened 2023-03-10
fix(zh-cn): reference to the image in en-US

DreamOfIce:
### Description
Fix broken image of https://developer.mozilla.org/zh-CN/docs/Web/HTML/Element/figure#图像

### Motivation
Sync https://github.com/mdn/content/commit/48294819a58cddae73d21f7c9439605c80da9e12


yin1999 (approved):
Sorry for the late. Since the image corruption problem cannot be solved in the expected future, I am referencing images from en-US as a workround.

PR #12171, opened 2023-03-13
fixed label errors in Chinese ver.

Dar-Xs:
Fixed several code label errors in simplified Chinese translation:

### Description

Fixed several code label errors in simplified Chinese translation.
修复了简体中文翻译中的几处代码标签错误。

### Motivation

This makes the distinction between `code blocks` and other words more correct.
这样会使`代码块`和"其他文字"的区分更加正确。


yin1999 (approved):
Thanks @Dar-Xs. This document is a bit old, so I've made some changes to reflect the current en-US document.

PR #12172, opened 2023-03-13
fix typo in `async function`

CY-Christin:
修复语句不通顺问题。

### Description

Fix the problem of incoherent statements.

### Motivation

Facilitate readers to better understand the meaning of the article.  

### Additional details  

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function


yin1999 (approved):
Actually, I could not see this paragraph in the reflecting en-US document. But it does still worth it.

Thank you @ch634421335. Welcome aboard :tada:

PR #12225, opened 2023-03-16
[fr] Fix encodeURIComponent description for sharp

Aeden-B:
### Description

Update encodeURIComponent description for sharp because the character is encoded.

### Motivation

The documentation is misleading about the result for sharp.


cw118 (approved):
Merci pour cette correction @Aeden-B !

PR #12297, opened 2023-03-21
fix: Console.group doc

eric-lua:
### Description

fix console.group and console.groupCollapsed document details.
Legacy documentation problem: missing documentation and descriptions related to console.group and console.groupCollapsed parameters.


yin1999 (approved):
Hi @eric-lua. As the zh-CN documents are only the translations of en-US document. So, if you want to add more infomation, please create a PR for the [reflecting en-US document](https://developer.mozilla.org/en-US/docs/Web/API/console/group). Thanks

eric-lua:
> Hi @eric-lua. As the zh-CN documents are only the translations of en-US document. So, if you want to add more infomation, please create a PR for the [reflecting en-US document](https://developer.mozilla.org/en-US/docs/Web/API/console/group). Thanks

I see. Thank you.

Issue #12128, opened 2023-03-11
Incorrect l10n.sourcecommit meta data


hochan222:
## Issue

The hash value of the source commit in meta data is incorrect. In most cases, there is no hash in the history for the same file in the content repo.

### ja

```txt
ja/glossary/falsy: ec5cdbe5e3e0198f60a537c7d4a9d31edab50920
ja/web/api/htmlelement: d954e1df58f18f15181bfe33ac64dfad9627f485
ja/web/api/mediastream_recording_api: 60c1df4b8b6ed30b2baec49ec3e21a1648a0ba86
ja/web/api/pointerevent: 708baf34eabb75789bcd3314a6879da3702024d1
ja/web/api/serviceworkerglobalscope: e540d79b5bc8893cdcd0b65ec498a6e2c22cfcfa
ja/web/api/uievent: 70ab349bf0271762d5d8d03b3cdf723fd7d13326
ja/learn/tools_and_testing/understanding_client-side_tools/overview: 99eb07dc4987d2c2f4982e81c046c9abbe827665
ja/mdn/community/contributing/security_vulnerability_response: 73dd350fd93be16bee3b9a6b860757265209b4b7
ja/web/http/resources_and_specifications: 07b16eba30c0f343199170145b57440f66501453
ja/mdn/writing_guidelines/howto/images_media: 294dc5cb436d4f86ed789be8fded3ff615768fb0
ja/mdn/writing_guidelines/page_structures/macros: 73dd350fd93be16bee3b9a6b860757265209b4b7
ja/mdn/writing_guidelines/page_structures/page_types: 73dd350fd93be16bee3b9a6b860757265209b4b7
ja/web/javascript/guide/expressions_and_operators: 6ebc97ff44612ef04d2cde60bfe9049bc8482bb5
ja/web/api/document/pointerlockerror_event: d4619276d67ca1ee02bd964d884ec52c2aa2ad4by
ja/web/api/element/lostpointercapture_event: 3efb6cc0c72738d127ccd0829196b9dd24a726a6
ja/glossary/html: 2ac7419230e51f9bca905f4ef14d9e44c4dd6dba
ja/web/api/html_drag_and_drop_api: 7e43ff6a4a712f9e4477a70c453ba116a4dd6601
ja/mdn/community/contributing/translated_content: 73dd350fd93be16bee3b9a6b860757265209b4b7
ja/web/api/element/pointerleave_event: 708baf34eabb75789bcd3314a6879da3702024d1
[...]

PR #12315, opened 2023-03-21
es: Fix commit hash

Graywolf9:
### Description

Fix the commit hash that show the parity with en-US

Relates to https://github.com/mdn/translated-content/issues/12128


JuanVqz (approved):
What is the supported length? the SHAs seems to be the same but shorter

Graywolf9:
The hash is repeated I reported here: https://github.com/liondocs/cli/issues/69

PR #12339, opened 2023-03-22
fix example of Atomics

Huauauaa:


yin1999 (approved):
Thanks. Welcome aboard 🎉

PR #12370, opened 2023-03-24
remove broken link

RexChien:
Because target page no longer exists, I recommend removing this link.


yin1999 (approved):
This document is out-of-date for both localized and en-US locals. But I still think it's a good idea to remove the broken links.

PR #12384, opened 2023-03-25
Translate "Non-cryptographic uses of SubtleCrypto"

mikecat:
### Description

Non-cryptographic uses of SubtleCrypto を新規翻訳。


mfuji09 (approved):
OKです。

Issue #9638, opened 2022-11-01
Has English content [es]


Graywolf9:
El siguiente listado de archivos fue creado a partir de una búsqueda por patrones.

- si crees que algún archivo no debería estar aquí por favor indicarlo en los comentarios
- si deseas corregir algún archivo por favor indicarlo en los comentarios para asignarlo, o
- crea tu pull request y menciona este issue utilizando "Relates to #" 

Nota: Se recomienda hacer la actualización completa, pero si se desea puede solo corregirse la parte en inglés.

Archivos:

- [ ] https://developer.mozilla.org/es/docs/conflicting/web/accessibility/aria_64707ba1917a56654679cbe273e2f4ea
- [ ] https://developer.mozilla.org/es/docs/conflicting/web/api/element/click_event
- [ ] https://developer.mozilla.org/es/docs/conflicting/web/html/global_attributes/contenteditable
- [ ] https://developer.mozilla.org/es/docs/conflicting/web/javascript/reference/global_objects/function/tostring
- [ ] https://developer.mozilla.org/es/docs/games/tutorials/2d_breakout_game_phaser
- [ ] https://developer.mozilla.org/es/docs/games/tutorials/2d_breakout_game_phaser/initialize_the_framework
- [ ] https://developer.mozilla.org/es/docs/games/tutorials/2d_breakout_game_phaser/move_the_ball
- [ ] https://developer.mozilla.org/es/docs/games/tutorials/2d_breakout_game_phaser/scaling
- [ ] https://developer.mozilla.org/es/docs/games/tutorials/2d_breakout_game_phaser/the_score
- [ ] https://developer.mozilla.org/es/docs/games/tutorials/2d_breakout_game_phaser/win_the_game
- [ ] https://developer.mozilla.org/es/docs/games/tutorials/2d_breakout_game_pure_javascript
- [ ] https://developer.mozilla.org/es/docs/games/tutorials/html5_gamedev_phaser_device_orientation
- [ ] https://developer.mozilla.org/es/docs/glossary/accessibility
- [ ] https://developer.mozilla.org/es/docs/glossary/argument
- [ ] https://developer.mozilla.org/es/docs/glossary/asynchronous
- [ ] https://developer.mozilla.org/es/docs/glossary/atag
- [ ] https://developer.mozilla.org/es/docs/glossary/base64
- [ ] https://developer.mozilla.org/es/docs/glossary/canvas
- [ ] https://developer.mozilla.org/es/docs/glossary/compile
[...]

PR #12493, opened 2023-04-01
[es] Update learn/html/introduction_to_html/marking_up_a_letter

dawei-wang:
Relates to mdn/translated-content#9638


Graywolf9 (approved):
Thanks for all your time and work!

PR #12589, opened 2023-04-06
Correct French sentence for :first-child use

flora-le:
Translation in French is wrong
" si je veux sélectionner le premier descendant de chaque élément `<article>` pour le mettre en gras" means "if I want to select the first descendant of each <article> element to make them bold". I understood that it means the first child of article would be bold, but when I did the test Selectors (CSS), task 3, it didn't give the expected result from the French documentation.

The English one is correct though, I've just checked it, so I based my French translation on it.

### Description

Translation in French was wrong. So when I did [task3 in Selectors test](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors/Selectors_Tasks), the result wasn't what was explained in the French sentence.
I corrected the French translation to match more or less the English one which was fine.

### Motivation

French translation was mistranslated. The sentence was changed to match the expected results, to be able to do task 3 for French speakers. I hope it is correct, at least I undertand it just like the English version.


cw118 (approved):
Merci encore @flora-le 💯

PR #12594, opened 2023-04-06
Typofix/clarification for `webRequest.BlockingResponse`

Lou8is:
### Description

Changes the french explanation that goes with the "cancel" value so it is more clear that is is restricted to some listeners (and not the other way around).

### Motivation

I struggled to understand whether I could use the "cancel' value in the "onBeforeRequest" listener, as the french translation was ambiguous. The original English text is clear on that point so I decided to correct the french one.

### Additional details

None.

None that I know of.


cw118 (approved):
Merci beaucoup pour cette correction @Lou8is !

PR #12596, opened 2023-04-06
Added the link to the glossary of "protocol"

nathalia-nobrega:
### Description

The "TLS" definition was **lacking the link** to the "Protocol"'s. I added it using _**{{Glossary("Protocol", "protocolo")}}**_ at line 7.

### Motivation

Some readers, like myself, could've thought that the documentation was lacking the defition of "protocol", which is an important one given the context of the page.


josielrocha (approved):

josielrocha:
Obrigado pela ajuda @nathalia-nobrega!

PR #12616, opened 2023-04-07
Added missing space, code/bolded second method mention

TheYuriG:
Minor nitpick changes that fix the formatting on the page for pt-BR language.

The array method lacked a space after the double asterisk, which caused the text to not bold when converting Markdown to HTML. This PR fixes that issue.

At the bottom, I've also bolded and coded another mention of the array method, but it seems like the pages usually only bold the first mention of the method and wrap the other mentions in a simple code block, so I guess I'll only code that mention in a bit.
EDIT: It's only coded now, not bolded anymore.

This change will make the markdown formatting actually work for visitors.

### DIFF
--- a/files/pt-br/web/javascript/reference/global_objects/array/shift/index.md
+++ b/files/pt-br/web/javascript/reference/global_objects/array/shift/index.md
@@ -7,7 +7,7 @@ slug: Web/JavaScript/Reference/Global_Objects/Array/shift
 
 ## Sumário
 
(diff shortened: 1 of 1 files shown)


josielrocha (approved):

josielrocha:
Thank you for your help @TheYuriG!

TheYuriG:
Happy to be able to help a bit. :)

PR #12829, opened 2023-04-19
chore(zh-cn): Remove tag mentions from translation-guide

jasonren0403:
### Description

Currently there are no tags in any of English content.


yin1999 (approved):
Thanks for the cleanup :)

Issue #12578, opened 2023-04-05
Resolve MacroLiveSampleErrors in translated-content


caugner:
While testing out https://github.com/mdn/yari/pull/8571, Sentry collected the following `MacroLiveSampleError`s.

The error is: `MacroLiveSampleError within ... (unable to find any live code samples for ...)`

## ja

- [ ] /ja/web/svg/attribute/onclick/index.md, line 25 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/onclick)
- [ ] /ja/web/svg/attribute/points/index.md, line 25 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/points)
- [ ] /ja/web/svg/attribute/r/index.md, line 33 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/r)
- [ ] /ja/web/svg/attribute/marker-mid/index.md, line 26 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/marker-mid)
- [ ] /ja/web/svg/attribute/fill/index.md, line 39 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/fill)
- [ ] /ja/web/svg/attribute/cy/index.md, line 21 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/cy)
- [ ] /ja/web/svg/attribute/d/index.md, line 20 column 7 ("example" within /ja/docs/Web/SVG/Attribute/d)
- [ ] /ja/web/svg/attribute/fill-opacity/index.md, line 30 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/fill-opacity)
- [ ] /ja/web/svg/attribute/href/index.md, line 19 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/href)
- [ ] /ja/web/svg/attribute/id/index.md, line 21 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/id)
- [ ] /ja/web/svg/attribute/cx/index.md, line 21 column 7 ("topexample" within /ja/docs/Web/SVG/Attribute/cx)
- [ ] /ja/web/css/width/index.md, line 98 column 7 ("percentage" within /ja/docs/Web/CSS/width)
- [ ] /ja/web/css/width/index.md, line 109 column 7 ("max-content" within /ja/docs/Web/CSS/width)
- [ ] /ja/web/css/width/index.md, line 119 column 7 ("min-content" within /ja/docs/Web/CSS/width)
[...]

PR #12830, opened 2023-04-19
fix(zh-tw): update the broken live sample

yin1999:
### Description

fix the broken live sample

Part of: #12578


yin1999 (commented):
as there is only typo fixes, and example fixes. Going to merge it

PR #12907, opened 2023-04-22
chore(zh-tw): run prettier on games and glossary folder

yin1999:
### Description

run prettier on games and glossary folder

as the first step of mdn/mdn#362 for zh-TW locale


.prettierignore
```diff
@@ -2,3 +2,7 @@ build/
 
 # Ignore markdown files till full pass is made on each folder
 *.md
+
+# The following paths have had a full pass
+!/files/zh-tw/games/**/*.md
+!/files/zh-tw/glossary/**/*.md
```
queengooborg, line 8: Upstream, we've switched to using .prettierignore as a breakdown list rather than telling it to ignore everything and then un-ignoring specific folders.  It might be nice to do so here as well!

queengooborg (approved):
These changes look good to me, they're all super simple (primarily wrapping links with parenthesis in them in angle brackets)!

.prettierignore
yin1999: I've made the changes in f58ed696ac11bdfb829d68696a5a6cd3113b83bb, could you have a double check on this :)

queengooborg (approved):
The breakdown list looks good, and the additional changes look good too!

Issue #12579, opened 2023-04-05
Image tags without src


caugner:
While testing out https://github.com/mdn/yari/pull/8571, Sentry collected the following errors regarding `<img>` tags without a `src` attribute.

## es

- [x] `In /es/docs/Web/API/Element/clientTop there's an img tag without src (<img alt="Image:clientTop.png">)`
- [x] `In /es/docs/Web/API/Element/clientLeft there's an img tag without src (<img alt="Image:clientLeft.png">)`
- [x] `In /es/docs/Web/API/Element/clientLeft there's an img tag without src (<img alt="La barra de desplazamiento vertical en el lado izquierdo, aparece cuando la propiedad layout.scrollbar.side tiene el valor 1 o 3">)`

## fr

- [x] `In /fr/docs/Mozilla/Add-ons/WebExtensions/Firefox_differentiators there's an img tag without src (<img alt="containers.png">)` https://github.com/mdn/content/pull/26242
- [ ] `In /fr/docs/Learn/Getting_started_with_the_web/HTML_basics there's an img tag without src (<img src="" alt="Mon image de test">)` https://github.com/mdn/translated-content/pull/12818

## ko

- [ ] `In /ko/docs/Web/SVG/Tutorial/SVG_and_CSS there's an img tag without src (<img alt="SVG demonstration">)`
- [ ] `In /ko/docs/Web/API/Window/toolbar there's an img tag without src (<img alt="Image:Modify_any_open_window_dialog.png">)`
- [ ] `In /ko/docs/orphaned/Web/SVG/SVG_1.1_Support_in_Firefox there's an img tag without src (<img alt="그림:그림-text-fill-stroke.png">)`

SphinxKnight:
https://github.com/mdn/translated-content/pull/12818 has been opened for the two docs in `fr`

alattalatta:
ko's fixed by #12488 and #12648.

Graywolf9:
`es` attended in https://github.com/mdn/translated-content/pull/12964

PR #12964, opened 2023-04-24
es: Fix img tags without src

Graywolf9:
### Description

Fix img tags without src

### Additional details

The images are no longer in the `en-US` version, so they were deleted

Relates to #12579


JuanVqz (approved):
@Graywolf9 looks good!

PR #12994, opened 2023-04-27
zh-CN: "支持"错打成"持支"

wang1xiang:


yin1999 (commented):
Thanks. Welcome aboard

PR #13034, opened 2023-04-29
fix prototype emphasis in Japanese translation

ksrnnb:
### Description

I've updated the emphasis on the term "プロトタイプ**" to show correctly as "**プロトタイプ**".

### Motivation

This change ensures that important keywords are properly emphasized, helping readers quickly identify key concepts within the content.


mfuji09 (approved):
OKです。ありがとうございます。

PR #13109, opened 2023-05-04
[zh-cn]: Fix the broken link

zhoushengdao:
### Description

Fix the broken link. See:
https://github.com/mdn/translated-content/blob/3b6592edc55bc48ebee6df023d3ebaf304909ec0/files/zh-cn/web/html/element/input/range/index.md?plain=1#L156


yin1999 (approved):
:+1:

PR #13171, opened 2023-05-08
[zh-CN]: Fix missing tailing spaces

050644zf:
### Description

Add spaces after bold symbols.

### Motivation

Missing spaces cause the md style wasn't applied properly


yin1999 (approved):
Thanks @050644zf. Welcome aboard 🎉

PR #13176, opened 2023-05-08
`fr` - Initial translation for Web/HTML/Constraint_validation

SphinxKnight:
This is an initial translation for https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation which is missing in `‘fr` atm


cw118 (approved):
Merci encore Sphinx :)

PR #13194, opened 2023-05-10
remove css-animations part <浏览器兼容性>, Sync with en documents

nnecec:
### Description

The CN translation ,  "浏览器兼容性"  shows `{{Compat}}`. I checked EN page, there is no "Browser compatibility" part.

image

and [the EN source code](https://github.com/mdn/content/blob/main/files/en-us/web/css/css_animations/index.md) still has no "Browser compatibility"


jasonren0403 (approved):
👍

PR #13199, opened 2023-05-10
`fr` - Initial translation for HTML/Element/script/type and HTML/Element/script/type/importmap

SphinxKnight:
Some missing pages in French :)


cw118 (approved):
🎉

PR #13355, opened 2023-05-22
zh-CN: sync the jpg files from en-US

hanyujie2002:


yin1999 (changes requested):
Please check for associated documents instead of adding all images to translated-content in one go. Additionally, the platform automatically redirects missing images to en-US (for most of cases). I think not all the images is needed here.

See the comment here: https://github.com/mdn/translated-content/pull/12103#issuecomment-1464057720

PR #13362, opened 2023-05-22
chore: change https -> http for syntax highlighting

bsmth:
There's a few build errors for these:

* changing to `http` instead of `https`
* using `http example-(bad|good)` instead of `example-bad`:

````md

```example-bad
// this is not a detected language
```


```js example-bad
// better
```

````


potappo (approved):
Thank you, LGTM.

PR #13371, opened 2023-05-23
chore(zh): run ptettier format on `/docs`

yin1999:
### Description

run ptettier format on `/docs`

mdn/mdn#362

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -35,6 +35,5 @@ build/
 /files/zh-cn/**/*.md
 
 # zh-tw
-/docs/zh-tw/**/*.md
 /files/zh-tw/web/api/[a-o]*/**/*.md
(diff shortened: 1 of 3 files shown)


yin1999:
Hi @jasonren0403, could you have a look on this, I've wrapper those examples with inline-code as a workround of [the problem metioned here](https://github.com/mdn/mdn/issues/362#issuecomment-1548838210).

jasonren0403:
> Hi @jasonren0403, could you have a look on this, I've wrapper those examples with inline-code as a workround of [the problem metioned here](https://github.com/mdn/mdn/issues/362#issuecomment-1548838210).

Yeah, I was using prettier 2.8.8 and Jetbrains IDEA and prettier seems to process it through. We may add backticks as a workaround.

queengooborg (approved):

PR #13397, opened 2023-05-25
[zh-CN] sync Notification.actions with en-US

skyclouds2001:
### Description

sync Notification.actions with en-US

### Motivation

sync Notification.actions with en-US

### Additional details

[https://developer.mozilla.org/en-US/docs/Web/API/Notification/actions](https://developer.mozilla.org/en-US/docs/Web/API/Notification/actions)
[https://developer.mozilla.org/zh-CN/docs/Web/API/Notification/actions](https://developer.mozilla.org/zh-CN/docs/Web/API/Notification/actions)


yin1999 (approved):
Great :100:

PR #13408, opened 2023-05-26
Fix broken markdown

shiomiyan:
### Description

Fixed broken markdown.

### Motivation

With respect to content that should have been displayed in BOLD, the asterisks were being displayed as is.

Also, content that should have been italicized was not italicized.


potappo (approved):
OK です。

PR #13409, opened 2023-05-26
[zh-cn]: Update translation of queueMicrotask

zhang-stone:


yin1999 (approved):
Thanks @zhang-stone, it's better than before. Welcome aboard :tada: ;)

PR #13433, opened 2023-05-27
fix punctuation mark and term in zh-tw

hms5232:
path: web/http/status/301

### Description

fix punctuation mark and term in zh-tw

### Motivation

1. punctuation mark should be fullwidth
2. "redirect" is "重新導向" in Taiwan. "重定向" is used in China not Taiwan.

### Additional details

None.

None.


yin1999 (approved):
Good catch @hms5232!

PR #13500, opened 2023-05-30
chore(fr): add HSTS redirect

caugner:
### Description

Redirect a frequently hit URL that 404s.

### Motivation

The URL gets about 700 weekly hits from https://france-visas.gouv.fr/, and it originally had the HSTS docs: https://web.archive.org/web/20200927142114/https://developer.mozilla.org/fr/docs/S%C3%A9curit%C3%A9/HTTP_Strict_Transport_Security

It is linked from here (currentlly in maintenance): https://web.archive.org/web/20221201140302/https://www.france-visas.gouv.fr/web/france-visas/compatibilite-navigateur
image


SphinxKnight (approved):
Thanks a lot @caugner :) much appreciated :) 
I'm not sure why there was no redirect in the first place but let's fix this first.

PR #13593, opened 2023-06-05
Update index.md

starrovoitov:
### Description
Updated translation. There was a mistake in the paragraph about the response header

### Motivation
The section talks about HTTP response headers, but there was an error in the text, which appears to have been taken from the text about request headers. The text was updated accordingly


saionaro (approved):
@starrovoitov привет! Спасибо за вклад в MDN!

PR #13648, opened 2023-06-10
fix(zh-cn): `zh_CN` to `zh-CN`

yin1999:
### Description

`zh_CN` to `zh-CN`


jasonren0403:
seems that #13643 already fixed them:(

PR #13696, opened 2023-06-11
Update index.md

BrunoDVolpe:
### Description

Update to some miswriting words and translated two comments in examples.

### Motivation

I'm studying through this tutorial and I think it may be useful applying these changes. They are grammatically correct and may help other Brazilian who are reading in Portuguese to have a better understanding.


josielrocha (approved):
LGTM ;)

PR #13718, opened 2023-06-13
[Fix] Fix French translate string in The_head_metadata_in_HTML

tisma95:
https://developer.mozilla.org/fr/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML

### Description

Fix the French translate string in https://developer.mozilla.org/fr/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML


SphinxKnight (approved):
merci @timael95 :)

PR #13842, opened 2023-06-20
Update index.md

ShylockWu:
### Description
The onmessage function of EventSource will only be triggered when an untyped sse request is made. The official website does not explain it, which is very misleading.

### Motivation
There are deficiencies in the interpretation of EventSource's related methods.The onmessage function of EventSource will only be triggered when an untyped sse request is made. The official website does not explain it, which is very misleading.

### Additional details
[EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/EventSource)

### DIFF
--- a/files/zh-cn/web/api/eventsource/message_event/index.md
+++ b/files/zh-cn/web/api/eventsource/message_event/index.md
@@ -57,7 +57,7 @@ evtSource.addEventListener("message", (e) => {
 ```
 
 ### onmessage 的等价写法
-
+仅仅在接收无类型消息的时候，才会通过onmessage处理函数。
 ```js
 evtSource.onmessage = (e) => {
   const newElement = document.createElement("li");


files/zh-cn/web/api/eventsource/message_event/index.md
```diff
@@ -57,7 +57,7 @@ evtSource.addEventListener("message", (e) => {
 ```
 
 ### onmessage 的等价写法
-
+仅仅在接收无类型消息的时候，才会通过onmessage处理函数。
```
yin1999, line 60: Hi @shylockWu. We only "[translate](https://github.com/mdn/translated-content/blob/main/docs/zh-cn/translation-guide.md)" en-US documents, and this line is not existed [here](https://github.com/mdn/content/blob/03ba2dad6861d4d58be6be41c2401dab7a604a75/files/en-us/web/api/eventsource/message_event/index.md?plain=1#L62-L71). I'm sorry we could not merge this (:

PR #13931, opened 2023-06-25
[zh-cn] The result of HTML example does not match the description

lensferno:
### Description

The description of `overflow-y:scroll` is reversed from that of `overflow-y:scroll` in HTML example, I swapped them.


yin1999 (approved):
Good catch. Welcome aboard @lensferno :tada:

PR #13940, opened 2023-06-26
Typofix on Cascade, spécificité et héritage

qmorrier:
Basic typo, deux règles takes an s

### Motivation
Anytime I see a mistake, I try to get it fixed (Boy Scouting Rule :))


cw118 (approved):
Thanks very much @qmorrier !

PR #13954, opened 2023-06-27
fix: BigInt typo

Meowu:
### Description
fix typo: BitInt -> BigInt


yin1999 (approved):
Good catch. Thank you :tada:

PR #13977, opened 2023-06-29
[ru] add missing symbol in contraints description

ad21ewda1231dsa123das12:
### Description

just add missing symbol in param name

### Motivation

purity of the text. maybe with a quick read the reader will misunderstand the information


saionaro (approved):
@themetronome привет! Спасибо за вклад в MDN!

PR #14048, opened 2023-07-05
updated autoDiscardable translation

Mozinet-fr:
updated autoDiscardable pref translation as it was added in Firefox Nightly, and corrections


cw118 (approved):
Merci beaucoup @Mozinet-fr 🙏

PR #14057, opened 2023-07-06
DO NOT MERGE Testing for pr13663

Lou8is:
Some tests to make pr13663 pass


cw118:
Hello @Lou8is, I believe you are correct about the use of deprecated parameters for EmbedLiveSample. I have asked about [removing the `$4` parameter on the en-US page](https://github.com/mdn/content/issues/27801), which should help us fix the issue on the translated pages as well.

PR #14100, opened 2023-07-09
pt-br: Remove {{bug}} macro

queengooborg:
This PR removes the `{{bug}}` macro and replaces it with a bare Markdown link within the Portuguese Brazilian files.  Note: this skips over files that are partially translated or untranslated.


josielrocha:
Hi @queengooborg!

Thank you a lot for your help.
I think I found a little mistake in your PR.. It looks like when the bug ids was written between quotes (`"`) everything is fine, but when it is not the first character is missing `{{bug(123)}}` becomes `[Link title](https://bugzil.la/23)`.

Could you please make those changes?

image

queengooborg:
Whoops, thanks for catching that!  Looks like I had a regex issue (I used `\w` instead of `\s`), it has now been fixed!

josielrocha (approved):
LGTM ;)

PR #14171, opened 2023-07-13
[zh-cn]: fix typo

Rainyrou:
### Description

fix translate typo（extra symbol）and the log statement should be consistent with the output

### Additional details

The MDN documentation has helped me a lot.


yin1999 (approved):
Good catch! Welcome aboard :tada:

PR #14195, opened 2023-07-14
Integrate Prettier into workflows and commands (sync w/ upstream)

queengooborg:
This PR integrates Prettier into the GitHub Actions workflows, package.json commands, and Git precommit hooks.  This synchronizes the workflows with mdn/content.


.github/workflows/markdown-lint-fix.yml
```diff
@@ -40,7 +40,7 @@ jobs:
 
       - name: Lint markdown files
         run: |
-          yarn markdownlint-cli2-fix "**/${{ matrix.lang }}/**/*.md"
+          yarn fix:md "**/${{ matrix.lang }}/**/*.md"
```
nschonni, line 43: This would run on everything now, as the old way would use the glob, but the script target will not

queengooborg: Ah, right -- would you be down to submit a PR to fix that up?  I'm currently in the middle of another project that I can't exactly `git stash` at the moment!

PR #14196, opened 2023-07-14
ja: format README.md with Prettier

queengooborg:
This PR formats the Japanese readme with Prettier.


queengooborg:
(Self-merging this because the file is already supposed to remain formatted with Prettier.)

PR #14201, opened 2023-07-14
ru: Fix issue reporting link

queengooborg:
This PR replaces the link to create a new issue in the old `mdn/sprints` repository, changing it to point to the `mdn/translated-content` repository instead.

**This is a temporary fix.  The permanent fix would be to rewrite the section to explain the Report links provided on every page, but I cannot make this change as I do not speak Russian.**


sashasushko (approved):
LGTM 🔥 Lets merge this and I will rewrite this section after

queengooborg:
Thank you very much, @sashasushko!

PR #14207, opened 2023-07-14
ru: Replace old-style spec/compat tables

queengooborg:
This PR replaces the old-style spec and compat tables with modern ones, removing compatibility notes from the documentation wherever found.

Note: the translation guide doesn't have a set definition for "Browser compatibility", so I used the first one I found ("Совместимость с браузерами").  If this is the wrong term to use, please let me know or push changes directly to this branch.


sashasushko (approved):

sashasushko:
Thanks for helping 🤝

queengooborg:
Of course, and thank you for the quick review and merge as well!

PR #14221, opened 2023-07-14
zh-cn: Replace smart quotes

queengooborg:
This PR replaces smart quotes with normal quotes for the Simplified Chinese locale.


yin1999 (commented):
See: https://github.com/mdn/translated-content/pull/14222#pullrequestreview-1531162782

PR #14222, opened 2023-07-14
zh-tw: Replace smart quotes

queengooborg:
This PR replaces smart quotes with normal quotes for the Traditional Chinese locale.


yin1999 (commented):
Hi @queengooborg. Actually, in zh-TW we would [use `「」`/`『』` rather than `“”` or `""`](https://github.com/mdn/translated-content/blob/main/docs/zh-tw/translation-guide.md#%E5%B8%B8%E8%A6%8B%E9%8C%AF%E8%AA%A4). Is there any rule could replace `“”` with `「」` for zh-TW.

While for zh-CN document, we [use `“”` instead of `""`](https://github.com/mdn/translated-content/blob/main/docs/zh-cn/translation-guide.md#%E5%B8%B8%E8%A7%81%E4%B8%AD%E8%8B%B1%E6%96%87%E6%A0%87%E7%82%B9).

I'm not familiar with prettier, could you provide some suggestion, or we should disable related rules on l10n-zh first?

queengooborg:
Hey @yin1999 -- the rule on quotes is actually from Markdownlint, not Prettier, which may give us more flexibility!

Admittedly though, I'm unsure how to override rules per folder.  I'm guessing that we put a Markdownlint config file in the appropriate folder, but I'm not sure about that.

In regards to the quotes, zh-tw should be alright since that quote style wont be affected by our rules!

For zh-cn, though, how important is it to use smart quotes?  It is just a nice-to-have, or is it actually critical for the locale?  One of the big reasons why the no-smart-quote rules are in place are to make sure that smart quotes aren't used in code samples (when normal quotes are needed), and I've noticed this issue in other locales already.  If possible, I'd love to get the same assurances in place for all locales!

yin1999:
> For zh-cn, though, how important is it to use smart quotes? It is just a nice-to-have, or is it actually critical for the locale? One of the big reasons why the no-smart-quote rules are in place are to make sure that smart quotes aren't used in code samples (when normal quotes are needed), and I've noticed this issue in other locales already. If possible, I'd love to get the same assurances in place for all locales!

Unfortunately, this comes from the [government's punctuation usage guidelines](http://www.moe.gov.cn/ewebeditor/uploadfile/2015/01/13/20150113091548267.pdf). And we are educated to use smart-quote (:

image

> we should use `“”` or `‘’`

queengooborg:
Gotcha...argh.  Alright, I'll look into disabling the rule specifically for the Chinese locales!

PR #14220, opened 2023-07-14
ru: Replace smart quotes

queengooborg:
This PR replaces the smart quotes with normal quotes for the Russian locale.

PR #14219, opened 2023-07-14
pt-br: Replace smart quotes

queengooborg:
This PR replaces smart quotes with normal quotes in the Brazilian Portuguese locale.

PR #14217, opened 2023-07-14
ja: Replace smart quotes

queengooborg:
This PR replaces smart quotes with normal quotes for the Japanese locale.

PR #14216, opened 2023-07-14
fr: Replace smart quotes

queengooborg:
This PR replaces smart quotes with normal quotes for the French locale.

PR #14223, opened 2023-07-14
Enable linting rules to prevent smart quotes

queengooborg:
This PR enables the linting rules to prevent smart quotes in translated content.

Depends on the following to be merged first:
- [x] https://github.com/mdn/translated-content/pull/14215
- [x] https://github.com/mdn/translated-content/pull/14216
- [x] https://github.com/mdn/translated-content/pull/14217
- [x] https://github.com/mdn/translated-content/pull/14218
- [x] https://github.com/mdn/translated-content/pull/14219
- [x] https://github.com/mdn/translated-content/pull/14220
- [x] https://github.com/mdn/translated-content/pull/14244


bsmth:
I've taken myself off as reviewer while it's in draft, but feel free to add me again or @ me when it's ready for a look 👀

caugner:
@queengooborg Can you explain the issue with quotes like `“` and `‘`?

Context: I'm working on the German locale, and German uses both of these (`„Sie hat ‚Hallo Welt!‘ gesagt.“`).

queengooborg:
See https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#apostrophes_and_quotation_marks for the writing guideline that disallows smart quotes -- basically, writers have accidentally used smart quotes in code snippets which has caused many issues with code examples.  By disabling them outright, the problem is eliminated altogether!

We did enable smart quotes for Simplified Chinese, however, because it is grammatically correct to use smart quotes in that locale.  We could also do the same for German/other locales if it is needed!

caugner:
Makes sense, and I didn't know I could override the rules, that's what I have done now. Thank you!

PR #14224, opened 2023-07-14
zh: Add Markdownlint ignores for bad vs. good examples

queengooborg:
This PR adds an ignore for Markdownlint for example blocks showing bad vs. good syntax.  The Markdownlint rules that will be enabled are meant to remove the bad examples, so they would be replaced otherwise.


yin1999 (approved):
LGTM. Thanks 🎉

PR #14230, opened 2023-07-14
zh-cn: format translation glossary with Prettier

queengooborg:
This PR formats the translation glossary for Simplified Chinese using Prettier.


yin1999:
Hi @queengooborg. Since we would add notes to some terms here, and term translations may be added frequently. I'm not sure enabling prettier formatting for this file is a good idea.

/cc @jasonren0403

queengooborg:
Hmm, I personally feel that all documentation should follow formatting rules, including translator guides.  Would this document be updated so often that Prettier would become cumbersome?

yin1999 (approved):
Let's merge it now, if it's a problem (in the future), we can disable the formatting of the table in the document.

PR #14233, opened 2023-07-14
es: Remove trailing punctuation in headers

queengooborg:
This PR removes trailing punctuation within headers for the Spanish locale.

Note: one header is also modified as it contains text it shouldn't.


Graywolf9 (approved):
Thank you so much!

PR #14225, opened 2023-07-14
Replace double-dash with m-dash

queengooborg:
This PR replaces all instances of a double-dash (--) with an m-dash (—) throughout translated content.  This synchronizes formatting with upstream.

PR #14248, opened 2023-07-15
zh: Replace double-dash with a double em dash

queengooborg:
This PR replaces all instances of a double dash (--) with a double em-dash (——) throughout the Chinese locales.  This also performs additional grammatical tweaks as requested by @yin1999 during the initial code review.

Cherry-picked from #14225.


yin1999 (approved):
Thanks @queengooborg. We've just resolve the mixed text problem in file `files/zh-cn/web/api/webvr_api/concepts/index.md` with PR #14250.

We can go ahead now :)

PR #14256, opened 2023-07-15
ko: Make sure localized links are used rather than /en-US/* (part 2)

queengooborg:
This PR fixes all of the links within the Korean locale so that they no longer point to English content. This is part 2 of 4.


hochan222 (approved):
LGTM :)

PR #14262, opened 2023-07-15
pt-br: Make sure localized links are used rather than /en-US/*

queengooborg:
This PR fixes all of the links within the Portuguese Brazilian locale so that they no longer point to English content.


josielrocha (approved):
LGTM ;)

PR #14263, opened 2023-07-15
ru: Make sure localized links are used rather than /en-US/* (part 1)

queengooborg:
This PR fixes all of the links within the Russian locale so that they no longer point to English content. This is part 1 of 4.


sashasushko:
@queengooborg thanks a lot for helping. There are strange [build errors](https://github.com/mdn/translated-content/actions/runs/5562740614/jobs/10161229243?pr=14263) with `EmbedLiveSample`. Do you have any idea why them appear? Because I don't see changes in those parts of code.

queengooborg:
Thanks for asking!  While we do not modify the macros, we do modify the files that contain them.  Since the CI test runs on all the modified files, we are notified that there are issues in the files even if the PR itself is not the source of said issues.

While we should address these issues at some point, I'd say that they are safe to ignore in this PR!

PR #14271, opened 2023-07-16
zh-cn: Remove untranslated Mozilla/Firefox/Releases/12

queengooborg:
This PR removes the untranslated Mozilla/Firefox/Releases/12 from the Simplified Chinese locale.


yin1999 (approved):
Thanks

PR #14275, opened 2023-07-16
zh-tw: Remove untranslated documents

queengooborg:
This PR removes a few untranslated documents in the Traditional Chinese locale.


yin1999 (approved):
Thanks

PR #14290, opened 2023-07-16
zh-cn: Remove untranslated files

queengooborg:
This PR removes two untranslated files from the Simplified Chinese locale.


yin1999 (approved):
:+1:

PR #14291, opened 2023-07-16
zh-cn: Format /games using Prettier

queengooborg:
This PR formats the `/games` folder for the Simplified Chinese locale using Prettier.


yin1999 (approved):
Thank you!

PR #14292, opened 2023-07-16
zh-cn: Remove more untranslated files

queengooborg:
This PR removes another untranslated file from the Simplified Chinese locale.


queengooborg:
I'm actually going to mark this PR as a draft for now and build it up as I run through Prettier formatting.  That way, it'll be one larger PR to remove all the files I catch, rather than a bunch of smaller PRs!

queengooborg:
Alright, I think I've found all I'm going to find for the time being, so I'm marking this as ready for review once again!

yin1999 (approved):
Thanks. Let's remove them

PR #14297, opened 2023-07-16
zh-ch: Format /glossary using Prettier

queengooborg:
This PR formats the `/glossary` folder within the Simplified Chinese locale using Prettier.


yin1999 (approved):
Thanks

PR #14319, opened 2023-07-16
es: Format /glossary using Prettier

queengooborg:
This PR formats the `/glossary` folder in the Spanish locale using Prettier.


files/es/glossary/base64/index.md
```diff
@@ -250,34 +290,33 @@ function strToUTF8Arr (sDOMStr) {
...
       aBytes[nIdx++] = 248 + (nChr >>> 24);
-      aBytes[nIdx++] = 128 + (nChr >>> 18 & 63);
-      aBytes[nIdx++] = 128 + (nChr >>> 12 & 63);
-      aBytes[nIdx++] = 128 + (nChr >>> 6 & 63);
+      aBytes[nIdx++] = 128 + ((nChr >>> 18) & 63);
+      aBytes[nIdx++] = 128 + ((nChr >>> 12) & 63);
+      aBytes[nIdx++] = 128 + ((nChr >>> 6) & 63);
       aBytes[nIdx++] = 128 + (nChr & 63);
-    } else /* if (nChr <= 0x7fffffff) */ {
+    } /* if (nChr <= 0x7fffffff) */ else {
```
Graywolf9, line 308: Not sure about this, I think the intention is read it like an  `else if`

```suggestion
    } else /* if (nChr <= 0x7fffffff) */ {
```

files/es/glossary/mixin/index.md
```diff
@@ -13,7 +13,7 @@ width: 100% ;
 
 }
 
-La forma más común de llamar a un mixin y aplicar esas propiedades y métodos es el _include_ (ejemplo @include nombre_del_mixin ). Al utilizar el _include_ para llamar las propiedades y métodos incluidos en el mixin, son entonces parte de la nueva interfaz.
+La forma más común de llamar a un mixin y aplicar esas propiedades y métodos es el _include_ (ejemplo @include nombre*del_mixin ). Al utilizar el \_include* para llamar las propiedades y métodos incluidos en el mixin, son entonces parte de la nueva interfaz.
```
Graywolf9, line 16: Seems that it missunderstood the unserscore

```suggestion
La forma más común de llamar a un mixin y aplicar esas propiedades y métodos es el _include_ (ejemplo `@include nombre_del_mixin`). Al utilizar el _include_ para llamar las propiedades y métodos incluidos en el mixin, son entonces parte de la nueva interfaz.
```

Graywolf9 (approved):
Thank you! There are two comments

files/es/glossary/mixin/index.md
queengooborg: Applying this suggestion, but we should probably delete this file as it is _very_ out-of-sync with upstream.  WDYT?

files/es/glossary/base64/index.md
queengooborg: This file also looks like it's quite out-of-sync with upstream, but I've made a tweak so we can get this PR landed!

Graywolf9:
Thank you so much! @queengooborg, unfortunately there is still having more outdated files than I wish, we rather keep due to it is easier that someone picks it to update instead of create from zero. I think I'm going to do a meta issue with oldest files to update

Thank yo so much again!

queengooborg:
Certainly, I know that performing updates on the files doesn't help with tracking more stale files if we don't have the `l10n.sourceCommit` frontmatter key -- I do, however, have some ideas on how we can tackle localization staleness across all locales, and will be talking with Ruth in the near future about it!

PR #14322, opened 2023-07-16
fr: Format /glossary using Prettier

queengooborg:
This PR formats the `/glossary` folder of the French locale using Prettier.


files/fr/glossary/falsy/index.md
```diff
@@ -13,17 +13,44 @@ Les valeurs fausses (**falsy**) sont des valeurs évaluées comme fausses quand
...
+  // Not reachable
+}
+
+if ("") {
+  // Not reachable
+}
+
+if (document.all) {
+  // Not reachable
+}
```
SphinxKnight, line 50:
```suggestion
if (false) {
  // Le code ici ne sera pas exécuté
}

if (null) {
  // Le code ici ne sera pas exécuté
}

if (undefined) {
  // Le code ici ne sera pas exécuté
}

if (0) {
  // Le code ici ne sera pas exécuté
}

if (-0) {
  // Le code ici ne sera pas exécuté
}

if (0n) {
  // Le code ici ne sera pas exécuté
}

if (NaN) {
  // Le code ici ne sera pas exécuté
}

if ("") {
  // Le code ici ne sera pas exécuté
}

if (document.all) {
  // Le code ici ne sera pas exécuté
}
```

files/fr/glossary/cross_axis/index.md
```diff
@@ -28,5 +28,5 @@ L'alignement des articles sur l'axe transversal est réalisé avec la propriét
 ### En lire plus
 
 - Guide Flexbox CSS : _[Les concepts de base pour flexbox](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Concepts_de_base_flexbox)_
-- Guide Flexbox CSS : _[Aligner des éléments dans un conteneur flexible](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Aligner_des_%C3%A9l%C3%A9ments_dans_un_conteneur_flexible)_
-- Guide Flexbox CSS : _[Maîtriser l'enveloppe des éléments flexibles](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Ma%C3%AEtriser_passage_%C3%A0_la_ligne_des_%C3%A9l%C3%A9ments_flexibles)_
+- Guide Flexbox CSS : _[Aligner des éléments dans un conteneur flexible](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Aligner_des_%C3%A9l%C3%A9ments*dans_un_conteneur_flexible)*
```
SphinxKnight, line 31: @queengooborg looks like the escapes are broken. The URLs should use the English slug now anyway. I wonder what would be the best solution: 
- Easiest: do nothing for such existing links in this current PR
- Longest/time consuming (as semi-manual action is needed/this could be scripted): change the links in another PR to use the English URL

SphinxKnight (changes requested):
There's an issue with the links containing escapes. I'm opened to debating. I'll be back at it and check the whole repo (or at least `fr`) with some regexps to see if we can batch change them.

files/fr/glossary/cross_axis/index.md
queengooborg: Ah yikes, gotta love Prettier bugs...

While I would say that using the existing URLs would be good, as they redirect to the new slugs, the escapes are a significant issue.  Luckily, it seems that `yarn tool fix-flaws` will automatically fix these links, and I plan to run that on this repository in the near future!

It seems like the HTTP encoding (%C3) are the issue, from what I can see -- for now, what if we just decoded the characters?

queengooborg: (I've actually just confirmed that `fix-flaws` will do the trick!)

SphinxKnight: yeah, decoding seems to work as well since the browser takes care of re-encoding. This should be easier as well.

image

queengooborg: Decoding complete!

SphinxKnight (approved):
let's get this merged. :) Thanks once again @queengooborg

PR #14350, opened 2023-07-18
Introduce frontmatter linter from upstream content

### LINKED ISSUE #7412: Remove duplicated frontmatter keys from localized content
As pointed out by @yin1999, frontmatter metadata (ex. tags, title) is [combined with the metadata from the original en-US docs](https://github.com/mdn/yari/blob/5d8d88976164b6f653c3f2de688f3cd76f8537bb/kumascript/index.js#L56-L69).  As such, keys such as tags, browser-compat, and spec-urls is redundant to copy to the localized content.  These duplicated keys should be removed from the localized docs for easier maintenance overall.  We should only keep: `title`, `short_title`, `slug`, ~~`original_slug`~~ and `l10n.*`.

**Edit; July 18, 2023**
A [configuration file](https://github.com/mdn/translated-content/blob/main/front-matter-config.json) is now present in this repository, which will let us utilize the frontmatter linter in upstream content!  To use the linter to fix the frontmatter, perform the following:

```sh
cd /path/to/mdn/content
yarn fix:fm --configFile /path/to/mdn/translated-content/front-matter-config.json --files /path/to/mdn/translated-content/files/<locale>
```

queengooborg:
This PR introduces the same frontmatter linter from the upstream content, tailoring the schema to match the needs of translated content.  This enforces and resolves #7412.

Note: this simply copies the scripts from upstream, which is probably not the best method.  I will be following up to determine how to best use the upstream script directly.

### DIFF
--- a/front-matter-config.json
+++ b/front-matter-config.json
@@ -1,0 +1,43 @@
+{
+  "lineWidth": -1,
+  "schema": {
+    "title": "Front matter schema",
+    "type": "object",
+    "additionalProperties": false,
+    "required": ["title", "slug"],
+    "properties": {
+      "title": {
+        "title": "Title",
+        "description": "Rendered page title and for SEO",
+        "type": "string",
+        "maxLength": 120
+      },
+      "short-title": {
+        "title": "Short Title",
+        "description": "To be used in sidebars",
+        "type": "string",
+        "maxLength": 60
+      },
+      "slug": {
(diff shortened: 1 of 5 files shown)


queengooborg:
I've decided that copying the scripts is not the right call here.  Instead, I will simply recommend calling the script in `mdn/content` directly, using the command line arguments available.

PR #13663, opened 2023-06-10
fr: Removes duplicated frontmatter keys (learn folder)

Lou8is:
### Description

Removes duplicated frontmatter keys (see #7412) from the **files/fr/learn** sub-folder.

Only `title`, `short_title`, `slug` and` l10n.*` are left, essentially removed `tags`, `original_slug`, `browser-compat`, `file-type`.

### Motivation

Help keeping the FR translation nice and up-to-date :)

### Related issues and pull requests

Relates to #7412 
Relates to 6 other PR for the other subfolders (#13629, #13630, #13631, #13632, #13633, #13664)

PR #14380, opened 2023-07-18
fr: Format /learn using Prettier

queengooborg:
This PR formats the `/learn` folder for the French locale using Prettier.


queengooborg:
I just checked to see if there would be any merge conflicts between these two PRs, and I am happy to say that there are none!  But regardless, if you'd like to hold off reviewing this PR for now, I'll happily await review at a later date!

(P.S. with the frontmatter keys, I recently added a configuration file for the localized content's frontmatter, so we can start using the frontmatter linter in upstream content as well!)

SphinxKnight:
> I just checked to see if there would be any merge conflicts between these two PRs, and I am happy to say that there are none! But regardless, if you'd like to hold off reviewing this PR for now, I'll happily await review at a later date!

Thanks a lot, I was just thinking out loud, remembering the markdown conversion time when conflicts were unavoidable ^ ^ I should have known better. Thanks for checking :)

queengooborg:
Of course!  I'll say that merge conflicts will probably occur regardless, but it shouldn't be anywhere near as big as the Markdown conversion (since we aren't renaming and completely changing files from the ground up).  I'm thinking that we'll end up with small merge conflicts that should be pretty easy to resolve if the need arises, and by taking it one folder/small batch at a time, it should stay at a minimum!

SphinxKnight (approved):
Scrolled through the changes and didn't notice anything, let's merge :) 

The EmbedLiveSample error in the test is also in https://github.com/mdn/translated-content/pull/13663 and I think we can bypass it for this one as this PR doesn't touch any EmbedLiveSample call

PR #14390, opened 2023-07-18
ja: Format /learn using Prettier

queengooborg:
This PR formats the `/learn` folder of the Japanese locale using Prettier.


mfuji09:
@queengooborg Can you resolve the CI's error?

queengooborg:
Hey @mfuji09!  I responded to a similar request in https://github.com/mdn/translated-content/pull/14263#issuecomment-1637757375:

> Thanks for asking!  While we do not modify the macros [in this PR], we do modify the files that contain them.  Since the CI test runs on all the modified files, we are notified that there are issues in the files even if the PR itself is not the source of said issues.
> 
> While we should address these issues at some point, I'd say that they are safe to ignore in this PR!

I do plan to resolve the CI failure in another PR (unless someone beats me to it)!  For now, though, I'm focusing on the formatting, particularly as it will greatly help with fixing up other issues like broken macros and issues like what the CI reports!

mfuji09 (approved):
Okay.

PR #14393, opened 2023-07-18
ko: Remove untranslated files

queengooborg:
This PR removes numerous files that are mostly or entirely untranslated from the Korean locae.


hochan222 (approved):

hochan222:
memo: Some precious paragraphs will be noted separately.

PR #14435, opened 2023-07-19
fr: Fix links broken by a Prettier bug

queengooborg:
This PR fixes links throughout the French locale that were broken by a nasty little Prettier bug.  This also adds ignores for those files, until the bug is resolved.

### DIFF
--- a/files/fr/glossary/cross_axis/index.md
+++ b/files/fr/glossary/cross_axis/index.md
@@ -28,5 +28,5 @@ L'alignement des articles sur l'axe transversal est réalisé avec la propriét�
 ### En lire plus
 
 - Guide Flexbox CSS : _[Les concepts de base pour flexbox](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Concepts_de_base_flexbox)_
-- Guide Flexbox CSS : *[Aligner des éléments dans un conteneur flexible](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Aligner_des*éléments*dans_un_conteneur_flexible)*
-- Guide Flexbox CSS : *[Maîtriser l'enveloppe des éléments flexibles](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Maîtriser_passage*à*la_ligne_des*éléments*flexibles)*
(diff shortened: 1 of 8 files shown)


SphinxKnight (approved):
Thank you @queengooborg

files/fr/glossary/cross_axis/index.md
```diff
@@ -28,5 +28,5 @@ L'alignement des articles sur l'axe transversal est réalisé avec la propriét
 ### En lire plus
 
 - Guide Flexbox CSS : _[Les concepts de base pour flexbox](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Concepts_de_base_flexbox)_
-- Guide Flexbox CSS : *[Aligner des éléments dans un conteneur flexible](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Aligner_des*éléments*dans_un_conteneur_flexible)*
-- Guide Flexbox CSS : *[Maîtriser l'enveloppe des éléments flexibles](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Maîtriser_passage*à*la_ligne_des*éléments*flexibles)*
+- Guide Flexbox CSS : _[Aligner des éléments dans un conteneur flexible](/fr/docs/Web/CSS/Disposition_flexbox_CSS/Aligner_des_éléments_dans_un_conteneur_flexible)_
```
yin1999, line 31: Hi @queengooborg, maybe we could fix those broken links to enable prettier format on those files again, as almost all slugs are in 'pure English' now (The link here should be due to historical reasons, we have localized the slugs, we translated slug in the past)

queengooborg: Hey @yin1999 -- unfortunately even if we updated the slugs to point to the new slug (which I plan to do!), the problem with Prettier will still be present.  It's an issue with Prettier's Markdown parser, to where it's thinking that the underscores inside of the links are italics formatting rather than part of a slug.  We have a similar issue upstream with underscores and asterisks inside of inline code:

https://github.com/mdn/content/blob/0c163056cfe83fba519b757f15d2e20f83eddaff/.prettierignore#L8-L10

PR #14446, opened 2023-07-20
add try it to scale

fyzhu:


yin1999 (approved):
Thanks. I've removed the unused image from the folder

PR #14462, opened 2023-07-20
es: Format /web/html using Prettier (part 1)

queengooborg:
This PR formats the `/web/html` folder of the Spanish locale using Prettier.  This is the first part.


Graywolf9 (approved):
Thank you!

PR #14467, opened 2023-07-21
fr: Format /learn/css using Prettier

queengooborg:
This PR formats the `/learn/css` folder of the French locale using Prettier.


SphinxKnight (approved):
let's go :) Thanks @queengooborg

PR #14470, opened 2023-07-21
Remove broken Compat section (not in en-us doc)

Hugothms:
Removed empty 'Compatibility' section

### Description

The Compatibility section is not present on the English version of the page and doesn't work on the French one so I removed it
The section on the French one before changes:
Screenshot 2023-07-21 at 12 01 36


SphinxKnight (approved):
Merci @Hugothms :)

PR #14480, opened 2023-07-22
Introduce script to check the locale of documents

queengooborg:
This PR introduces a script that checks the languages found within documents and reports them in Markdown or JSON format.  This will be crucial to both [cleanup the docs](https://github.com/mdn/mdn/issues/428), and to help catch similar issues that may arise in the future (not that I expect there to be any; cheers to the amazing localization teams).

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "type": "git",
     "url": "git+https://github.com/mdn/translated-content.git"
   },
+  "type": "module",
   "scripts": {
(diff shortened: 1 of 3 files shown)


queengooborg:
For those curious, this is the output of the script for each locale: https://drive.google.com/drive/folders/1HWp_sUNJhYmgwJDMZpQ4toXMLY17fx_X?usp=drive_link

PR #14525, opened 2023-07-23
fr: Format /learn/server-side using Prettier

queengooborg:
This PR formats the `/learn/server-side` folder of the French locale using Prettier.  This also fixes a lot of instances where the code language was incorrectly set.


SphinxKnight (approved):
Thanks @queengooborg, esp. for switching the `html` block to `django` where necessary :pray:

PR #14539, opened 2023-07-23
ru: Format /learn/html using Prettier

queengooborg:
This PR formats the `/learn/html` folder of the Russian locale using Prettier.


.prettierignore
```diff
@@ -21,6 +21,7 @@ build/
 /files/fr/glossary/main_axis/index.md
 /files/fr/learn/server-side/django/forms/index.md
 /files/pt-br/learn/server-side/django/forms/index.md
+/files/ru/learn/html/introduction_to_html/the_head_metadata_in_html/index.md
```
sashasushko, line 24: @queengooborg I see that you added this file to Prettier ignore, but also I see changes in it.

I'm confused a little bit.

queengooborg: I add files to the `.prettierignore` if Prettier is bugging out and making a change as described in https://github.com/prettier/prettier/issues/11828.  I still run Prettier on that file and commit the _good_ changes, though; that way once the bug is fixed, we can just remove the ignore and there should be no other changes to make!

queengooborg: (P.S. I know that we can add `<!-- prettier-ignore-start -->` around the affected lines instead of ignoring the whole file -- but to be honest, I want to formulate a list of affected files that can be shown to the Prettier team to put a little pressure on them to fix the nasty bug... :D)

sashasushko: Cool. Thanks for the quick response 🙏

sashasushko (approved):

PR #14540, opened 2023-07-23
Fix #14405

### LINKED ISSUE #14405: fr: Web/HTTP/Content_negotiation partially translated
The French localization of `Web/HTTP/Content_negotiation` is partially translated.  It may need to be resynchronized with upstream.

SphinxKnight:
Fixes #14405
Rework the whole page, include images' translation as well.


SphinxKnight:
I don't know where we should store original copies of diagrams (versions that can be edited instead of pngs,jpgs), so here we are: https://gist.github.com/SphinxKnight/5e78b8133db882fc34c479fae893b4ea

cw118:
> I don't know where we should store original copies of diagrams (versions that can be edited instead of pngs,jpgs), so here we are: https://gist.github.com/SphinxKnight/5e78b8133db882fc34c479fae893b4ea

That is great 👍 Maybe we can insert an HTML comment above/below the image with the link of the gist corresponding to it, or create a document in this repository with links pointing to the gists all in one place?

cw118 (approved):
LGTM 🙏 I will merge this now, though I am open to (more) discussion on the placement/location of localized image files ^^

PR #14556, opened 2023-07-23
[ru] add `Web/JavaScript/Reference/Global_Objects/DataView/buffer` translation

42proger:
### Description
Translated: DataView.prototype.buffer article translated into Russian language.

### Motivation
This article was not originally translated to Russian language.


leon-win (approved):
I removed unnecessary front matter fields. Usually only `title`, `short-title` and `slug` are required, and the rest are inherited from the `en-US` version of the document.

Also improved some wording))

Thank you for PR, @42proger!

PR #14574, opened 2023-07-24
ru: add a missing word to the translation

Handsome1337:
Adding a missing word.

### Description

A mistake was made in the translation - a word was missing.

### Motivation

I noticed this mistake while reading the article. I wanted to make the MDN project better because it is useful for a huge number of people.


Saionaro:
@Handsome1337 привет! Спасибо за вклад в MDN!

PR #14582, opened 2023-07-25
es: Format /web/javascript using Prettier (part 2)

queengooborg:
This PR formats the `/web/javascript` of the Spanish locale using Prettier.  This is the second part.


Graywolf9:
👍

PR #14614, opened 2023-07-26
fr: Format /web/html using Prettier (part 1)

queengooborg:
This PR formats the `/web/html` folder of the French locale using Prettier.  This is the first part.


SphinxKnight (approved):
easy review :)

PR #14616, opened 2023-07-26
fr: Format /web/html using Prettier (part 3)

queengooborg:
This PR formats the `/web/html` folder of the French locale using Prettier.  This is the third part.


SphinxKnight (approved):
A minor thought about Prettier putting newlines before some `>` but nothing blocking

PR #14625, opened 2023-07-27
fr: Format /web/css using Prettier (part 2)

queengooborg:
This PR formats the `/web/css` folder of the French locale using Prettier.  This is the second part.


cw118 (approved):
💯

PR #14626, opened 2023-07-27
fix typo in pointer event

520ph:
fix 乘坐 > 称作


yin1999 (approved):
Thanks. Welcome aboard.

PR #14660, opened 2023-07-27
zh-cn: Format /web/css using Prettier (part 2)

queengooborg:
This PR formats the `/web/css` folder of the Simplified Chinese locale using Prettier.  This is the second part.


yin1999 (approved):
Thanks @queengooborg :)

PR #14670, opened 2023-07-28
zh-cn: Format /web/http using Prettier

queengooborg:
This PR formats the `/web/http` folder of the Simplified Chinese locale using Prettier.


yin1999 (approved):
Thanks

PR #14672, opened 2023-07-28
zh-cn: Format /web/svg using Prettier

queengooborg:
This PR formats the `/web/svg` folder of the Simplified Chinese locale using Prettier.


yin1999 (approved):
Thanks

PR #14673, opened 2023-07-28
zh-cn: Format /web/api using Prettier (part 1)

queengooborg:
This PR formats the `/web/api` folder of the Simplified Chinese locale using Prettier.  This is the first part.


jasonren0403 (approved):
Thank you @queengooborg , I've allowed myself to fix some occurences where 'nolint' should be added.

files/zh-cn/web/api/canvas_api/tutorial/compositing/index.md
jasonren0403, line 1: It seems that we should not put anything in 5th param of EmbedLiveSample, or build would broke up. 

I'm not sure if we should do a scan to batch remove 5th param in en-US content.

image

yin1999: There does have a issue: mdn/content#25005

PR #14676, opened 2023-07-28
zh-cn: Format /web/api using Prettier (part 3)

queengooborg:
This PR formats the `/web/api` folder of the Simplified Chinese locale using Prettier.  This is the third part.


yin1999 (approved):
Thanks @queengooborg. I've pushed e7fbb25b74fb36ee2f5728b147622484971d7714 to prevent the format of syntax code blocks :)

It looks good to me now, I'm going to merge this.

PR #14680, opened 2023-07-28
zh-cn: Format /web/api using Prettier (part 6)

queengooborg:
This PR formats the `/web/api` folder of the Simplified Chinese locale using Prettier.  This is the sixth locale.


yin1999 (approved):
Thanks

PR #14689, opened 2023-07-28
pt-br: Format /web/api using Prettier (part 2)

queengooborg:
This PR formats the `/web/api` folder of the Brazilian Portuguese locale using Prettier.  This is the second part.

Additionally, this replaces the old spec tables and compat table calls, along with other little tweaks here and there.


queengooborg:

nathipg (approved):
Thank you!

PR #14690, opened 2023-07-28
pt-br: Format /web/api using Prettier (part 3)

queengooborg:
This PR formats the `/web/api` folder of the Simplified Chinese locale using Prettier.  This is the third part.

Note: this PR also performs replacement of the old specification tables and compat table calls.


nathipg (approved):
Thank you!

PR #14691, opened 2023-07-28
pt-br: Format /web/api using Prettier (part 4)

queengooborg:
This PR formats the `/web/api` folder of the Brazilian Portuguese locale using Prettier.  This is the fourth part.

This PR also performs replacement of old spec tables and compat table calls.


josielrocha (approved):
LGTM ;)

PR #14710, opened 2023-07-29
ko: Format /web/api using Pretier (part 5)

queengooborg:
This PR formats the `/web/api` folder of the Korean locale using Prettier.  This is the fifth part.


hochan222 (approved):
LGTM :)

PR #14757, opened 2023-07-30
ja: Format /web/html using Prettier (part 2)

queengooborg:
This PR formats the `/web/html` folder of the Japanese locale using Prettier.


files/ja/web/html/element/input/color/index.md
```diff
@@ -117,15 +117,17 @@ colorWell.select();
...
-  <code>&lt;input type="color"&gt;</code> コントロールを使用する例です。
-</p>
+<p><code>&lt;input type="color"&gt;</code> コントロールを使用する例です。</p>
 
 <label for="colorWell">色:</label>
 <input type="color" value="#ff0000" id="colorWell" />
 
 <p>
-  カラーピッカーを動かしたときに段落の色が変わるのを見てください。カラーピッカーに変更を加えると、最初の段落の色が、プレビューとして変化します（<code>input</code> イベントを使用します）。カラー ピッカーを閉じると、<code>change</code> イベントが実行され、全ての段落が選択された色に変化することが分かります。
+  カラーピッカーを動かしたときに段落の色が変わるのを見てください。カラーピッカーに変更を加えると、最初の段落の色が、プレビューとして変化します（<code
```
mfuji09, line 126: This paragraph shouldn't be broken.

files/ja/web/html/element/input/file/index.md
```diff
@@ -323,9 +325,9 @@ function updateImageDisplay() {
       const listItem = document.createElement("li");
       const para = document.createElement("p");
       if (validFileType(file)) {
-        para.textContent = `ファイル名: ${file.name}, ファイルサイズ: ${returnFileSize(
-          file.size
-        )}.`;
+        para.textContent = `ファイル名: ${
```
mfuji09, line 328: This line shouldn't be broken, too.

files/ja/web/html/element/video/index.md
```diff
@@ -261,7 +261,10 @@ l10n:
   <source src="myVideo.webm" type="video/webm" />
   <source src="myVideo.mp4" type="video/mp4" />
   <p>
-    このブラウザーは HTML の動画に対応していません。代わりに<a href="myVideo.mp4">動画へのリンク</a>があります。
+    このブラウザーは HTML の動画に対応していません。代わりに<a
```
mfuji09, line 264: This paragraph shouldn't be broken.

files/ja/web/html/element/video/index.md
```diff
@@ -351,7 +354,10 @@ AddType video/webm .webm
   src="https://archive.org/download/BigBuckBunny_124/Content/big_buck_bunny_720p_surround.mp4"
   poster="https://peach.blender.org/wp-content/uploads/title_anouncement.jpg?x11217"
   width="620">
-  恐れ入りますが、このブラウザーは埋め込み動画に対応していません。しかし、<a href="https://archive.org/details/BigBuckBunny_124">ダウンロード</a>してお好きな動画プレイヤーで見ることはできます。
+  恐れ入りますが、このブラウザーは埋め込み動画に対応していません。しかし、<a
```
mfuji09, line 357: This paragraph shouldn't be broken, too.

files/ja/web/html/element/video/index.md
```diff
@@ -385,7 +391,10 @@ AddType video/webm .webm
     src="https://archive.org/download/ElephantsDream/ed_1024_512kb.mp4"
     type="video/mp4" />
 
-  恐れ入りますが、このブラウザーは埋め込み動画に対応していません。しかし、<a href="https://archive.org/download/ElephantsDream/ed_1024_512kb.mp4">MP4 をダウンロード</a>してお好きな動画プレイヤーで見ることはできます。
+  恐れ入りますが、このブラウザーは埋め込み動画に対応していません。しかし、<a
```
mfuji09, line 394: This paragraph shouldn't be broken, too.

mfuji09 (changes requested):
Please review comments.

files/ja/web/html/element/input/file/index.md
queengooborg: Prettier is putting the line break in an odd spot...  Setting `printWidth: 100` in the Prettier configuration seems to bypass this and retain formatting that's similar to where the line break is in upstream.  Maybe Prettier is treating Kanji as double-width characters or something like that?

(In this particular case, though, I don't think that it matters much; the break does not actually affect the output itself.)

files/ja/web/html/element/input/color/index.md
queengooborg: Actually, this would be an alright place to put a line break.  Note that the line break happens just before the closing angle bracket, so the rendered page looks the same.

mfuji09: I am saying that the readability of **source code* includes Japanese text becames awful. 

For example, what do you think if the a code written as:
Line breaking of so
urce code doesn't ef
fect of output. And th
e readability of so
urce code is not a b
ig matter.

PR #14766, opened 2023-07-30
ko: Format various docs using Prettier

queengooborg:
This PR uses Prettier to format a few docs in the Korean locale for folders that have already gotten a Prettier pass and no longer match Prettier formatting since.


hochan222 (approved):
LGTM :)

PR #14787, opened 2023-08-01
ja: Format /web/css using Prettier (part 3)

queengooborg:
This PR formats the `/web/css` folder of the Japanese locale using Prettier.  This is the third part.


files/ja/web/css/_colon_target/index.md
```diff
@@ -52,15 +52,14 @@ http://www.example.com/index.html#section2
...
-      このリンクは対象がないので、どこにも行きません。
-    </a>
+    <a href="#nowhere">このリンクは対象がないので、どこにも行きません。</a>
   </li>
 </ol>
 
 <h3>面白い記事</h3>
 <p id="p1">
-  URL フラグメントを使うと<i>この段落</i>をターゲットにすることができます。上記のリンクをクリックして試してみてください。
+  URL
```
mfuji09, line 61: @queengooborg 
Inserting line break in this place makes readability worse. "URL フラグメント" ("URL fragment") should be in a single line.

mfuji09 (dismissed):
Please confirm one comment.

files/ja/web/css/_colon_target/index.md
queengooborg: Unfortunately, we don't have much control over these sorts of details with Prettier -- because there's a space in between the two words, Prettier sees it as an opportunity to break the line, and thus, does.

queengooborg: If it is needed, we can suffix the code language with `-nolint` instead, so that Prettier won't format the section (ex. `html-nolint` instead of `html`).

mfuji09: So do you add it by yourself?
I think that the responsibility for keeping the integrity of a pull request should be borne by the person who created the pull request, as written in <https://developer.mozilla.org/en-US/docs/MDN/Community/Pull_requests>.

queengooborg: I wasn't sure if you wanted `-nolint` added or not, so I did not make changes to the pull request.  It would probably be better for someone else to tackle making those changes, though -- I am also responsible for many other projects across MDN, and the time I have available to work on translated content is significantly less than I had last month.

mfuji09: Then, why don't you ask or consult us (the `ja` community) before you do it?
If you discussed the community about the work **before** you do it, you would be able to be noticed side effects and problems, and the task which should be done would become clearer. The work would be lighter and sooner.
And the community might help the work.

We also have a lot of tasks (either in or out of the MDN), not only you.
I spend a lot of time for reviews for your mechanical editing, however in my heart, I want to spend the time to improve translations which has been old or to translate articles which hasn't been done instead.

queengooborg: > Then, why don't you ask or consult us (the ja community) before you do it?

Uh...all translation teams were pinged on the relevant MDN project weeks ago: https://github.com/mdn/mdn/issues/428

> I spend a lot of time for reviews for your mechanical editing, however in my heart, I want to spend the time to improve translations which has been old or to translate articles which hasn't been done instead.

This is exactly why MDN uses Prettier to format all pages.  Nobody wants to spend the time reviewing formatting, whether it's for a missing backslash or a table not having enough whitespace, when automatic tooling can perform this work for us instead.  Often times, automated tooling has also helped catch issues in the documents that have been/could be missed during human review -- and this was proven so many times when I had done the formatting on English content.

Upstream, we do understand that Prettier isn't the perfect tool.  None of us enjoy its formatting choices of XML-based languages (like HTML); it does some wonky formatting in English content too.  I've even found that it has some nasty bugs that break links if they're surrounded by formatting syntax.  But in the end...there is just nothing else like Prettier.

My point is: it's going to be a lot of work to perform and review these mechanical formatting changes.  But once we get over this mountain, it should improve things for everyone involved.

> We also have a lot of tasks (either in or out of the MDN), not only you.

I apologize, I didn't mean to imply that I'm the only busy one.  I'm very grateful for the time that you, and every localization team member, spends on MDN.  What I meant by my last statement is that I have gone over my allotted time budget for working on the localized content, and that I now must turn my primary focus back to `browser-compat-data`.

yin1999: As I haven't studied Japanese yet, if there is something wrong, please point it out:

I just read the documentation of prettier: https://prettier.io/blog/2023/07/05/3.0.0.html#stop-inserting-spaces-between-chinese-or-japanese-and-western-characters .

A better solution is (maybe), we can [remove the spaces between Japanese/Chinese and Latin letters](https://github.com/tats-u/prettier-plugin-md-nocjsp#why-this-plugin-is-needed) (referenced by the document of prettier) . In this way, we can avoid such problems.

But unfortunately, many fonts have trouble handling this (rendering text in the browser without small spaces between Japanese/Chinese and Latin letters). And common Markdown could not handle markdown styles with mixed Chinese and Latin letters perfectly. We (for `l10n-zh`) had to add spaces between Chinese and Latin letters.

I'm not sure if there is a similar situation in Japanese.Especially the second point I mentioned. If we don't have the second problem in Japanese, is it possible to consider removing these spaces (maybe we just need to deal with font issues) ?

mfuji09: @yin1999 
Thank you for your comment!
Unfortunately, Japanese rendering in most web browsers (or fonts) also have the same problem to Chinese, i.e. they don't insert small spaces between Chinese/Japanese characters and Latin characters. So we insert spaces between them instead.
Removing these space characters makes bad readability for the result of rendering. So we want to keep the spaces characters.
(Word processors (such as Micorsoft Word), DTPs and TeX insert such a spaces, regardless what font is used.)

I hope that Prettier stops to insert line breaks for Chinese and Japanese sentences.

queengooborg: Looking back at the thread, I think that there may be a slight confusion about these line breaks.  The line breaks are _only_ visible in the source, i.e. the browser still renders those lines as a single line!  (In other words, `URL\nフラグメント` is rendered as `URL フラグメント`.)

yin1999: Yes, consecutive whitespace characters are converted to single spaces in the rendered text. I believe mfuji09 was thinking that the reader might misunderstand what the source code is conveying.

The only way we can consider at present is to add `-nolint`.

queengooborg:
Self-merging as feedback has been addressed and this PR has been sitting around for weeks.

Regarding the open conversation, it would probably be good to discuss it in an issue or discussion.

PR #14794, opened 2023-08-01
ko: Format /web/css using Prettier (part 2)

queengooborg:
This PR formats the `/web/css` folder of the Korean locale using Prettier.  This is the second part.


queengooborg:

hochan222 (approved):
LGTM :)

PR #14796, opened 2023-08-01
pt-br: Format /web/css using Prettier (part 1)

queengooborg:
This PR formats the `/web/css` folder of the Brazilian Portuguese folder using Prettier.  This is the first part.


queengooborg:

nathipg (approved):
Thank you!

PR #14807, opened 2023-08-01
ja: Format /web/javascript using Prettier (part 2)

queengooborg:
This PR formats the `/web/javascript` folder of the Japanese locale using Prettier.  This is the second part.


queengooborg:
Self-merging as this has been sitting around for a while without review.

PR #14777, opened 2023-07-31
translate cryptographic hash function for zh-CN

fwqaaq:

PR #14838, opened 2023-08-02
Update cryptographic translation about zh-CN

fwqaaq:
https://github.com/mdn/translated-content/pull/14777#discussion_r1280239883

@yin1999 's translation is good, but I forgot to update my translation at that time.


yin1999 (approved):
Thanks @fwqaaq

PR #14842, opened 2023-08-02
fix broken background_tasks_api html and polish translation

yunsii:
### Description

Fix broken background_tasks_api html and polish translation with more fluent statement

### Motivation

Example broken

### Additional details

https://developer.mozilla.org/zh-CN/docs/Web/API/Background_Tasks_API#%E7%BB%93%E6%9E%9C


yin1999 (approved):
Thanks @yunsii. As the format of this document does not meet the requirements of the CI we are running now, I have updated the full text. Welcome aboard :tada:

PR #14845, opened 2023-08-02
zh-cn: Lint front matter keys (part 1)

queengooborg:
This PR uses the front matter linter to lint the front matter keys of the Simplified Chinese locale.  This is the first part.


yin1999:
The sync content tool will still add original_slug to fromt-matter. I have no idea about removing this key. But we should advise collaborators to [do some chores according to the modification in content repo before merging content sync PRs](https://github.com/mdn/translated-content/blob/main/docs/orphaned_and_conflicting.md).

yin1999 (approved):

queengooborg:
Ah, right, I had forgotten to avoid performing formatting on conflicting/orphaned documents...  🤦‍♀️   (On the upside, maybe this linting will force a resolution before merging, so we never end up with those two folders...!)

PR #14846, opened 2023-08-02
zh-cn: Lint front matter keys (part 2)

queengooborg:
This PR uses the front matter linter to lint the front matter of the Simplified Chinese locale.  This is the second part.


yin1999 (approved):
Thanks 🙏

PR #14872, opened 2023-08-03
fr: add precision about selecting numeric IDs with QuerySelector

AppyGG:
### Description

Add a precision about selecting item with id starting by a number

### Motivation

I encountered this issue a few times and only really found a [StackOverflow](https://stackoverflow.com/questions/20306204/using-queryselector-with-ids-that-are-numbers) post about this without proper documentation

### Additional details

- [StackOverflow discussion about this problem](https://stackoverflow.com/questions/20306204/using-queryselector-with-ids-that-are-numbers)
- [W3Docs about HTML5 IDs](https://www.w3docs.com/snippets/html/which-are-the-valid-values-for-the-id-attribute.html#:~:text=In%20HTML5%2C%20the%20id%20attribute,id%20attribute%20is%20case%20sensitive.)

### DIFF
--- a/files/fr/web/api/document/queryselector/index.md
+++ b/files/fr/web/api/document/queryselector/index.md
@@ -40,6 +40,17 @@ Si le sélecteur correspond à un ID et que cet ID est utilisé de façon erron�
 
 Les [pseudo-éléments](/fr/docs/Web/CSS/Pseudo-elements) CSS ne retourneront jamais aucun élément, comme spécifié dans l'[API des sélecteurs](http://www.w3.org/TR/selectors-api/#grammar) (en).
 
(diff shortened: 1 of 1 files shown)


cw118:
Waiting for a response in mdn/content#28752

cw118:
Hi @AppyGG, apologies for the delay here and thank you for the pull request.

Unfortunately, based on [the issue progress and comments on mdn/content](https://github.com/mdn/content/issues/28752#issuecomment-1707035191), mentioning the precision about selecting numeric `id`s with `querySelector` will require updating multiple pages. And since French content is translated from its en-US counterpart, I would like to wait for the en-US updates, so I will close this one for now.

Don't hesitate if you have any questions :)

PR #14880, opened 2023-08-03
ru: Lint front matter keys

queengooborg:
This PR lints and fixes the front matter keys throughout the Russian locale using the front matter linter, followed by manual fixes as needed.

Note: this will be self-merged as it involves infrastructure and linting that should not require locale-specific review.


sashasushko:
@queengooborg thanks. I noticed that you removed a few front matter keys, especially the `tags` key. Does that mean we should not use it in translation?

queengooborg:
That's correct -- translated content should only use four front matter keys: `title`, `short-title` (optional), `slug` and `l10n.sourceCommit`.  All other tags are automatically copied from upstream during rendering!

(Side note, tags have actually been removed from MDN altogether -- they're not helpful for SEO, and the tags that affect page rendering have been migrated to more appropriate front matter keys!)

PR #14907, opened 2023-08-04
fix(lint-staged): markdownlint and prettier can race on md files

yin1999:
### Description

Synchronize script changes from upstream.

upstream changes: mdn/content#28306


LeoMcA (approved):
thanks for porting this over!

PR #14909, opened 2023-08-04
docs(ja): WebAssembly/JavaScript_interface/Global 以下を更新

bsmth:
Test PR to reproduce build errs seen in https://github.com/mdn/translated-content/pull/14765

### DIFF
--- a/files/ja/webassembly/javascript_interface/global/global/index.md
+++ b/files/ja/webassembly/javascript_interface/global/global/index.md
@@ -1,47 +1,60 @@
 ---
 title: WebAssembly.Global() コンストラクター
 slug: WebAssembly/JavaScript_interface/Global/Global
+original_slug: Web/JavaScript/Reference/Global_Objects/WebAssembly/Global/Global
+l10n:
+  sourceCommit: acfe8c9f1f4145f77653a2bc64a9744b001358dc
 ---
 
 {{WebAssemblySidebar}}
 
-**`WebAssembly.Global()`** コンストラクターは、グローバル変数のインスタンスを表す新しい `Global` オブジェクトを表し、これは JavaScript からアクセス可能で、 1 つ以上の {{jsxref("WebAssembly.Module")}} インスタンスの間でインポート/エクスポート可能です。これにより、複数のモジュールを動的リンクすることができます。
+**`WebAssembly.Global()`** コンストラクターは、グローバル変数のインスタンスを表す新しい `Global` オブジェクトを表します。これは JavaScript からアクセス可能で、1 つ以上の [`WebAssembly.Module`](/ja/docs/WebAssembly/JavaScript_interface/Module) インスタンスの間でインポート/エクスポート可能です。これにより、複数のモジュールを動的リンクすることができます。
 
 ## 構文
 
-```js
-new WebAssembly.Global(descriptor, value);
+```js-nolint
+new WebAssembly.Global(descriptor, value)
 ```
 
 ### 引数
 
-- _descriptor_
+- `descriptor`
 
-  - : `GlobalDescriptor` 辞書オブジェクトで、 2 つのプロパティを持っています。
+  - : オブジェクトで、 2 つのプロパティを持ちます。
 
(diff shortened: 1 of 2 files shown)


bsmth:
CI fails here:

<details>

```
  # Use the GitHub API to get the list of changed files
  # documenation: https://docs.github.com/rest/commits/commits#compare-two-commits
  DIFF_DOCUMENTS=$(gh api repos/{owner}/{repo}/compare/ba5878ee9ac3f9d3374d82186771a07c1abbf364...5843766c13a3eda14723df4b4ac226aa8548828f \
    --jq '.files | .[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) | .filename')
  
  # filter out files that are not markdown files
  GIT_DIFF_CONTENT=$(echo "${DIFF_DOCUMENTS}" | egrep -i "^files/.*\.md$" | xargs)
  echo "GIT_DIFF_CONTENT=${GIT_DIFF_CONTENT}" >> $GITHUB_ENV
  
  # filter out files that are not attachments
  # note that we should get the absolute path of the changed attachments
  GIT_DIFF_FILES=$(echo "${DIFF_DOCUMENTS}" | egrep -i "^files/.*\.(png|jpeg|jpg|gif|svg|webp)$" | xargs readlink -e | xargs)
  echo "GIT_DIFF_FILES=${GIT_DIFF_FILES}" >> $GITHUB_ENV
  shell: /usr/bin/bash -e {0}
  env:
    BASE_SHA: ba5878ee9ac3f9d3374d82186771a07c1abbf364
    HEAD_SHA: 5843766c13a3eda14723df4b4ac226aa8548828f
    GITHUB_TOKEN: ***
transform: short source buffer
```

</details>

Passing `gh api` call:

```bash
gh api repos/mdn/translated-content/compare/ba5878ee9ac3f9d3374d82186771a07c1abbf364...5843766c13a3eda14723df4b4ac226aa8548828f
```

Failing:

```bash
gh api repos/mdn/translated-content/compare/ba5878ee9ac3f9d3374d82186771a07c1abbf364...5843766c13a3eda14723df4b4ac226aa8548828f \
    --jq '.files | .[] | select(.status|IN("added", "modified", "renamed", "copied", "changed")) | .filename'
```

And `transform: short source buffer` is a Go err: https://go.googlesource.com/text/+/23ae387dee1f90d29a23c0e87ee0b46038fbed0e/transform/transform.go#23

mfuji09:
@bsmth 
Thank you very much, for identifying the reason of problem!

However, the unicode character reference has been introduced from the English version of that time (before the commit <https://github.com/mdn/content/commit/f28478220b51fdba9d7ca82cfd5dcde29b104881#diff-ab009fdab2d2d37768c947522209ddd397af896272bc13e84e7d61e0151f7cc8>).

This PR removes the character (corresponding to the commit above for the `content` repo). But, do you know how should I do to solve it? (should I force to merge?)

bsmth:
My pleasure. Any pull request that has a diff including this character `�` will fail in the same way in future. We might want to replace occurrences of this with the HTML entity: `&#65533;` so that it displays properly in the rendered docs but is parsed okay by the `gh api` tool. Open to opinions on that.

This PR removes it so other branches updating from main won't experience this problem, so I believe it's safe to merge. What do you think?

yin1999:
I would also like to have a deeper look into [GitHub cli](https://github.com/cli/cli). I think we have a chance to fix this so that future CIs will handle this special character correctly :)

bsmth:
> I would also like to have a deeper look into [GitHub cli](https://github.com/cli/cli). I think we have a chance to fix this so that future CIs will handle this special character correctly :)

True, we might want to report this as an issue [on the repo](https://github.com/cli/cli) that fetching patches containing this character with `gh api` is not handled well with current encoding used.

yin1999:
> ```shell
> repos/mdn/translated-content/commits/7144ca0458a33571ade432951827e61443d85ff5
> ```

I can now confirm that the issue is caused by a dependency of the GitHub cli:

[Repro code](https://github.com/mdn/translated-content/files/12319797/src.zip)

It is caused by the [sanitizer](https://github.com/cli/go-gh/blob/baae867137c67165ebd17e8ced4427fd9fcc44f5/pkg/asciisanitizer/sanitizer.go#L28-L97) used by the transformer.

image

I'll dig further into this issue :)

PR #14924, opened 2023-08-05
es: Remove unneeded Prettier ignore

queengooborg:
This PR removes an unneeded Prettier ignore for a folder in the Spanish locale which has been entirely formatted already.


Graywolf9:
👍

PR #14954, opened 2023-08-07
chore(CI): limit actions not to run on forks, cron rewrite

bsmth:
I believe we shouldn't be running these actions on forks, if I'm wrong, feel free to object and let me know.


LeoMcA (commented):
My concern is that these won't now run on PRs from forks, though the GitHub docs suggest they should:

> For pull requests from a forked repository to the base repository, GitHub sends the pull_request, issue_comment, pull_request_review_comment, pull_request_review, and pull_request_target events to the base repository. No pull request events occur on the forked repository.

\- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull-request-events-for-forked-repositories

Have we tested this? Because the docs also suggest that workflows won't run in forks:

> Workflows don't run in forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.

\- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories

bsmth:
I believe this is about the base repository that the pull request is opened against, see:

- PR: https://github.com/mdn/content/pull/28447
- Run: https://github.com/mdn/content/actions/runs/5797069215/job/15711890012?pr=28447
- Workflow: https://github.com/mdn/content/actions/runs/5797069215/workflow?pr=28447

bsmth:
Overlooked this part:

>The docs also suggest that workflows won't run in forks:
>> Workflows don't run in forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.

This wasn't always the case, so they run if forks are older than whenever this policy was rolled out. Mine is one of them (see https://github.com/bsmth/translated-content/pull/1).

LeoMcA (approved):
Cool, let's give it a go, worse case we can always revert if it doesn't work as expected.

PR #14972, opened 2023-08-08
delete untranslated content - glossary

hochan222:
https://github.com/mdn/translated-content/issues/14965 의 glossary 폴더 부분입니다.


yin1999:
> @yin1999, Hello! There seems to be an error in DIFF_DOCUMENTS. Can you check the current issue? https://github.com/mdn/translated-content/blob/main/.github/workflows/pr-check-lint_content.yml (cc. @queengooborg )
> 
> image

We could add a if condition just like [this](https://github.com/mdn/translated-content/blob/e46df3f02195162c35701c30d4bcaa0a7a4b129f/.github/workflows/pr-test.yml#L71). We also need to do this in mdn/content.

```yaml
if: env.DIFF_DOCUMENTS
```

PR #14973, opened 2023-08-08
Fix lint script when files are only deleted

queengooborg:
This PR fixes the lint workflow to only perform formatting when there are files to format.  This prevents a failure when documents have only been deleted.

Additionally, this removes a now outdated comment.


yin1999 (approved):
LGTM. As we have done this in [pr-test](https://github.com/mdn/translated-content/blob/e46df3f02195162c35701c30d4bcaa0a7a4b129f/.github/workflows/pr-test.yml#L71)

PR #14975, opened 2023-08-08
Fix Typo Error in Chinese Void Element Page

TakumiBC:
### Description
There is a typo error. It shall be "自闭合元素" not ”子闭合元素".

### Motivation
Because it is a typo error, it may make it confusing for the reader to understand.


yin1999 (approved):
Good catch @thomaswcy. Welcome aboard 🎉

thomaswcy:
Thank you for merging! I would love to help with the Chinese translation of MDN in the future.

PR #14989, opened 2023-08-08
es: Observations from frontmatter lintintg

Graywolf9:
### Description

`original_slug` are a needy frontmatter for fixing `orphaned` and `conflicting` files. I remember that it was an optional frontmatter but we love to keep at least in those folders

### DIFF
--- a/files/es/conflicting/learn/javascript/objects/classes_in_javascript/index.md
+++ b/files/es/conflicting/learn/javascript/objects/classes_in_javascript/index.md
@@ -1,6 +1,7 @@
 ---
 title: JavaScript orientado a objetos para principiantes
 slug: conflicting/Learn/JavaScript/Objects/Classes_in_JavaScript
+original_slug: Learn/JavaScript/Objects/Object-oriented_JS
 ---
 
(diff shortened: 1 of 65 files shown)


queengooborg (changes requested):
Re-adding `orignal_slug` sounds good to me if it is needed for `conflicting/` and `orphaned/`!  It looks like some other keys were added back, however (ex. `browser-compat`), may we remove those?

Graywolf9:
I remember added some of that in `conflicting` due to the `spec` and `compat` macros, but let me take a look

Graywolf9:
I already checked and in fact, all those files (10) have the `browser-compat` macro to not break `specs` and `compat` sections. If it's possible I would rather keep until I solve and move or delete that files.

queengooborg:
I gotta be honest, keeping those keys will cause an issue with front matter linting when it is introduced.  (The front matter linter doesn't have a file ignore feature, and it would be best not to have to add one in.)

Graywolf9:
keep `original_slug` in `conflicting`/`orphaned` does not have that problem?

queengooborg:
Not anymore; I just updated the front matter linting config to allow for `original_slug`!

Graywolf9:
Great! Thanks for all your time, I removed `browser-compat` frontmatter and the `specs`/`compat` sections to not show the errors

queengooborg (approved):
Thanks!

I imagine that we'll probably remove a lot of these files in the end,

Issue #3752, opened 2022-01-21
Issue with "filter": (the images in live samples are broken)


yusixian:
MDN URL: https://developer.mozilla.org/docs/Web/CSS/filter (see localized document)

#### What information was incorrect, unhelpful, or incomplete?

There are broken images in the live samples. Update is need for those examples.

#### What did you expect to see?

Update the examples and the images are not broken.

> Note: this is not the original issue (see the edit history for details, it's a system issue filled in mdn/yari#5652).

<details>
<summary>MDN Content page report details</summary>

* Folder: `zh-cn/web/css/filter`
* MDN URL: https://developer.mozilla.org/zh-CN/docs/Web/CSS/filter
* GitHub URL: https://github.com/mdn/translated-content/blob/main/files/zh-cn/web/css/filter/index.html
* Last commit: https://github.com/mdn/translated-content/commit/fb769838d655c3e6d28afdb3236ab9f3d9712d10
* Document last modified: 2022-01-20T00:58:58.000Z

</details>

yin1999:
Those examples should be synced with english version. I'll work on it.

yin1999:
@SphinxKnight this issue is both on:

- [de](https://developer.mozilla.org/de/docs/Web/CSS/filter)
- [es](https://developer.mozilla.org/es/docs/Web/CSS/filter)
- [fr](https://developer.mozilla.org/fr/docs/Web/CSS/filter)
- [ja](https://developer.mozilla.org/ja/docs/Web/CSS/filter)
- [ko](https://developer.mozilla.org/ko/docs/Web/CSS/filter)
- [ru](https://developer.mozilla.org/ru/docs/Web/CSS/filter)
- [zh-CN](https://developer.mozilla.org/zh-CN/docs/Web/CSS/filter)

yin1999:
> @SphinxKnight this issue is both on:
> 
> * [de](https://developer.mozilla.org/de/docs/Web/CSS/filter)
> * [es](https://developer.mozilla.org/es/docs/Web/CSS/filter)
> * [fr](https://developer.mozilla.org/fr/docs/Web/CSS/filter)
> * [ja](https://developer.mozilla.org/ja/docs/Web/CSS/filter)
> * [ko](https://developer.mozilla.org/ko/docs/Web/CSS/filter)
> * [ru](https://developer.mozilla.org/ru/docs/Web/CSS/filter)
[...]

PR #6270, opened 2022-06-14
Fix #5717

SphinxKnight:
Fixes #5717

- [x] [`files/fr/learn/css/css_layout/positioning`](https://github.com/mdn/translated-content/blob/main/files/fr/learn/css/css_layout/positioning/index.md)
- [x] [`files/fr/web/api/batterymanager/level`](https://github.com/mdn/translated-content/blob/main/files/fr/web/api/batterymanager/level/index.md)
- [x] [`files/fr/web/api/document/caretrangefrompoint`](https://github.com/mdn/translated-content/blob/main/files/fr/web/api/document/caretrangefrompoint/index.md)
- [x] [`files/fr/web/api/event/eventphase`](https://github.com/mdn/translated-content/blob/main/files/fr/web/api/event/eventphase/index.md)
- [x] [`files/fr/web/css/_colon_-moz-list-bullet`](https://github.com/mdn/translated-content/blob/main/files/fr/web/css/_colon_-moz-list-bullet/index.md)
- [x] [`files/fr/web/css/font-variant-alternates`](https://github.com/mdn/translated-content/blob/main/files/fr/web/css/font-variant-alternates/index.md)
- [x] [`files/fr/web/css/font-variant-ligatures`](https://github.com/mdn/translated-content/blob/main/files/fr/web/css/font-variant-ligatures/index.md)
- [x] [`files/fr/web/svg/attribute/stroke-dasharray`](https://github.com/mdn/translated-content/blob/main/files/fr/web/svg/attribute/stroke-dasharray/index.md)
- [x] [`files/fr/web/svg/element/radialgradient`](https://github.com/mdn/translated-content/blob/main/files/fr/web/svg/element/radialgradient/index.md)

PR #14992, opened 2023-08-08
es: update css/filter

Graywolf9:
### Description

Sync with upstream

Relates to https://github.com/mdn/translated-content/issues/3752

### DIFF
--- a/files/es/web/css/filter/index.md
+++ b/files/es/web/css/filter/index.md
@@ -1,24 +1,22 @@
 ---
 title: filter
 slug: Web/CSS/filter
+l10n:
+  sourceCommit: 856b52f634b889084869d2ee0b8bb62c084be04d
 ---
 
-{{SeeCompatTable}}{{CSSRef}}
+{{CSSRef}}
 
-## Resumen
+La propiedad **`filter`** de [CSS](/es/docs/Web/CSS) aplica efectos gráficos como desenfoque o cambio de color a un elemento. Los filtros se usan comúnmente para ajustar la representación de imágenes, fondos y bordes.
 
-La propiedad CSS **`filter`** dota de efectos gráficos como el desenfoque o cambio de color en la representación antes de que se muestre el elemento. Los filtros se utilizan comúnmente para ajustar la representación de imágenes, fondos o bordes.
+Varias [funciones](#funciones), como `blur()` y `contrast()`, están disponibles para ayudarte a lograr efectos predefinidos.
 
-Hay varias funciones Incluidas en el estándar CSS que logran efectos predefinidos. También puede hacer referencia a un filtro especificado en SVG con una URL a un [filtro de un elemento SVG](/es/docs/Web/SVG/Element/filter).
-
-> **Nota:** Versiones anteriores (4.0 hasta 9.0) del navegador Internet Explorer de Windows admiten una propiedad ["filter"](<https://msdn.microsoft.com/es-es/library/ms532853(v=vs.85).aspx>) no incluida en el estándar que ha quedado en desuso.
-
-{{cssinfo}}
+{{EmbedInteractiveExample("pages/css/filter.html")}}
 
 ## Sintaxis
 
 ```css
-filter: url("filters.svg#filter-id");
+/* Valores de <filter-function> */
 filter: blur(5px);
 filter: brightness(0.4);
 filter: contrast(200%);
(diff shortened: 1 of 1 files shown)


files/es/web/css/filter/index.md
```diff
@@ -30,1214 +28,192 @@ filter: opacity(25%);
...
 
-{{EmbedLiveSample('sepia_example','100%','229px','')}}
+```html
+<img src="pencil.jpg" alt="La imagen original es nítida" />
+<img
+  src="pencil.jpg"
+  alt="La imagen y el borde están borrosos y silenciados." />
+```
+
+{{EmbedLiveSample('Applying_filter_functions','100%','229px')}}
```
yin1999, line 182: We should match [the heading](https://github.com/mdn/translated-content/pull/14992/files#diff-58885109584a3af1c4e2de2ebda18d2692939e592c43df78a366ba45ef2f718fR161), or use the id generated by Yari: `aplicar_funciones_de_filtro` (case insensitive).

You could get this through devtools:

image

Reference: [MDN Live sample](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Live_samples#embedlivesample_macro)

```suggestion
{{EmbedLiveSample('Aplicar funciones de filtro','100%','229px')}}
```

files/es/web/css/filter/index.md
```diff
@@ -30,1214 +28,192 @@ filter: opacity(25%);
...
-  text-align: left;
-  vertical-align: top;
-  width: 25%;
-  height: auto;
-}
-#img3 {
-  height: 100%;
-}
-```
+{{EmbedLiveSample('Repeating_filter_functions','100%','229px')}}
```
yin1999, line 214: So we should also modify this line.

```suggestion
{{EmbedLiveSample('Repetir funciones de filtro','100%','229px')}}
```

JuanVqz (approved):

files/es/web/css/filter/index.md
Graywolf9: Hi! Thanks for the review! @yin1999 

I remember from the conversation on removing macro transclusion that these parameter are not required anymore so you can use `{{EmbedLiveSample}}` without issues (as you can see [here](https://github.com/mdn/translated-content/blob/921baf1dc1827f133afd60f05f937d71cc57ef29/files/fr/web/api/batterymanager/level/index.md?plain=1#L38) from this [PR](https://github.com/mdn/translated-content/pull/6270)), and it seems that Yari throw flaws if the names are different, that's why I keep like that

yin1999: Yes, yari throws bugs, but I believe this is caused by requiring localized headings to use the same IDs as the English docs when we write the docs in HTML: https://github.com/mdn/translated-content/blob/0caa9937af24d77b83ea51de17d14673fd729f8e/docs/README.md?plain=1#L32-L40

Graywolf9: I don't know why it is, but this is the message when the parameter is different to `en-US`

imagen

yin1999: Hey @Graywolf9. Localization support in yari is slow at the moment, and I think we'll have to ignore with the flaws it throws for now. Because if the macro doesn't find the corresponding ID, it will contain the wrong code block (use the [closest heading instead](https://github.com/mdn/yari/blob/e09ed3bc39900f500a02e6a29d335926e0673d09/client/src/document/code/playground.ts#L161)), and I have encountered this problem many times (use the wrong id and broken the livesample), I've provided a example here: #15137.

The sender and receivers are all in the same iframe ([broken example](https://pr15137.content.dev.mdn.mozit.cloud/zh-CN/docs/Web/API/BroadcastChannel/message_event#%E7%BB%93%E6%9E%9C)):

image

While they should be spread into three iframes ([correct example](https://developer.mozilla.org/zh-CN/docs/Web/API/BroadcastChannel/message_event#%E7%BB%93%E6%9E%9C)):

image

@queengooborg, may be we should remove the different id flaws report from [flaw system](https://github.com/mdn/yari/blob/main/build/flaws/translation-differences.ts), and should throw flaws if the incorrect id is used here (if set). This makes contributors use the wrong ids.

queengooborg: Yeah, that flaw sounds like it's causing more issues than it's helping, not to mention we have a better way of tracking translation differences now.  Unfortunately, I don't have write access to Yari (and my PRs are considered "community" PRs...even though I'm a part of the MDN team...), so I cannot make those changes.  :(

Graywolf9: Thank you so much for the example! @yin1999 As I can see there is a problem only when the macro is used more than once in a header

Thank you so much again!

PR #14925, opened 2023-08-05
Overhaul contribution docs

queengooborg:
This PR overhauls the contribution docs, performing a few main updates:

- Adds CONTRIBUTING.md with a structure similar to upstream
  - Moves the "Getting started" section in the main readme to there
  - Adds a new section for common contribution types
  - Better explains the setup process
- Updates the general translation guidelines
  - Removes outdated guidelines and adds new ones

PR #15050, opened 2023-08-11
chore(ci): remove 'Check for duplicates' workflow

yin1999:
### Description

Since all open PRs' changed (`add`, `mv`, or `update`) documents in the translated-content repository are in markdown format, I propose to remove the 'Check for duplicates' workflow.

### Motivation

Since we currently only use markdown format, and the documents in the repository are also [updated according to markdown format](https://github.com/mdn/translated-content/pull/14925), I think it's time for us to remove the workflow of checking whether `index.md` is duplicated with the `index.html` document.


queengooborg (approved):
Farewell, redundant workflow, you will (not) be missed!

PR #15051, opened 2023-08-11
fix a typo in the CSS document

EvenZhu:
### Description

Fixed a typo.

### Motivation

I found this text error while reading the documentation, which may cause users to wonder about the technical details described in the document, and the problem will be solved after fixing.


yin1999 (approved):
LGTM. Thanks @EvenZhu.

PR #15052, opened 2023-08-11
chore(autocorrect): remove single case in ignore rules

yin1999:
### Description

remove single case in ignore rules


jasonren0403 (approved):
👍

PR #15054, opened 2023-08-11
zh-cn: translate `getBigUint64()` method

btea:


yin1999 (approved):
Thanks. I've allowed myself to make a full translation.

PR #15058, opened 2023-08-11
zh-cn: translate replaceChild returned value

btea:


yin1999 (approved):
Thanks for translate this line. I've allowed myself to sync en-US content for these modifications.

PR #15081, opened 2023-08-13
zh-tw: Remove untranslated files

queengooborg:
This PR removes a number of untranslated files from the Traditional Chinese locale.


yin1999 (approved):
LGTM. Thank you @queengooborg

PR #15083, opened 2023-08-13
zh-tw: Remove untranslated files

queengooborg:
This PR removes various untranslated files from the Traditional Chinese locale.


yin1999 (approved):
Thanks. Let's remove them!

PR #15095, opened 2023-08-14
zh-cn: Remove untranslated files

queengooborg:
This PR removes various untranslated files from the Simplified Chinese locale.


yin1999 (approved):
Thanks for the cleanup @queengooborg. I'm still amazed at so many untranslated documents we've introduced when we are on wiki.

PR #15106, opened 2023-08-14
fix(ci): Reuse markdownlint config files for 'files' and 'docs' hiera…

OnkarRuikar:
Addresses https://github.com/mdn/translated-content/pull/14246#issuecomment-1676833120

> No, we don't need different definitions for the `docs` folder -- what I mean is, the `extends` property isn't working.

The config under `files/zh-cn/` directory can be reused in `docs/zh-cn/` using the changes in the PR. Same for `zh-tw` locale.

## Explanation

The config file `.markdownlint.jsonc` is related to the core `markdownlint` library. And `.markdownlint-cli2.jsonc` is related to `markdownlint-cli2`	tool that uses the markdownlint.

The `extends` property belongs to the `config` property in `.markdownlint-cli2.jsonc`. Under the hood the `config` property is the configuration for `.markdownlint.jsonc`. As the `config` is solely related to `makrdownlint` the  `extends` need to point to `.markdownlint.jsonc` files not `.markdownlint-cli2.jsonc`.

**Note** `.markdownlint-cli2.jsonc` is used only at the root level to provide CLI specific configs. And all core markdownlint related configurations have been put in `.markdownlint.jsonc` files.


OnkarRuikar:
@queengooborg we can go with this as this is [what you were trying to implement](https://github.com/mdn/translated-content/pull/14246#issuecomment-1676833120).

queengooborg (approved):
Confirming this works as intended.

I see there's a Prettier formatting performed as well, but this was probably performed during the merge commit.

yin1999:
Thanks @OnkarRuikar and @queengooborg!

PR #15111, opened 2023-08-15
es: Remove wikihistory from untranslated files

queengooborg:
This PR removes the wiki history from untranslated documents from the Spanish locale.  Since the users attributed did not actually contribute to localization, there's no reason to give them credit.


queengooborg:
I see that this PR has been treated as a breakdown list of pages to translate.  I'm glad to see these files getting taken care of!  Although, I would have recommended that we take care of other pages first, particularly translated ones that don't yet have a `sourceCommit` set and are probably quite out of date.  Those are likely more problematic, as we can't really track those programmatically!

queengooborg:
Looks like all of the documents this PR would delete have now been localized, so I've stripped this PR down to only update the wikihistory and remove undue attribution.  Self-merging.

PR #15115, opened 2023-08-15
ja: Remove untranslated files

queengooborg:
This PR removes various untranslated files from the Japanese locale.


mfuji09 (approved):
LGTM.

PR #15121, opened 2023-08-15
pt-br: Remove untranslated files

queengooborg:
This PR removes many untranslated files from the Brazilian Portuguese locale.


queengooborg:
Self-merging as there has been no review for over a week.

PR #15141, opened 2023-08-16
Update paths in markdown-lint.yml

queengooborg:
This updates the `markdown-lint` workflow to check for changes to any `.markdownlint.jsonc` file as well. This is a follow-up to #15106.

### DIFF
--- a/.github/workflows/markdown-lint.yml
+++ b/.github/workflows/markdown-lint.yml
@@ -7,6 +7,8 @@ on:
     paths:
       - .nvmrc
       - .markdownlint-cli2.jsonc
+      - .markdownlint.jsonc
+      - "**/.markdownlint.jsonc"
       - package.json
       - yarn.lock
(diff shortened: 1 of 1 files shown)


.github/workflows/markdown-lint.yml
```diff
@@ -7,6 +7,8 @@ on:
     paths:
       - .nvmrc
       - .markdownlint-cli2.jsonc
+      - .markdownlint.jsonc
+      - "**/.markdownlint.jsonc"
```
yin1999, line 11: I think we could just use:

```suggestion
      - "**/.markdownlint.jsonc"
```

documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths

image

yin1999 (approved):
Looks good to me. Thanks @queengooborg :+1:

PR #15278, opened 2023-08-20
fix: remove invalid option for front-matter_linter.js

SnowMarble:
I found there's invalid option(`--files`) for front-matter lint script, which causes the workflow to fail.
You can see https://github.com/mdn/translated-content/actions/workflows/markdown-lint-fix.yml to find more information about it.

image

so, I remove the option.


yin1999 (approved):
Good catch! Thank you @SnowMarble :+1:

PR #15329, opened 2023-08-22
es: Redirect MDN/Structures as en-US

Graywolf9:
### Description

Redirect MDN/Structures as en-US in https://github.com/mdn/content/pull/19117

In `content` repo the `structures` directory was moved to `Writing_guidelines/Page_structures`
The file was already created and updated in `es` so I just delete and redirect to the new file

`MDN/Contribute/Howto/Convert_code_samples_to_be_live` and `MDN/Structures/Live_samples` both was redirected to `MDN/Writing_guidelines/Page_structures/Live_samples` the first one keep and the second one moved as conflicting, but the second was the correct, so deleted the first one and moved the second one


JuanVqz (approved):
was it moved?

Graywolf9:
Yes, there is no so much information in the mentioned PR, but in files and redirects it shows that all under `mdn/structures` was moved to `Writing_guidelines/Page_structures`

Graywolf9:
Sorry, there was another file under that directory that was needed to fix

PR #15508, opened 2023-08-30
[zh-CN]: batch replace "您" with "你"

hanyujie2002:
### Description

replace "您" with "你"
### Motivation

ditto
### Additional details
My command: go to folder `files/zh-CN`, then run command `find ./ -type f -exec sed -i 's/您/你/g' {} \;` In this command, `-type f` means regular file(different than folders, symbolic files and device files). `{}` is replaced with file name when running command.


hanyujie2002:
I don't understand why some image was broken after the command, is that because there is some binary that coincidently be the same as the binary of "您"？

files/zh-cn/learn/forms/advanced_form_styling/chrome-mac-select-1-closed.png
yin1999, line 1: Could you filter out non-markdown file first

PR #15515, opened 2023-08-30
[zh-cn]: batch replace "您" with "你" part two

hanyujie2002:
### Description

replace "您" with "你"
### Motivation

ditto
### Additional details
My command: go to folder `files/zh-CN`, then run command `find ./ -type f -name ".md" -exec sed -i 's/您/你/g' {} \;` In this command, `-type f` means regular file(different than folders, symbolic files and device files). `{}` is replaced with file name when running command. 

By the way, I find that VS Code also provide native support for batch replacement

image

None


yin1999 (approved):
Thanks

PR #15520, opened 2023-08-30
[zh-cn] markdown syntax error

Rainyrou:
### Description

Removed a backtick (`) so that the markdown syntax could work correctly at this point.This is to keep consistency with the context.


jasonren0403 (approved):
Good catch!

jasonren0403:
Noticed that the document is not fresh enough, let's do the sync later 🤔

PR #15523, opened 2023-08-30
ja: Format remaining files using Prettier

queengooborg:
This PR formats the remaining files of the Japanese locale using Prettier.  These are the very last files that do not adhere to Prettier formatting.


queengooborg:
Looks like there's only one file remaining with a very small formatting change.  Since this has been sitting around for a while, I'm going to self-merge this PR.

PR #15536, opened 2023-08-31
chore(zh-cn): add rule to disallow '您'

yin1999:
### Description

add rule to disallow '您'

### Motivation

We have added this rule in [translation guide](https://github.com/mdn/translated-content/blob/main/docs/zh-cn/translation-guide.md#%E4%BB%A3%E5%90%8D%E8%AF%8D).

### DIFF
--- a/files/zh-cn/.markdownlint.jsonc
+++ b/files/zh-cn/.markdownlint.jsonc
@@ -68,6 +68,7 @@
         "name": "force-pronoun",
         "message": "Don't use '您'",
         "searchPattern": "/您/g",
+        "replace": "你",
         "searchScope": "text"
       }
     ]
--- a/files/zh-cn/.markdownlint.jsonc
+++ b/files/zh-cn/.markdownlint.jsonc
@@ -66,9 +66,8 @@
       },
       {
         "name": "force-pronoun",
-        "message": "Don't use '您'",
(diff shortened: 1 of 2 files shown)


yin1999:
As we could be noted in vs code when we use "您" in text with [markdown-lint extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) installed. So I'm for adding replace field in the rule:

image

files/zh-cn/.markdownlint.jsonc
```diff
@@ -63,6 +63,13 @@
         "searchPattern": "/\\]\\(https?:\\/\\/localhost:\\d+\\//g",
         "replace": "](/",
         "searchScope": "text"
+      },
+      {
+        "name": "force-pronoun",
+        "message": "Don't use '您'",
```
jasonren0403, line 69: What about

```
Consider using '你' instead of '您'
```

I'm not sure if message here should instruct user what to replace 🤔

yin1999: Should we also remove the replace field?

jasonren0403: > Should we also remove the replace field?

Make sense 👍

PR #15570, opened 2023-09-01
batch fix space problem

hanyujie2002:
### Description
batch fix space problem

### Motivation
ditto

### Additional details

Sed don't support unicode, so we need to set it as `[㐀-䶵一-鿏]`

Example code, we hope `文[A` be fixed as `文 [A`, we could use the command: 

```bash
find ./ -type f -name "*.md" -exec sed -i 's/\([㐀-䶵一-鿏]\)\[\([A-Za-z]\)/\1 [\2/g' {} \;
```
then is `A)文` be fixed as `A) 文`, we could use the command:

```bash
find ./ -type f -name "*.md" -exec sed -i 's/\([A-Za-z]\))\([㐀-䶵一-鿏]\)/\1) \2/g' {} \;
```

What's more, there are over 1000 changed files, that's huge number so I write a bash script
 (first we use `git diff --name-only > file_names` in the `files/zh-cn` folder to get all the changed files' names)
```bash
#!/bin/bash

# Set the starting and ending numbers for the file names
start=1
end=10

# Loop through the numbers
for ((num=start; num<=end; num++))
do
  # Create the target file name
  target_file="files/zh-cn/file_name_$num"
  source_file="files/zh-cn/file_names"

  # Create the branch name
  branch_name="batch_space_fix_$num"

  # Cat the first 200 lines from the source file to the target file
  head --lines 200 "$source_file" > "$target_file"

  # Checkout to the new branch
  git checkout "$branch_name"

  # add 200 files from target file
  while read -r file; do git add "$file"; done < "$target_file"

  # Push the branch to the remote repository
  git push origin "$branch_name"

  # Switch back to the original branch
  git checkout batch_space_fix

  tail -n +200 "$source_file"> temp.txt && mv temp.txt "$source_file";
done

```
None


hanyujie2002:
OOPs, I forget that, only the first edition `文[A` must be problem, another one `A)文` is not necessary a problem

PR #15574, opened 2023-09-02
batch space fix part 2

hanyujie2002:
### Description

Same as title

### Motivation

fix the typesetting

### Additional details

Sed don't support unicode, so we need to set it as `[㐀-䶵一-鿏]`

Example code, we hope `文[A` be fixed as `文 [A`, we could use the command: 

```bash
find ./ -type f -name "*.md" -exec sed -i 's/\([㐀-䶵一-鿏]\)\[\([A-Za-z]\)/\1 [\2/g' {} \;
```

What's more, there are over 500 changed files, that's huge number so I write a bash script

```bash
#!/bin/bash

# create the file which contains all changed files' name
git diff --name-only > file_names

# Set the starting and ending numbers for the file names
start=1
end=3

# Loop through the numbers
for ((num=start; num<=end; num++))
do
  # Create the target file name
  target_file="files/zh-cn/file_name_$num"
  source_file="files/zh-cn/file_names"

  # Create the branch name
  branch_name="batch_space_fix_neo_$num"

  # Cat the first 200 lines from the source file to the target file
  head --lines 200 "$source_file" > "$target_file"

  # Checkout to the new branch
  git checkout "$branch_name"

  # add 200 files from target file
  while read -r file; do git add "$file"; done < "$target_file"

  # commit to origin repository
  git commit -m "batch space fix part $num"

  # Push the branch to the remote repository
  git push origin "$branch_name"

  # Switch back to the original branch
  git checkout batch_space_fix_neo

  # delete the first 200 lines of source file
  tail -n +200 "$source_file"> temp.txt && mv temp.txt "$source_file";
done
```


hanyujie2002:
Why it don't pass the check? Plz help me @jasonren0403

jasonren0403:
https://github.com/mdn/translated-content/actions/runs/6055501066/job/16434497593?pr=15574

Run logs show that markdownlint failed in these two files:
* files/zh-cn/web/api/fetch/index.md
* files/zh-cn/web/api/idbtransaction/index.md

But I'm not sure what's wrong with these files, because the diff seems identical. /cc @yin1999

yin1999:
> https://github.com/mdn/translated-content/actions/runs/6055501066/job/16434497593?pr=15574
> 
> Run logs show that markdownlint failed in these two files:
> 
> * files/zh-cn/web/api/fetch/index.md
> * files/zh-cn/web/api/idbtransaction/index.md
> 
> But I'm not sure what's wrong with these files, because the diff seems identical. /cc @yin1999

One more space would be required to align the table. But the reviewer dog could not give suggestions as there are only one row of the table has been modified.

Please run `yarn install` before `git commit` to trigger a local pre-commit action to format all the files.

yin1999:
By the way, we really need to handle all wrong formatting, but I feel like fixing this in the [upstream project](https://github.com/huacnlee/autocorrect) is a better option.

PR #15585, opened 2023-09-02
zh-cn: init the translation of "document a CSS property"

yin1999:
### Description

init the translation of "document a CSS property"


jasonren0403 (approved):
👍

PR #15602, opened 2023-09-03
[ko] web/html/reference update

soorokim:
### Description

#15008 작업 내용 중 web/html/reference

Relates to #15008


yechoi42 (approved):
기여 감사합니다!

PR #15618, opened 2023-09-04
FR: Fix default referrer-policy

p1gp1g:
### Description

The default referrer-policy is `strict-origin-when-cross-origin` which hasn't been updated on the directives' description. Note that it has been correctly set in the table at the end of the page.

### Motivation

The document is actually inconsistent.


SphinxKnight:
Merci pour cette correction @p1gp1g ! Je me suis permis un commit pour ajuster les macros/liens qui avaient bougé ou étaient KO.

SphinxKnight (approved):

Issue #12632, opened 2023-04-08
zh-CN: Block- and inline-related glossary revamp


yarusome:
https://github.com/mdn/translated-content/issues/12567#issuecomment-1499840029 对 CSS 的 block 和 inline 及其相关术语的译法作出了如下统一：

| 英文术语 | 中文译法 | 英文术语 | 中文译法 |
| - | - | - | - |
| block direction | 块向 | inline direction | 行向 |
| block end | 块末 | inline end | 行末 |
| block start | 块首 | inline start | 行首 |
| block-level | 块级 | inline-level | 行级 |
| block | 【形】块向（的）（= in the block direction） | inline | 【形】行向的（= in the inline direction） |
| block | 【形】区块（的）（= of the block） |
| block | 【名】区块 |

此议题用于追踪相关术语的更新进度，同时欢迎大家贡献力量。

---

以“内联”和“块”为关键词搜索结果如下（不保证涵盖出现其他旧式译法的文档）：

- [ ] `/glossary/`
  - [ ] `/glossary/block/css/index.md`
  - [ ] `/glossary/control_flow/index.md`
  - [ ] `/glossary/element/index.md`
  - [ ] `/glossary/flex_container/index.md`
  - [ ] `/glossary/grid_axis/index.md`
  - [ ] `/glossary/main_axis/index.md`
- [ ] `/learn/css/building_blocks/`
  - [ ] `/learn/css/building_blocks/index.md`
  - [ ] `/learn/css/building_blocks/a_cool_looking_box/index.md`
  - [ ] `/learn/css/building_blocks/advanced_styling_effects/index.md`
  - [ ] `/learn/css/building_blocks/cascade_and_inheritance/index.md`
  - [ ] `/learn/css/building_blocks/creating_fancy_letterheaded_paper/index.md`
  - [ ] `/learn/css/building_blocks/debugging_css/index.md`
  - [ ] `/learn/css/building_blocks/fundamental_css_comprehension/index.md`
  - [ ] `/learn/css/building_blocks/handling_different_text_directions/index.md`
  - [ ] `/learn/css/building_blocks/images_media_form_elements/index.md`
  - [ ] `/learn/css/building_blocks/organizing/index.md`
  - [ ] `/learn/css/building_blocks/overflowing_content/index.md`
  - [ ] `/learn/css/building_blocks/selectors/selectors_tasks/index.md`
  - [ ] `/learn/css/building_blocks/sizing_items_in_css/index.md`
[...]

PR #15622, opened 2023-09-04
[zh-cn] revise css index

jasonren0403:
### Description

revise translation of 'block' and update the document

#12632


yin1999 (approved):
Thanks

PR #15633, opened 2023-09-05
zh-cn: update the translation of "webkitdirectory"

yin1999:
### Description

update the translation of "webkitdirectory"


jasonren0403 (approved):
👍

PR #15667, opened 2023-09-06
fix(ko): correct the front matter key

yin1999:
### Description

correct the front matter key

### Motivation

see: https://github.com/mdn/translated-content/actions/runs/6091127264/job/16527220673#step:8:2528


SnowMarble:
Thanks for the PR 😂

hochan222 (approved):
@yin1999 Ah.. I made a mistake.. I thought I caught the spacing in the git action(for mdn-bot), but this was also the cause..

Thank you for your correction. Have a good weekend :)

PR #15709, opened 2023-09-08
es: fix url locales

yin1999:
### Description

fix url locales, using the script in #15638


Graywolf9 (approved):
Great! Thank you so much!

PR #15710, opened 2023-09-08
ja: fix url locales

yin1999:
### Description

fix url locales, using the script in #15638


mfuji09 (approved):
Thank you, @yin1999 !

PR #15718, opened 2023-09-08
Fix translation strings

tisma95:


queengooborg:
Looks like this PR is blank, potentially opened by mistake.

PR #15841, opened 2023-09-14
zh-cn: add translation for the "Issue with serializing circular references" section in JSON.stringify()

codersjj:
### Description

  add translation for the "Issue with serializing circular references" section in [JSON.stringify()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#issue_with_serializing_circular_references)


yin1999 (approved):
Great. Thank you @codersjj

PR #15942, opened 2023-09-19
chore(zh-cn): unify all usage of 'Objective' and 'Prerequisites'

jasonren0403:
### Description

make sure current learn area table use the same translations in table headers


yin1999 (approved):
Great job 👍

PR #15966, opened 2023-09-20
zh-cn: align the translation of "identity" to "标识"

hanyujie2002:
### Description

Same as title
### Motivation

"标识" is better for the translation of "identity" than "身份" in the context
### Additional details

None

None


yin1999 (approved):
:+1:

PR #15974, opened 2023-09-20
fr: update Web/CSS/CSS_grid_layout/Realizing_common_layouts_using_grids

SphinxKnight:
A general update of the page (first intended as a "first step before actually fixing #15354")


SphinxKnight:
not sure if anything needs fixing for#15354 but an update is welcome

tristantheb:
@SphinxKnight I don't see anything special in the examples, for me the only difference there could be would result from the screen size and/or browser window.

tristantheb (approved):

cw118 (approved):
Désolé pour le délai ici 😖 

Merci beaucoup Sphinx !

PR #16020, opened 2023-09-22
zh-cn: update the description section for the `Object.hasOwn()` page

codersjj:
### Description

Update the description section for the `Object.hasOwn()` page.


yin1999 (approved):
Good catch @codersjj ! It looks better now :+1:

PR #16159, opened 2023-09-27
fix(ru): live example in `background-clip` article

ivan-developer-01:
### Description

I went to the page [background-clip](https://developer.mozilla.org/ru/docs/Web/CSS/background-clip) and have noticed an error. The example does not work properly.

*\*I'm writing on English because it's easier for me.*

### Motivation

It fixes a bug that provides wrong example. **Currently, the example don't applies styles.**

### Additional details

- https://developer.mozilla.org/ru/docs/Web/CSS/background-clip

None.

### DIFF
--- a/files/ru/web/css/background-clip/index.md
+++ b/files/ru/web/css/background-clip/index.md
@@ -43,7 +43,7 @@ background-clip: unset;
 
 ## Примеры
 
-#### HTML
+### HTML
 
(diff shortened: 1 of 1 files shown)


sashasushko (approved):

sashasushko:
@ivan-developer-01 спасибо 🙏

PR #16287, opened 2023-10-02
Update index.md

moshuying:
### Description

delete '然后‘ from zh-cn/web/javascript/inheritance_and_the_prototype_chain/index.md

### Motivation

上文中介绍了`doSomeInstancing`向上查找一步`[[Prototype]]`后获得到了结果就会使用查找结果。下文中说明了如果向上查找不到就继续向上查找到`Object.prototype`，此时应使用一个承接关系词来表达因为上方的原因所以获得了如此这般的结果。而此处用了两个承接关系词并且其中“然后”使用的并不恰当，然后的意思是前面内容结束后，后面内容接着发生。所以此处应该删去然后。

>The above section describes how to use the result of `doSomeInstancing` when you get a result after looking up one step to `[[Prototype]]`. The following section explains that if you can't go up to `Object.prototype`, you should use one conjunction to express that you got such-and-such a result because of the above. In this case, two conjunctions are used and the use of "然后" is not appropriate, as "然后" means that after the previous content is finished, the next content will follow. Therefore, the word "然后" should be deleted.


yin1999 (approved):
Looks better. Thanks @moshuying 🙏

PR #16354, opened 2023-10-04
zh-CN: update the translation of CORS

zjffun:
For detail see: https://dochub.zjffun.com/translate/mdn-zh-cn/mdn/translated-content/main/files/zh-cn/web/http/cors/index.md


yin1999 (approved):
Thanks @zjffun 👍

Issue #16288, opened 2023-10-02
zh-CN: Proofreading and update plan for the JavaScript `Map` section


JasonLamv-t:
I plan to proofread and update the translation of the Javascript.Mapsection. Anyone is welcome to contribute. If you find any issues with the Simplified Chinese Map section or have already submitted a PR, please mention me so that I can add your work to the progress list below:

- [x] overview, in #16347.
- [x] constructor, in #16443.
- [x] @@iterator, in #16444.
- [x] @@species, in #16457.
- [x] clear, in #16458.
- [x] delete, in #16458.
- [x] entries, in #16459.
- [x] foreach, in #16460.
- [x] get, in #16458.
- [x] groupby, in #16460.
- [x] has, in #16460.
- [x] keys, in #16459.
- [x] set, in #16458.
- [x] size, in #16460.
- [x] values, in #16459.

yin1999:
Thank you @JasonLamv-t :tada:

PR #16460, opened 2023-10-08
[zh-cn]: update the translation of Map.size&forEach()&groupBy()&has()

JasonLamv-t:
### Description

Same as the title.

### Motivation

Ditto.

### Additional details

None.

#16288


yin1999 (approved):
Thanks

PR #16495, opened 2023-10-10
fix metadata key

yin1999:
### Description

fix metadata key


hochan222 (approved):
We missed that part. Thank you!! LGTM :)

yin1999:
> We missed that part. Thank you!! LGTM :)

Don't worry about this. I've fixed the [pr-check-lint_content](https://github.com/mdn/translated-content/blob/main/.github/workflows/pr-check-lint_content.yml) workflow, and now it could validate the metadata key :)

PR #16518, opened 2023-10-12
zh-cn: fix translate error for chained promises

CS-Tao:
### Description

A translate error, maybe typo?


yin1999 (approved):
Good catch @CS-Tao. Welcome aboard :tada:

PR #16603, opened 2023-10-17
Update AbortController Chinese doc

Meowu:
### Description

Update AbortController Chinese doc

### Motivation

The current translation is incorrect.


jasonren0403 (approved):
👍 Thanks @Meowu

PR #16606, opened 2023-10-17
[ko]: fix typo in Array.from() reference

omins:
### Description

Fix typo in `Array.from()` reference.
* "전체 배열 없어" -> "전체 배열 없이"

### Motivation

* Found typo while reading the docs.
* It seems make sense when "없어" has changed to "없이"


wisedog (approved):
LGTM

PR #16612, opened 2023-10-17
[ko]: fix typo in  String.prototype.toLocaleLowerCase() reference

Juyeong-Byeon:
### Description

fix typo in  String.prototype.toLocaleLowerCase() reference
"맴핑" -> "매핑"

### Motivation

It seems make sense when "맴핑" has changed to "매핑"
"맴핑" is not correct word.


jho2301 (approved):
기여 감사합니다 :) 좋은 주말 되세요

PR #16620, opened 2023-10-18
fix cookie-value wrap with double quotes zh-cn translation error

dispensable:
### Description
"A \<cookie-value\> can optionally be wrapped in double quotes ..." should be translated to Chinese as "可以选择包裹在双引号中". Original translation means cookie-value is optional itself and if exists, must be wrapped with double quotes.

### Motivation

translation error makes me confused.

### Additional details

- English version is correct: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes
- For compatibility, it's optional to wrap cookie-value with double quotes. https://stackoverflow.com/questions/25395845/double-quotes-and-trailing-equal-sign-missing-from-cookies-values-in-servletexec

N/A


yin1999 (approved):
Thanks @dispensable. Welcome to the community :tada:

PR #16688, opened 2023-10-24
fix typo

hbsowo58:
It seems that the typo was integer, not interger, which means integer.

I'm sorry if there are intergers I don't know about.


wisedog (approved):
감사합니다!

PR #16718, opened 2023-10-26
Update index.md

NVPanda:
### Description

I noticed that there were letters missing from some parts of the text. I included them. I hope my review makes a good change.
The change log will show up the letter "C" same as "H" from "HOW that means "Como" in brazilian portuguese.

### Motivation

To make it more attractive to natives of Brazil and draw attention to details that often go unnoticed by ourselves.
I've learning with MDN and would like to contribute many as possible of changes. 
Thanks for the documentation it's really unbelivable and make me so happy while studying.


josielrocha:
Obrigado pela ajuda @NVPanda!

PR #16729, opened 2023-10-27
ru: add embedded interactive example

rustam-mukhametshin:
Add embedded interactive example for ru page.

According to this page:
en: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

ru: 
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/assign


saionaro (approved):
@rustam-mukhametshin привет! Спасибо за вклад в MDN!

PR #16730, opened 2023-10-27
ru: update Object description

rustam-mukhametshin:
Update description accroding to this page:

En:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#description

Ru:
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object


rustam-mukhametshin:
Also I suppose, I need to remove **Syntax** and translate other stuff from **Description** too,
according to en version.

but I'll do it with small portions (PRs). (easier to review)

saionaro (approved):
@rustam-mukhametshin спасибо за вклад в MDN!

PR #16762, opened 2023-10-30
update the translation of `contenteditable`

kiling:
add value "plaintext-only" description for Chinese(Simplified) translation.


yin1999 (commented):
How about this?

kiling:
> How about this?

Your description is better.

yin1999 (approved):
Thanks @kiling

PR #16798, opened 2023-11-01
[ru] improve `Learn/HTML/Introduction_to_HTML/Creating_hyperlinks` translation

vncrrct:
### Description

 Small fix of example


sashasushko:
@vncrrct can I ask you to match Russian version of this article to [English one](https://github.com/mdn/content/blob/main/files/en-us/learn/html/introduction_to_html/creating_hyperlinks/index.md?plain=1)? The translation should be correct, only syntax is requiring the changes

vncrrct:
English version contains EmbedLiveSample for examples. I think it would be great to use live samples in Russian version, and I may help with it. But I consider this PR is not for it.

leon-win (approved):
@vncrrct thank you for your desire to help!

I agree with @sashasushko that when making changes to documents, it makes sense to take into account the current state of the original (`en-US` version).

So I updated the translation of this page, about a third of it, including the example you corrected))

PR #16802, opened 2023-11-02
zh-cn: update the content of the `Why use dns-prefetch?` section

codersjj:
### Description

Update the content of the `Why use dns-prefetch?` section.


jasonren0403 (approved):
Thanks, it looks better now 👍

PR #16872, opened 2023-11-08
ru: Fix formatting of Web/CSS/display-outside

queengooborg:
This PR fixes the formatting in the Web/CSS/display-outside page for the Russian locale, removing codeblocks applied to the entire bottom half of the page. This PR also replaces the `{{Compat}}` macro with a non-argument version.


leon-win (changes requested):
I think we should also remove the quotes from the headers:
```
### `HTML`
### `CSS`
### `Result`
### `Support of run-in`
## `Смотрите также`
```

Because they are also not needed there and apparently got there by accident))

queengooborg:
Good catch, @leon-win!  I completely skimmed over that, it looks like what happened was probably that the closing `<code>` tag was missing its slash when the file was converted from HTML to Markdown, resulting in the odd codeblock situation.

leon-win:
> Good catch, @leon-win! I completely skimmed over that, it looks like what happened was probably that the closing `<code>` tag was missing its slash when the file was converted from HTML to Markdown, resulting in the odd codeblock situation.

Yes, it's quite possible.
It's much better now, thank you, @queengooborg! :+1:
I would have approved if I had enough rights :slightly_smiling_face:

queengooborg:
I'll take that statement as enough approval and merge this.  ;)  (Self-merging since it's the last PR needed to complete the removal of `Compat` macro args!)

leon-win:
> I'll take that statement as enough approval and merge this. ;) (Self-merging since it's the last PR needed to complete the removal of `Compat` macro args!)

This is fair and logical! :+1:

PR #16873, opened 2023-11-08
Remove arguments to {{Compat}} macro calls

queengooborg:
This PR is a programmatic removal of all arguments in `{{Compat}}` macro calls, which is part of work for #11594.

Generated programmatically by https://github.com/queengooborg/mdn-toolkit/blob/main/replace-macro.mjs.


josielrocha (approved):
Approve for `pt-BR`

queengooborg:
If it looks good for one locale, it should be good for all locales -- merging!

PR #16894, opened 2023-11-10
Added missing words to the translation of the documentation for slice()

roose:
### Description

Added missing words to the translation of the documentation for `slice()`

### Motivation

In the translation, an important clarification about the `end` parameter was missed.


saionaro (approved):
@roose спасибо за вклад в MDN!

PR #16899, opened 2023-11-10
fix(ru): replace remaining calls to `{{htmlattrdef}}` macro

queengooborg:
This PR replaces all the remaining calls to the deprecated `{{htmlattrdef}}` macro.


yin1999:
Hey @queengooborg. It would be a good idea to also remove the `attr-` prefix in fragment like [this one](https://github.com/mdn/translated-content/blob/29c692d54a3fc7491ace057438c3bc0fd2f84020/files/ru/web/api/canvas_api/tutorial/pixel_manipulation_with_canvas/index.md?plain=1#L269) (but make sure that the referenced item is not wrapped with `SVGAttr`, like [this one](https://github.com/mdn/translated-content/blob/29c692d54a3fc7491ace057438c3bc0fd2f84020/files/ru/web/svg/attribute/core/index.md?plain=1#L9-L13))

```diff
- [download attribute](/ru/docs/Web/HTML/Element/a#attr-download)
+ [download attribute](/ru/docs/Web/HTML/Element/a#download)
```

After we remove the macro, those urls will be broken.

sashasushko (approved):

sashasushko:
@queengooborg tanks, changes LGTM. What do you think about the message from **yin1999**?

queengooborg:
That would be good to do, yeah!  Would you be down to open an issue for that, @yin1999?  (Those changes are beyond the scope of this PR.)

yin1999:
> That would be good to do, yeah! Would you be down to open an issue for that, @yin1999? (Those changes are beyond the scope of this PR.)

I would like to do this :)

PR #16908, opened 2023-11-10
`fr` - Initial translation for Glossary/UUID and crypto/randomuuid()

SphinxKnight:
Part of #16598 
Went with translating Web/API/Crypto/randomUUID as well :)


cw118:
Merci encore Sphinx !

PR #16911, opened 2023-11-11
chore(deps): replace "fs-extra" with "node:fs/promise"

yin1999:
### Description

As all the file system methods we used in this repo are available in "fs/promise", it would be ok to replace "fs-extra" and remove the dependency.


yin1999 (commented):
As this PR won't break those scripts, and used method are added in [node:v10](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api) (Meet our minimum node version requirements (v18)). I'm going to merge this

PR #16919, opened 2023-11-11
Remove confusing and misleading word in sentence about accessibility

sylbru:
### Description

The text contained the word "malvoyants" (≃ blind) where it was supposed to refer to sighted users.

### Motivation

The beginning was misleading and confusing because of this word.


SphinxKnight (approved):
Merci @sylbru pour cette correction !

PR #16924, opened 2023-11-12
zh-cn: sync the `Data.toJSON()` example from en-US

shiyubo:
The second log should be the same 54:04

### Motivation

Make it clearer.


yin1999 (approved):
Thanks @shiyubo. I've synced the example from en-US document to make this example  reproductible.

Issue #13499, opened 2023-05-30
Resolve macro render errors in translated-content


caugner:
We have begun collecting build errors during our nightly builds (see also: https://github.com/mdn/translated-content/issues/12578), and with https://github.com/mdn/yari/pull/8953 we now also see where other macros fail to render:

## es

- [x] The compat macro on /es/docs/Web/API/Fetch_API/Using_Fetch failed to render. Error: /home/runner/...
- [x] The compat macro on /es/docs/conflicting/Web/API/CustomElementRegistry/define failed to render. E...
- [x] The compat macro on /es/docs/conflicting/Web/API/Geolocation/getCurrentPosition failed to render....
- [x] The compat macro on /es/docs/conflicting/Web/CSS/outline failed to render. Error: /home/runner/wo...
- [x] The compat macro on /es/docs/conflicting/Web/HTML/Element/audio failed to render. Error: /home/ru...
- [x] The compat macro on /es/docs/conflicting/Web/HTML/Element/slot failed to render. Error: /home/run...
- [x] The compat macro on /es/docs/orphaned/Web/API/HTMLShadowElement failed to render. Error: /home/ru...
- [x] The compat macro on /es/docs/orphaned/Web/API/HTMLShadowElement/getDistributedNodes failed to ren...
- [x] The compat macro on /es/docs/orphaned/Web/API/NavigatorLanguage failed to render. Error: /home/ru...
- [x] The compat macro on /es/docs/orphaned/Web/API/NavigatorLanguage/language failed to render. Error:...
- [x] The compat macro on /es/docs/orphaned/Web/API/NavigatorOnLine failed to render. Error: /home/runn...
- [x] The compat macro on /es/docs/orphaned/Web/API/NavigatorOnLine/onLine failed to render. Error: /ho...
- [x] The specifications macro on /es/docs/Web/API/Fetch_API/Using_Fetch failed to render. Error: /home...
- [x] The specifications macro on /es/docs/Web/CSS/CSS_box_model/Introduction_to_the_CSS_box_model fail...
- [x] The specifications macro on /es/docs/Web/CSS/CSS_box_model/Mastering_margin_collapsing failed to ...
- [x] The specifications macro on /es/docs/Web/JavaScript/Data_structures failed to render. Error: /hom...
- [x] The specifications macro on /es/docs/conflicting/Web/API/CanvasRenderingContext2D failed to rende...
[...]

PR #16996, opened 2023-11-16
Remove obsolete `Node.isSupported` method

leon-win:
### Description

This PR remove obsolete `Node.isSupported` method from MDN translated content.

### Motivation

It has been removed from en-US locale a long time ago.

Relates to https://github.com/mdn/content/pull/22863, #13499


cw118 (approved):
Looks good for `l10n-fr`, thank you @leon-win !

(It may be easier/faster to remove all obsolete methods in one locale in each PR, instead of removing one obsolete method for all locales, because only one reviewer would be needed for each PR.)

leon-win:
> Looks good for `l10n-fr`, thank you @leon-win !
> 
> (It may be easier/faster to remove all obsolete methods in one locale in each PR, instead of removing one obsolete method for all locales, because only one reviewer would be needed for each PR.)

@cw118 thank you for reply!

I also wanted to do exactly that at first.
But then I thought that the changes are almost identical for all languages (might say trivial), and therefore one reviewer can look at them all :)
That is, in my opinion, the changes are understandable even without knowledge of a specific language. How do you think?

cw118:
> I also wanted to do exactly that at first. But then I thought that the changes are almost identical for all languages (might say trivial), and therefore one reviewer can look at them all :) That is, in my opinion, the changes are understandable even without knowledge of a specific language. How do you think?

I agree that the changes are very small for all locales, but it happens somewhat often that locales have different conventions and requests, so even for this type of PR we prefer to wait for approval from a reviewer for each locale. Thanks for understanding!

leon-win:
> > I also wanted to do exactly that at first. But then I thought that the changes are almost identical for all languages (might say trivial), and therefore one reviewer can look at them all :) That is, in my opinion, the changes are understandable even without knowledge of a specific language. How do you think?
> 
> I agree that the changes are very small for all locales, but it happens somewhat often that locales have different conventions and requests, so even for this type of PR we prefer to wait for approval from a reviewer for each locale. Thanks for understanding!

I'm understood, thank you!
I actually have already created several more similar PRs :smiling_face_with_tear: 
I am still convinced that bureaucracy can be unnecessary and can hinder progress.
In any case, the amount of time spent will be the same (if we wait for confirmation from all reviewers) -- it doesn’t matter one or several PR's =)

Graywolf9 (approved):
LGTM for `es`

As @cw118 says, sometimes is better a PR per locale, so the locale can be updated soon, and then you can identify older locales that maybe ended hidden. Other times is good to have a unified change of all locales in a PR, we respect each locale lead so we wait to their review, depends on the change after two weeks someone more of the core team could approve the whole PR

Thank you so much for all your work and time!

JuanVqz (approved):

yin1999 (approved):
LGTM for `zh-CN`. Thank you @leon-win

saionaro (approved):
LGTM for RU

leon-win:
> LGTM for `es`
> 
> As @cw118 says, sometimes is better a PR per locale, so the locale can be updated soon, and then you can identify older locales that maybe ended hidden. Other times is good to have a unified change of all locales in a PR, we respect each locale lead so we wait to their review, depends on the change after two weeks someone more of the core team could approve the whole PR
> 
> Thank you so much for all your work and time!

Thanks for the explanation, @Graywolf9 :hugs: 
I will keep it in mind in the future!

josielrocha (approved):
Approve for `pt-BR`

PR #17010, opened 2023-11-18
zh-cn: init the translation of `Element.setHTML()`

yin1999:
### Description

init the translation of `Element.setHTML()`

### References

The translation of `Santizer` is from:

- https://learn.microsoft.com/zh-cn/cpp/sanitizers/asan?view=msvc-170
- https://developer.nvidia.com/zh-cn/blog/debugging-cuda-more-efficiently-with-nvidia-compute-sanitizer/


jasonren0403 (approved):
LGTM 👍 Thanks @yin1999

PR #17069, opened 2023-11-22
chore(zh-cn): batch replace '相关链接' to '参见' part 3

jasonren0403:
### Motivation

enforce translation guide


yin1999 (approved):
Thank you @jasonren0403

PR #17122, opened 2023-11-26
zh-cn: fix typo

HanaChan233:
### Description

const sortedValues = values.toSorted((a, b) => a - b));
to
const sortedValues = values.toSorted((a, b) => a - b);

### Motivation

原来的代码多了个右括号，删掉了


yin1999 (approved):
Good catch! Welcome aboard @HuaChan233 :tada:

PR #17144, opened 2023-11-27
Fix wrong code example

Luise8:
### Description

There is an error. The example doesn't show the text when the option change because the resultado varible is create inside of the event listener. The declaration of the "resultado" should be outside of event listener.  If you see the example in english it works due to this small change: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event 
const resultado = document.querySelector(".resultado");
listener...

### Motivation

The example doesn't work. The declaration of the "resultado" variable should be outside of event listener.


JuanVqz (approved):
Hey @Luise8, thank you for fixing it, it's a great contribution 🚀

PR #17177, opened 2023-11-29
zh-cn: update the translation of <acronym> element

yin1999:
### Description

update the translation of <acronym> element


jasonren0403 (approved):
Thanks @yin1999

PR #17200, opened 2023-11-30
Fix: modify main.js link address

dmwin72015:


yin1999 (approved):
Good catch @dmwin72015! Welcome to the community :tada:

PR #17207, opened 2023-11-30
zh-CN: Synchronize the translation of Blob.arrayBuffer

Yukiniro:
en-US: https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer
zh-CN: https://developer.mozilla.org/zh-CN/docs/Web/API/Blob/arrayBuffer


files/zh-cn/web/api/blob/arraybuffer/index.md
```diff
@@ -1,37 +1,33 @@
 ---
-title: Blob.arrayBuffer()
+title: Blob: arrayBuffer() 方法
```
yin1999, line 2: The problem why the test is failing is: the value of the metadata could not contain `:` without quotation marks wrapping. And for zh-CN, please use _full-width colon_（`：`）.

```suggestion
title: Blob：arrayBuffer() 方法
```

yin1999 (commented):
Let's align the translation format of "promise that resolves with" with other documentation.

yin1999 (approved):
Thanks

PR #17209, opened 2023-11-30
zh-cn: update the description of `receiver` parameter in `handler.set()` method

codersjj:
### Description

Update the description of `receiver` parameter in `handler.set()` method.

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/set#receiver


yin1999 (approved):
LGTM. Thank you @codersjj

PR #17233, opened 2023-12-02
[fr] Adds double-quotes around layer name for @import

jmpp:
It appears that without double-quotes around the layer name, current modern browsers won't respect the last layer priority if it implies a "!important" inside the first imported layer.

For example, this code should make any element having the "bg-primary" class red, because it's defined in the last layer and also have an "!important" :

```css
@layer framework, base;
@import 'https://cdnjs.cloudflare.com/ajax/libs/bootstrap/4.5.3/css/bootstrap.min.css'
    layer(framework);

@layer base {
  .bg-primary {
    background-color: red !important;
  }
}
```
```html
<div class="bg-primary">Lorem ipsum</div>
```
But instead of this, elements are still rendering blue (the default color that's defined inside "bootstrap.min.css" also with an "!important".

Using double quotes around the "framework" will solve the problem :

```css
@import 'https://cdnjs.cloudflare.com/ajax/libs/bootstrap/4.5.3/css/bootstrap.min.css'
    layer("framework");
```
Now the div will turn red as it's initially expected.

### Description

I've just added double quotes around the example to make it work in all situations (even those that implies "!important" rules.

### Motivation

I've struggled to understand why my code didn't work whereas I was following exactly what the Docs said. I want to help people not loosing the time I did lose while reading an "official" resource.

### DIFF
--- a/files/fr/web/css/@layer/index.md
+++ b/files/fr/web/css/@layer/index.md
@@ -78,7 +78,7 @@ Une autre façon pour créer une couche de cascade consiste à utiliser [`@impor
 > **Attention :** La règle @ `@import` doit précéder tous les autres types de règles, à l'exception des règles `@charset`.
 
 ```css
-@import "theme.css" layer(utilities);
+@import "theme.css" layer("utilities");
 ```
 
 ### Couches imbriquées


SphinxKnight:
See also: 
https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade#importance_and_layers
https://drafts.csswg.org/css-cascade-5/#cascade-layering

SphinxKnight:
@jmpp `!important!` reverses the order of precedence of the layers (as explained in https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade#importance_and_layers) 
That explains the result you have in your example without the quotes. I made a (horrendous) CodePen here https://codepen.io/SphinxKnight/pen/poGqJqp 

However, adding quotes is not the right solution as it breaks the import. As far as I can tell, no rule is imported anymore where referring to `framework` and using `layer("framework")`. For instance, when I add quotes to my CodePen, we can see that `text-decoration` is not taken into account:

image

In regards to this PR, I think we should close it and/or propose a new one. The underlying problem with the docs now is that the information from https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade#importance_and_layers cannot easily be found from https://developer.mozilla.org/en-US/docs/Web/CSS/@layer. Adding a note including a small explanation and a pointer for more details would be helpful (first in mdn/content, then in mdn/translated-content).

@jmpp is this answer clear/helpful?

SphinxKnight (changes requested):
cf. comment above

jmpp:
@SphinxKnight Thank you for your detailed response and in-depth research on the subject.

I clearly missed the explanation regarding reversing the precedence order. That explains lots of things.

Now I do agree that this should be explained a bit differently in the docs (maybe with screen captures like you did) to make is more clear.

I'll think about making something more properly.

Best

SphinxKnight:
Closing this one for now in favor of proposing a nicer wording and warning on the upstream English content.

PR #17304, opened 2023-12-07
fix JSX code diff with original English edition

fengliutie:
fix JSX code difference between translated Chinese edition and original English edition


yin1999:
Thanks @fengliutie. I've pushed the fix and it looks good to me now. Welcome the community 🎉

yin1999 (approved):

PR #17345, opened 2023-12-08
init translation for FontData

skyclouds2001:
### Description

see https://developer.mozilla.org/en-US/docs/Web/API/FontData


files/zh-cn/web/api/fontdata/blob/index.md
```diff
@@ -0,0 +1,73 @@
...
+
+{{Specifications}}
+
+## 浏览器兼容性
+
+{{Compat}}
+
+## 参见
+
+- [使用带有本地字体的高级排版](https://developer.chrome.com/articles/local-fonts/)
```
yin1999, line 72: Follow the redirects: mdn/content#30896

yin1999 (approved):
LGTM. Thanks @skyclouds2001 :+1:

PR #17410, opened 2023-12-11
[zh-cn]: corrected translation error of size attribute of  `<font>` element

meowrain:
### Description

 Corrected translation error of size attribute of  `<font>` element

### Motivation

In the original article, the minimum value of the size attribute is written as "minimum 7", which is wrong and should be 1.
The original translation is misleading.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font


yin1999 (approved):
Thanks @meowrain

PR #17439, opened 2023-12-13
[es] Corrected Grammar error

MrDaPoyo:
"le" isn't correct, while "el" is


JuanVqz (approved):
Hey @MrDaPoyo gracias por la aportación, saludos!

PR #17492, opened 2023-12-16
Remove unusual closing brackets in FR translation

bluwv:
### Description

There was typo with closed brackets near HTML tag on the FR translation:
https://developer.mozilla.org/fr/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

### Motivation

This  could seems weird for a beginner who learn  code to see something like this for an HTML tag:
<picture>}


SphinxKnight (approved):
Thanks a lot @bluwv!

Issue #17542, opened 2023-12-19
[zh-cn] init translation for Web Locks API pages plans


skyclouds2001:
init translation for Web Locks API pages

```[tasklist]
### Web Locks API translation plans
- [x] [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) in #17567
- [x] [Navigator.locks](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/locks) in #17572
- [x] [WorkerNavigator.locks](https://developer.mozilla.org/en-US/docs/Web/API/WorkerNavigator/locks) in #17572
- [x] [LockManager](https://developer.mozilla.org/en-US/docs/Web/API/LockManager) in #17620
- [x] [LockManager.query()](https://developer.mozilla.org/en-US/docs/Web/API/LockManager/query) in #17766
- [x] [LockManager.request()](https://developer.mozilla.org/en-US/docs/Web/API/LockManager/request) in #17767
- [x] [Lock](https://developer.mozilla.org/en-US/docs/Web/API/Lock) in #17621
- [x] [Lock.mode](https://developer.mozilla.org/en-US/docs/Web/API/Lock/mode) in #17717
- [x] [Lock.name](https://developer.mozilla.org/en-US/docs/Web/API/Lock/name) in #17717
```

PR #17572, opened 2023-12-21
init translation for `Navigator.locks` & `WorkerNavigator.locks`

skyclouds2001:
### Description

see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/locks and https://developer.mozilla.org/en-US/docs/Web/API/WorkerNavigator/locks

part of #17542


yin1999 (approved):
LGTM. Thank you @skyclouds2001

PR #17587, opened 2023-12-22
fix: remove extra curly braces from macro calls

caugner:
### Description

Removes extra curly braces before/after macro calls.

### Motivation

Macro calls require exactly two curly braces both before and after, but as part of https://github.com/mdn/yari/pull/10224 I noticed that there were two macro calls with more leading curly braces, which should result in a macro render error. Looking closer, I also found several occurrences with more trailing curly braces, one of which looks like it should have been a closing parenthesis.

### Additional details

See https://developer.mozilla.org/ko/docs/Web/API/MouseEvent/clientX for failed macro renders:

image

See https://developer.mozilla.org/es/docs/Web/HTTP/CSP#describiendo_una_pol%C3%ADtica for how the additional trailing curly braces render as content:

image

PR #17599, opened 2023-12-24
fix(ru): remove extra curly braces from macro calls

jdeeline:
### Description

Removed third curly braces from macro calls.


yin1999:
duplicate of #17587

PR #17700, opened 2024-01-03
[zh-cn]: more accurate description for `side-or-conrner` value of `linear-gradient`

makedopamine:
### Description
I think it would be closer what is meant by the [original](https://github.com/mdn/content/blob/487a6049ab60bc78e1a076d9ce6684d0986122f3/files/en-us/web/css/gradient/linear-gradient/index.md?plain=1#L56).


yin1999 (approved):
Good catch @makedopamine. And welcome to the community as this is your first PR here :tada:

PR #17809, opened 2024-01-09
fix(zh-cn): remove wiki history to the deleted document

yin1999:
### Description

remove wiki history to the deleted document

#17805


yin1999 (commented):
self-approval

PR #17874, opened 2024-01-12
[ja] update positive lookahead matching explanation

joseph082:
### Description

The strings that the RegExps match are incorrect.

### Motivation

The matched strings are wrong, so these changes help readers better understand that lookaheads don't consume the strings they match.

See https://github.com/mdn/content/pull/30715 and https://github.com/mdn/content/pull/30715/files

### Additional details

If anyone wants to add a translation for `because lookaheads don't consume their matched strings` which was also added in https://github.com/mdn/content/pull/30715, please do.


mfuji09 (approved):
LGTM. Thank you.

PR #17912, opened 2024-01-16
zh-cn: update the translation of `<mspace>` MathML element

yin1999:
### Description

update the translation of `<mspace>` MathML element


jasonren0403 (approved):
LGTM, Thank you @yin1999

PR #17940, opened 2024-01-17
zh-cn: update the translation of HTMLElement.contentEditable property

xiaotong-tong:
Relates to #17934


yin1999 (commented):
Great. But can we translate it in a more concise form?

yin1999 (approved):
Thank you @xiaotong-tong :+1:

PR #17955, opened 2024-01-18
Spellcheck in french translation

sblondon:
Add missing 'e' character.


tristantheb (approved):
Thanks for your first contribution on the MDN ! Nice to see you agin if you want to help us to correct or translate pages :)

PR #17958, opened 2024-01-19
[ru] remove outdated warning about subgrid

anastasiavyalt:
### Description

Removed outdated warning about subgrid being available only in Firefox in russian translation. Subgrid is available in all major browsers now


myshov (approved):
Thanks!

PR #17976, opened 2024-01-20
[fr] remove useless translation on `Web/API/FileReader/readAsArrayBuffer`

MeNearly:
This translation may lead to misunderstanding..


cw118 (approved):
Thank you @MeNearly for this correction!

PR #18058, opened 2024-01-26
[zh-cn] fix: regexp

akijoey:
Fix a regexp error.


jasonren0403 (approved):
👍 Good catch, and welcome contributing to MDN!

PR #18085, opened 2024-01-29
[zh-cn] fix content error

teacup418:


yin1999 (approved):
Good catch @mind-dance

PR #18086, opened 2024-01-29
refactor(zh-cn): modify Node.ATTRIBUTE_NODE constant to use a string literal '2' for better readability

Saraph1nes:
### Description

This change updates the constant value for Node.ATTRIBUTE_NODE from the numeric literal 2 to the string literal '2' for improved code readability.


yin1999 (approved):
Thank you @Saraph1nes. As this table will be formated by prettier, so I pushed a fix commit to do this with format align to en-US document. Welcome to the community 🎉

Issue #17632, opened 2023-12-26
[zh-cn] Fix up missing HTML attributes translation


jasonren0403:
Fix up missing HTML attributes translation

```[tasklist]
### Global Attributes
- [ ] [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus)
- [ ] [`enterkeyhint`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/enterkeyhint)
- [ ] [`exportparts`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts)
- [ ] [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert)
- [ ] [`nonce`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce)
```

```[tasklist]
### Attributes
- [ ] [`accept`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept)
- [ ] [`capture`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture)
- [ ] [`dirname`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/dirname)
- [ ] [`elementtiming`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/elementtiming)
- [ ] [`for`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/for)
- [ ] [`max`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/max)
- [ ] [`maxlength`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength)
- [ ] [`min`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/min)
- [ ] [`minlength`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/minlength)
- [ ] [`multiple`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/multiple)
- [ ] [`placeholder`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/placeholder)
- [ ] [`readonly`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly)
- [ ] [`required`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required)
[...]

PR #18111, opened 2024-02-01
[zh-cn] init translation of size attribute

jasonren0403:

part of #17632


yin1999 (approved):
Thank you @jasonren0403

PR #18120, opened 2024-02-01
[zh-tw]: translate HTML element `<footer>`

Dr-XYZ:


yin1999 (approved):
Thank you @Dr-XYZ. We recommend to use `git` with code editor like vs code to contribute to MDN. As in this way, we can prevent the problem like `LF`/`CRLF` style and markdown style format problem. Using the way like "push commit via file upload" would make us always need to pay attention to these "less important" format issues.

PR #18131, opened 2024-02-02
zh-cn: synchronize with BCD (v5.5.4 to v5.5.9)

skyclouds2001:
### Description

see https://github.com/mdn/content/issues?q=Synchronize+with+BCD+in%3Atitle


yin1999 (approved):
Thank you @skyclouds2001

PR #18159, opened 2024-02-05
Deleted "sinusoïdale" add replaced "onde" with "forme d'onde"

ManuLintz:
### Description
I adapted the text so it's closer to the English text for 2 reasons:
- The french translation said "sert à définir une onde sinusoïdale". The use of "sinusoïdale" is not correct here because the point of "createPeriodicWave" is to be able to create custom waves (not only sine wave)
- I replaced "onde" with "form d'onde". The English is indeed "waveform" and is more appropriate

### Additional details

See the spec: https://webaudio.github.io/web-audio-api/#PeriodicWave

>[PeriodicWave](https://webaudio.github.io/web-audio-api/#PeriodicWave) represents an **arbitrary periodic waveform** to be used with an [OscillatorNode](https://webaudio.github.io/web-audio-api/#OscillatorNode).

### DIFF
--- a/files/fr/web/api/baseaudiocontext/createperiodicwave/index.md
+++ b/files/fr/web/api/baseaudiocontext/createperiodicwave/index.md
@@ -5,7 +5,7 @@ slug: Web/API/BaseAudioContext/createPeriodicWave
 
 {{ APIRef("Web Audio API") }}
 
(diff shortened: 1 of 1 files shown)


cw118 (approved):
Merci pour cette contribution @ManuLintz !

PR #18215, opened 2024-02-11
[zh-cn]: update the translation of `Element: input event`

xiyuesaves:
Fix non-standard objects in example

### Description

Use `Event.target` to replace `Event.srcElement`

### Motivation

The deprecated `Event.srcElement` is an alias for the `Event.target` property.

### Additional details

[Event: srcElement property](https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement)


yin1999 (approved):
Thank you @xiyuesaves. Welcome to the community :tada:

PR #18243, opened 2024-02-16
Remove missed `{{unimplemented_inline}}` in fr

cw118:
Continuation of #18228 😅 I will merge this one since 18228 was approved and this is the same (type of) change.


cw118:
The checks passed, self-merging (cf. above)

PR #18254, opened 2024-02-17
chore(ci/lint-fix): setup node only once

yin1999:
### Description

remove the duplicated "setup node" step.

### Additional details

Follow the ["check lint" workflow](https://github.com/mdn/translated-content/blob/a623aead87b937fc688ded558f2d17c2665ae93e/.github/workflows/pr-check-lint_content.yml#L70-L76)

### DIFF
--- a/.github/workflows/markdown-lint-fix.yml
+++ b/.github/workflows/markdown-lint-fix.yml
@@ -24,6 +24,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions/checkout@v4
+        with:
+          repository: mdn/content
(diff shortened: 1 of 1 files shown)


.github/workflows/markdown-lint-fix.yml
```diff
@@ -43,13 +44,6 @@ jobs:
           repository: mdn/content
           path: mdn/content
```
LeoMcA, line 45: I think this checkout action needs to be moved above the setup-node action to cache appropriately

LeoMcA (approved):
looks good, thank you!

PR #18282, opened 2024-02-20
es: Fix sourceCommit to Criteria_for_inclusion

Graywolf9:
### Description

Fix sourceCommit to Criteria_for_inclusion


JuanVqz (approved):
Thanks for checking it

PR #18283, opened 2024-02-21
[zh-cn]: update the translation of Array.prototype.sort()

zhou2k:
### Description

中文翻译关于 语法-参数 的部分需要更新
The Chinese translation on Syntax-Parameters needs to be updated

### Additional details

This is English Original Page

> ### Parameters
> 
> - `compareFn` {{optional_inline}}
> 
>   - : A function that defines the sort order. The return value should be a number whose sign indicates the relative order of the two elements: negative if `a` is less than `b`, positive if `a` is greater than `b`, and zero if they are equal. `NaN` is treated as `0`. The function is called with the following arguments:

[Array.prototype.sort() - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)


JasonLamv-t (approved):
LGTM.
Other full-text changes about Array sections will be updated in next step, in my plan.

PR #18287, opened 2024-02-22
zh-cn: remove the `none` value of the "min-hight" attribute

skyming97:
The attribute “min-height” has no attribute value “done”

### Description
“min-height” has no attribute value “done” according to mdn documentation

### Motivation

In the spirit of dedication and open source, avoid misleading readers.

### Additional details

image


yin1999 (approved):
Thank you @skyming97

PR #18298, opened 2024-02-22
[fr] add link to Pulsar editor

jnupjun:


leon-win:
I think it would be more correct to make such a change in the main repository (https://github.com/mdn/content) and then apply these changes to all translations

cw118:
Sorry for the delay @jnupjun and thank you firstly for the PR.

French content should be translated from the original English content. It looks like the en-US page only mentions VS Code now, but the French version is quite out of date so I would be open to keeping your proposed changes until someone is able to fully update the French version (I currently do not have time :/).

cc/ @mdn/yari-content-fr

SphinxKnight:
@jnupjun @cw118 I apologize for the (real) late comment here.
Agreeing with @cw118 regarding the need of an update on the page, I went with https://github.com/mdn/translated-content/pull/19237 and included Pulsar in it (though it doesn't exist on the English side, I'm not against more diversity :))

SphinxKnight:
let's get this one merged, we still can follow-up on the other one afterwards.

Thanks @jnupjun

PR #18351, opened 2024-02-29
zh-cn: update the translation of margin-right

yin1999:
### Description

update the translation of margin-right


JasonLamv-t (approved):
LGTM. Thank you! @yin1999

PR #18380, opened 2024-03-03
[zh-cn]: fix a typo, and modify the translation of words in codes

leverimmy:
### Description

First of all, there's a minor typo at the end of the file, so I added a quotation mark.

Secondly, I realized that it is not suggested that we translate the "Hello Chris: welcome to our company" program because right after the code there's a hyperlink to the source code in GitHub - which is English, and if we translate the program into Chinese, it would not correspond.

I didn't open an issue.


yin1999 (approved):
Thank you @LeverImmy. At this time we have not determined how the github examples should be localized, so we still need to align such code examples with the referenced "non-localized" GitHub examples.

PR #18394, opened 2024-03-03
chore(zh-cn): unify 'see also' translation

jasonren0403:
### Description

continue cleanup for 'see also' header translation


JasonLamv-t (approved):
LGTM.

PR #18405, opened 2024-03-05
zh-cn: update the translation of `Event.timeStamp`

yin1999:
### Description

update the translation of `Event.timeStamp`


JasonLamv-t (approved):
LGTM. Thanks @yin1999 .

PR #18428, opened 2024-03-06
[fr] Updating outdated JSON Global Object links

deltoro05:


cw118 (approved):
Merci beaucoup pour ces corrections @deltoro05 :)

PR #18430, opened 2024-03-06
zh-cn: fix the broken link in Regular Expressions guide

deltoro05:


yin1999 (approved):
Thanks for the cleanup. Although the English document has removed this link, it may take a long time for us to synchronize the corresponding changes. Let us deal with the redirect first!

PR #18437, opened 2024-03-06
[zh-cn] update `Web/JavaScript/Reference/Global_Objects/JSON`

deltoro05:
This URL https://unicode-table.com/ has changed to https://symbl.cc/, it will be cool if you fix it 😊

PR #18440, opened 2024-03-06
Update index.md

deltoro05:


jasonren0403:
duplicated with #18437

PR #18595, opened 2024-03-11
[zh-TW]: translate  `web/html/element/b`

Dr-XYZ:


yin1999 (commented):
Hey @Dr-XYZ. Could you resolve the issue of EOL characters?

I've mentioned this problem here: https://github.com/mdn/translated-content/pull/18120#pullrequestreview-1897309154

image

PR #18758, opened 2024-03-18
[zh-cn] init translation of HTML attribute capture

jasonren0403:


yin1999 (approved):
LGTM. Thank you @jasonren0403 :>

PR #18762, opened 2024-03-18
Replaced deprecated method "addListener" to "addEventListener"

fr1g:
### Description

Replaced deprecated method "addListener" to "addEventListener"

### Motivation

Found a warning from Rider said that addListener was deprecated and found this while reading the docs about difference between "addListener" and "addEventListener"

### Additional details

@*nothing*@

@*nothing*@


fr1g:
warn: this translated article seems is way behind the origin content.

yin1999 (approved):
Thank you @fr1g

yin1999:
> warn: this translated article seems is way behind the origin content.

Yes, we need to synchronize document content

PR #18771, opened 2024-03-19
[zh-cn] Sync with en-us 'column-count'

T34-active:
### Description

[zh-CN column-count](https://developer.mozilla.org/zh-CN/docs/Web/CSS/column-count) Not sync with en-US markdown [en-US column-count](https://developer.mozilla.org/en-US/docs/Web/CSS/column-count)

### Motivation

Maintain consistency with the en-US document version

### Additional details

[zh-CN column-count](https://developer.mozilla.org/zh-CN/docs/Web/CSS/column-count)
[en-US column-count](https://developer.mozilla.org/en-US/docs/Web/CSS/column-count)


T34-active:
push wrong branch sry :(

PR #18776, opened 2024-03-19
[zh-cn] Update description of `WeakMap` key

xiaoxiyao:
### Description

"primitive type" should be "原始类型" instead of "类型"


yin1999 (approved):
Thank you @xiaoxiyao! Welcome to the community :tada:

PR #18797, opened 2024-03-20
[fr] fix: Replace "schemas" by "schemes"

regseb:
### Description

Replace `schemas` by `schemes` in french [`UrlFilter`](https://developer.mozilla.org/fr/docs/Mozilla/Add-ons/WebExtensions/API/events/UrlFilter#schemas) page.

### Motivation

The right value is `schemes` (english [`UrlFilter`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/events/UrlFilter#schemes) page).


SphinxKnight (approved):
Merci pour cette correction @regseb !

PR #18823, opened 2024-03-21
[fr] Fix issues where EmbedLiveSample doesn't work properly

cyrilf:
### Description

I'm fixing the usage of `EmbedLiveSample` in two pages here, as a first validation step.
To go further, maybe the macro can be improved, or we need to go through all files and fix the broken ones.

### Motivation

The macro `EmbedLiveSample` isn't working properly in some files.
Because it's not using the required [`sample_id`](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Live_samples#sample_id) parameter, it results on some broken demos.

### Additional details

You can see the [demos on the French gap page](https://developer.mozilla.org/fr/docs/Web/CSS/gap) have broken demos (only the HTML is rendered, not picking up the CSS).
But the [English demos](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) are working as expected.

| French | English |
-|-
| image | image |

Same for the [list-style-type](https://developer.mozilla.org/fr/docs/Web/CSS/list-style-type#exemples).

The difference is that the English version does:
`{{EmbedLiveSample("Grid_layout", "auto", 250)}}` on [this line](https://github.com/mdn/content/blob/main/files/en-us/web/css/gap/index.md?plain=1#L153)
The French version does:
`{{EmbedLiveSample("", "auto", "230px")}}` on [this line](https://github.com/mdn/translated-content/blob/main/files/fr/web/css/gap/index.md?plain=1#L150)

It's not providing a `sample_id` and because there is some various heading levels, maybe the macro automatic matching gets lost..

----

~:warning: I haven't tested this locally, I only commited via the Github Codespace interface as a proof of concept of what can be done to solve this.~
The auto-deploy shows that this fix is working properly.

### DIFF
--- a/files/fr/web/css/gap/index.md
+++ b/files/fr/web/css/gap/index.md
@@ -103,7 +103,7 @@ Cette propriété est définie avec une valeur `<'row-gap'>`, éventuellement su
 
 #### Résultat
 
-{{EmbedLiveSample("", "auto", "230px")}}
+{{EmbedLiveSample("Disposition flexible", "auto", "230px")}}
 
 ### Disposition en grille
 
@@ -147,7 +147,7 @@ Cette propriété est définie avec une valeur `<'row-gap'>`, éventuellement su
 
 #### Résultat
 
-{{EmbedLiveSample("", "auto", "230px")}}
+{{EmbedLiveSample("Disposition en grille", "auto", "230px")}}
 
 ### Disposition multi-colonnes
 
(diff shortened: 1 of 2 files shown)


SphinxKnight:
Thanks a lot @cyrilf, both for the PR itself and the detailed explanation helping for the review
EmbedLiveSample and l10n have a "long story": https://github.com/mdn/yari/issues/4971#issuecomment-1315780006 / https://github.com/mdn/yari/issues/5016

SphinxKnight (approved):

PR #18826, opened 2024-03-22
zh-cn: update the translation of `border-right-width`

yin1999:
### Description

update the translation of `border-right-width`


jasonren0403 (approved):
👍 Thank you @yin1999

PR #18877, opened 2024-03-23
tweak french SEO page

v-gb:
See https://www.wordreference.com/enfr/copywriter for the meaning of copywriting .

### Description

Fixing a misunderstanding about the meaning of "copywriting".

### Motivation

The impacted sentence didn't make sense before.


tristantheb (approved):
Good first contribution 👍🏻

Issue #18191, opened 2024-02-07
[zh-cn]: add & sync translation for Battery Status API


skyclouds2001:
```[tasklist]
### Battery Status API translation statuses
- [x] [Battery Status API](https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API) summary, in #18232
- [x] [`Navigator.getBattery()`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getBattery) method, in #18233
- [x] [`BatteryManager`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager) interface, in #18770
- [x] [`BatteryManager.charging`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/charging) property, in #18928
- [x] [`BatteryManager.chargingTime`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/chargingTime) property, in #18929
- [x] [`BatteryManager.dischargingTime`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/dischargingTime) property, in #18930
- [x] [`BatteryManager.level`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/level) property, in #18931
- [x] [`BatteryManager.chargingchange`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/chargingchange_event) event, in #18967
- [x] [`BatteryManager.chargingtimechange`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/chargingtimechange_event) event, in #18968
- [x] [`BatteryManager.dischargingtimechange`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/dischargingtimechange_event) event, in #18969
- [x] [`BatteryManager.levelchange`](https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/levelchange_event) event, in #18970
```

battery - 电池

relates to https://github.com/mdn/content/pull/32218

PR #18928, opened 2024-03-24
[zh-cn]: sync translation for `BatteryManager.charging`

skyclouds2001:

part of the https://github.com/mdn/translated-content/issues/18191

depends on https://github.com/mdn/content/pull/32218

see also https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager/charging


yin1999 (approved):
LGTM. Thanks @skyclouds2001

PR #19021, opened 2024-03-28
chore(zh-cn): translate for promise.finally, rewrite finally()'s "Return value" section

fuchunhui:
### Description
1. rewrite finally()'s "Return value" section.


JasonLamv-t (approved):
Well done! @fuchunhui

Issue #10487, opened 2022-12-05
[Meta] Replace @api/deki references


caugner:
## Background

A legacy way of referencing images is to use the magic `/@api/deki` prefix, which is essentially a shortcut for `https://mdn.mozillademos.org`, and this has been **deprecated** similar to using the mozillademos domain itself (https://github.com/mdn/translated-content/issues/10400).

## Problem

There are still 750 occurrences on 606 page of `@api/deki` in mdn/translated-content that need to be replaced.

## Scope

- [x] es (139 pages) #11385
- [x] fr (4 pages) - https://github.com/mdn/translated-content/pull/10727
- [x] ja (16 pages) - https://github.com/mdn/translated-content/pull/10682
- [x] ko (67 pages) https://github.com/mdn/translated-content/pull/10507
- [x] pt-BR (69 pages) #16510
- [x] ru (79 pages) #17260
- [x] zh-CN (171 pages) #10854 #10875 #10907 #10936 #10952 #10953
- [x] zh-TW (61 pages) #10854 #10875

SphinxKnight:
@caugner I'm getting 1 page with 4 occurrences in `fr`. I think the content might have changed/deleted since the creation of this. If you have a couple of spare minutes, I'd appreciate a check for the counts :) (PR incoming for `fr` though : ))

yin1999:
Done for l10n-zh

hochan222:
Done for `l10n-ko`.

Graywolf9:
Done for `es`, Thank you

leon-win:
Done for `ru`

PR #19063, opened 2024-03-29
chore(pt-br): remove the last reference to `@api/deki`

yin1999:
### Description

remove the last reference to `@api/deki`. The replaced example is from [the en-US document](https://github.com/mdn/content/blob/71251662105fbac882aca295c641500175bb9ef1/files/en-us/web/html/element/audio/index.md?plain=1#L317-L320).

I can't speak Portuguese (Brazil), so please check if the translation is correct.

Part of: #10487


josielrocha (approved):
LGTM ;)

PR #19187, opened 2024-04-03
Update fr page for local-link to fit with en-us

ghost:
All is in the title ^^.


SphinxKnight (approved):
Merci @Arthur-GYT :bow:

PR #19233, opened 2024-04-05
zh-cn: fix the markdown rendering error in `function*` statement

kiner-tang:
### Description

Before:

image

After:

image


yin1999 (approved):
Good catch. Thank you @kiner-tang

PR #19347, opened 2024-04-09
[zh-cn] remove scrollbar-color non-standard syntax

sechi747:
### Description

[zh-cn] remove scrollbar-color non-standard syntax
### Motivation

Synchronize with en document, `scrollbar-color` can't set with `dark` or `light` .
Link draft: https://drafts.csswg.org/css-scrollbars/#scrollbar-color


yin1999 (approved):
:+1:

PR #19366, opened 2024-04-10
zh-cn: correct the sourceCommit sha of term "Accessibility"

yin1999:
### Description

correct the sourceCommit sha of term "Accessibility"


yin1999 (commented):
Simple fix, self-approval

PR #19410, opened 2024-04-11
[fr] Follow-up of #18401- localization of SVGs in French

SphinxKnight:
Adding a French version of the diagrams.
AFAICT the rendering looks ok from here (Fx + Chromium)


Mehrm0:
merci d'avoir fusionné la branche.

cw118 (approved):
🙏

PR #19493, opened 2024-04-14
zh-cn: update the translation of `418` status code

yin1999:
### Description

update the translation of `418` status code


jasonren0403 (approved):
👍

PR #19532, opened 2024-04-15
[zh-cn]: empty element replace to void element

T34-active:


yin1999 (approved):
:+1:

T34-active:
Thank you @yin1999

PR #19586, opened 2024-04-18
[zh-cn]: improve the translation of the description of the typed data usage

kiner-tang:


yin1999 (approved):
Good catch, thanks @kiner-tang

PR #19591, opened 2024-04-18
[zh-cn]: update translation of `Comment()` constructor

T34-active:


yin1999 (commented):
One suggestion:

yin1999 (approved):
Thanks

PR #19620, opened 2024-04-20
[fr] remove odd parameters from `Web/API/EventTarget/*`

leon-win:
### Description

This PR removes odd parameters in `fr` locale:
* `mozSystemGroup` from `Web/API/EventTarget/addEventListener` and `Web/API/EventTarget/removeEventListener`;
* `passive` from `Web/API/EventTarget/removeEventListener`.

Relates to https://github.com/mdn/content/pull/9979, #19427


SphinxKnight (approved):
Thanks @leon-win for this :bow: always really appreciate going for the extra-mile for other locales.

I'm wondering why the en-US page still has `passive` in the examples for `removeEventListener()` but that's not a blocker for this one :)

leon-win:
> Thanks @leon-win for this 🙇 always really appreciate going for the extra-mile for other locales.
> 
> I'm wondering why the en-US page still has `passive` in the examples for `removeEventListener()` but that's not a blocker for this one :)

I always try to help if possible :hugs: 

Regarding the use of the `passive` parameter in the examples for `removeEventListener()`, I agree that they are rather confusing. But on the other hand, I understand why such examples are given: to show that it is (and only) the `capture` flag that is influences.

It might be more accurate to point out that the possible options for `addEventListener()` and `removeEventListener()` are different:
https://dom.spec.whatwg.org/#interface-eventtarget
image

Perhaps it makes sense to propose such an idea to the `mdn/content`)

PR #19639, opened 2024-04-22
[zh-cn]: update translation of `<center>` element

T34-active:


yin1999 (approved):
Thank you @T34-active

PR #19930, opened 2024-05-01
ja: There is no need to describe the scoped attribute of the style element.

juner:
style要素の scoped属性がついているときの コンテンツカテゴリー の説明は不要です。

### Description

There is no need to describe the scoped attribute of the style element.
Removed description of behavior when scoped attribute is attached to content category item.

### Motivation

The scoped attribute is no longer recommended, and there is no explanation as an attribute of the style element.

### Additional details

https://html.spec.whatwg.org/multipage/semantics.html#the-style-element
※ not found scoped attribute description.


mfuji09 (approved):
OKです。ありがとうございます。

PR #19966, opened 2024-05-02
[zh-cn]: "无" replace "否" patch 2

T34-active:


yin1999:
见 https://github.com/mdn/translated-content/pull/19964#issuecomment-2089900608

PR #20107, opened 2024-05-07
fr: Glossary/Certificate_authority

PassionPenguin:
feat: update translation

this file contains more updates (sync) needed than upstream pr 33413, thus is opened separately

related pr:

- #20108 files from abstration to cms (except canvas)

### DIFF
--- a/files/fr/glossary/certificate_authority/index.md
+++ b/files/fr/glossary/certificate_authority/index.md
@@ -1,13 +1,15 @@
 ---
 title: Autorité de certification
 slug: Glossary/Certificate_authority
+l10n:
+  sourceCommit: 7a551aaa034fbada3eb99e6fc924a0313b78307f
 ---
 
 {{GlossarySidebar}}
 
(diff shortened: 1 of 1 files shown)


SphinxKnight (changes requested):
Even though I don't have much to say regarding the text modified here, referencing the `sourceCommit` metadata is a bit blocking here since the (updated) content does not match the English content for this commit.

Hence https://github.com/PassionPenguin/translated-content/pull/2 to seize the occasion since the page is short.

PassionPenguin:
> Even though I don't have much to say regarding the text modified here, referencing the `sourceCommit` metadata is a bit blocking here since the (updated) content does not match the English content for this commit.
> 
> 
> 
> Hence https://github.com/PassionPenguin/translated-content/pull/2 to seize the occasion since the page is short.

oops, must have left some previous updates! Sorry about that!

Merci beaucoup!

SphinxKnight (approved):

Issue #20165, opened 2024-05-08
[zh-cn]: add & sync transaltion for File API


skyclouds2001:
```[tasklist]
### File API
- [ ] #21119 
- [ ] #21056 
- [ ] #21057
- [ ] #21058 
- [ ] #20433 
- [ ] #20418 
- [ ] #20356 
- [ ] #20358 
- [ ] #20357 
- [ ] #20324 
- [ ] #20335 
- [ ] #20359 
- [ ] #20325 
- [ ] #20392
- [ ] #20170
- [ ] #20270 
- [ ] #20272 
- [ ] #20271 
- [ ] #20273 
- [ ] #20274 
- [ ] #20673 
- [ ] #20676 
```

PR #20170, opened 2024-05-08
[zh-cn]: sync translation for `FileReaderSync`

skyclouds2001:
see https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync


jasonren0403 (approved):
LGTM 👍

PR #20178, opened 2024-05-09
[zh-cn]: update the translation of Document `close()` method

T34-active:


JasonLamv-t (approved):
Well done! Thank you! @T34-active

PR #20200, opened 2024-05-09
l10n-zh: turn the lower case "h" for "GitHub" to upper case

crrashh1542:
### Description
This PR fixes the wrongly appeared lower-case letter "h" in the word "GitHub", which was supposed to be the upper-case one, in the Simplified Chinese and Traditional Chinese translations.

### Motivation
I accidentally found the word with lower-case letter "h" was wrongly presented in the page [MDN Web 文档的本地化](https://developer.mozilla.org/zh-CN/docs/MDN/Community/Contributing/Translated_content). However, according to the [English original text](https://developer.mozilla.org/en-US/docs/MDN/Community/Contributing/Translated_content) and the spelling in GitHub official website, it should be spelled in upper case.


JasonLamv-t (approved):
LGTM. Thank you! @crrashh1542 ❤️

PR #20280, opened 2024-05-12
[ja] correction of mistranslation for non-async

ytakio:
- non-`async`
  - x : 非同期関数
  - o : async ではない関数

### Description

In Current Japanese document, that non-`async` function is translated to 非同期関数 (asynchronous function) in Japanese.
I've corrected it with `async` を使わない関数 which means that a function doesn't use `asynx` in Japanese.

### Motivation

It just mistranslation to opposition mean.


mfuji09 (approved):
主題以外の用語の変更を元に戻しました。

PR #20331, opened 2024-05-13
[zh-cn] update the translation of `<colgroup>` element

jasonren0403:


files/zh-cn/web/html/element/colgroup/index.md
```diff
@@ -1,112 +1,215 @@
...
+    <th>下午</th>
+    <td>瑜伽</td>
+    <td>棋类俱乐部</td>
+    <td>见朋友</td>
+    <td>体操</td>
+    <td>生日派对</td>
+    <td>钓鱼之旅</td>
+    <td>自由时间</td>
+  </tr>
+</table>
```
JasonLamv-t, line 123: I think the change you made, that is, adding the `th` in the `tr` elemment, is better. Going to update the upstream?

files/zh-cn/web/html/element/colgroup/index.md
```diff
@@ -1,112 +1,215 @@
...
+```
 
-  - : This attribute contains a positive integer indicating the number of consecutive columns the `<colgroup>` element spans. If not present, its default value is `1`.
+### CSS
 
-    > **备注：** This attribute is applied on the attributes of the column group, it has no effect on the CSS styling rules associated with it or, even more, to the cells of the column's members of the group. The `span` attribute is not permitted if there are one or more `<col>` elements within the `<colgroup>`.
+分组列可以使用 CSS 在视觉上突出显示结构：
```
JasonLamv-t, line 128:
```suggestion
分组列可用于使用 CSS 直观地突出显示结构：
```

JasonLamv-t (approved):
LGTM. Thx! @jasonren0403

files/zh-cn/web/html/element/colgroup/index.md
jasonren0403: oops, I've copied them from `<col>` examples...

I would update them to match the original document.

JasonLamv-t (approved):
Hah, LGTM now. Thx @jasonren0403.

PR #20358, opened 2024-05-13
[zh-cn]: add translation for `FileList.length`

skyclouds2001:
### Description

see https://developer.mozilla.org/en-US/docs/Web/API/FileList/length


yin1999 (approved):
:+1:

PR #20377, opened 2024-05-14
[zh-cn]: update the translation of window `dump()` method

T34-active:


JasonLamv-t (approved):
LGTM. Thanks. :+1: @T34-active

PR #20437, opened 2024-05-17
[zh-cn]: update the translation of URL `toString()` method

T34-active:
### Description
https://developer.mozilla.org/en-US/docs/Web/API/URL/toString


yin1999 (approved):
Thank you @T34-active

PR #20654, opened 2024-05-19
zh-cn: update the syntax format of `Reflect.set()`

xiaoAn19:
correct spelling mistakes

### DIFF
--- a/files/zh-cn/web/javascript/reference/global_objects/reflect/set/index.md
+++ b/files/zh-cn/web/javascript/reference/global_objects/reflect/set/index.md
@@ -10,7 +10,7 @@ slug: Web/JavaScript/Reference/Global_Objects/Reflect/set
 ## 语法
 
 ```plain
-Reflect.set(target, propertyKey, value[, receiver])
+Reflect.set(target, propertyKey, value, receiver)
 ```
 
 ### 参数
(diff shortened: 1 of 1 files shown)


files/zh-cn/web/javascript/reference/global_objects/reflect/set/index.md
```diff
@@ -10,7 +10,7 @@ slug: Web/JavaScript/Reference/Global_Objects/Reflect/set
 ## 语法
 
 ```plain
-Reflect.set(target, propertyKey, value[, receiver])
+Reflect.set(target, propertyKey, value, receiver)
 ```
```
yin1999, line 14:
````suggestion
```js-nolint
Reflect.set(target, propertyKey, value)
Reflect.set(target, propertyKey, value, receiver)
```
````

yin1999 (commented):
Hey @xiaoAn19. Actually, the original format is correct, the brackets here mean that this parameter is optional.

But it's still worth to update the format, as it's not the same as the [format in the en-US doc](https://github.com/mdn/content/blob/b9c5849612a5beed7f2b8ab5e3c0a336bd585674/files/en-us/web/javascript/reference/global_objects/reflect/set/index.md?plain=1#L16-L19)

files/zh-cn/web/javascript/reference/global_objects/reflect/set/index.md
xiaoAn19: Thanks for explaining it to me

yin1999 (approved):
Thank you @xiaoAn19. Welcome to the community :tada:

PR #20728, opened 2024-05-22
learn/javascript/objects/basics Update index.md

juampibruno:
This pull request makes the following changes:

1. **Replace `var` with `let`**:
   - Replaced all instances of `var` with `let` to improve code consistency and follow modern JavaScript best practices.

2. **Fix template literal in the `alert` function**:
   - The original code had an error in the template literal, causing a variable to not display correctly.


Graywolf9 (changes requested):
Thank you so much for your help and contribution!

We use the English version as the main line of the content and they are using `const` in this examples, so we need to use it too.

English: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/objects/basics

Thank you so much again, tell me if you have any doubts

Graywolf9 (approved):

Issue #20746, opened 2024-05-24
[zh-cn]: proofread and update `web/mathml`


tianyeeT:
### MDN URL

https://developer.mozilla.org/zh-CN/docs/Web/MathML

### What information was incorrect, unhelpful, or incomplete?

Things need to be updated: the **task** below

### What did you expect to see?

Update the existing and add the missing.

```[tasklist]
### Tasks
- [ ] https://github.com/mdn/translated-content/pull/20823
- [ ] https://github.com/mdn/translated-content/pull/20825
- [x] element/maction
- [ ] https://github.com/mdn/translated-content/pull/20841
- [ ] https://github.com/mdn/translated-content/pull/20842
- [ ] https://github.com/mdn/translated-content/pull/20843
- [ ] https://github.com/mdn/translated-content/pull/20844
- [ ] https://github.com/mdn/translated-content/pull/20845
- [ ] https://github.com/mdn/translated-content/pull/20846
- [ ] https://github.com/mdn/translated-content/pull/20847
- [ ] https://github.com/mdn/translated-content/pull/20848
- [ ] https://github.com/mdn/translated-content/pull/20850
- [ ] https://github.com/mdn/translated-content/pull/20851
- [ ] https://github.com/mdn/translated-content/pull/20852
- [ ] https://github.com/mdn/translated-content/pull/20853
- [ ] https://github.com/mdn/translated-content/pull/20854
- [ ] https://github.com/mdn/translated-content/pull/20855
- [ ] https://github.com/mdn/translated-content/pull/20856
- [x] element/mspace
- [ ] https://github.com/mdn/translated-content/pull/20857
- [ ] https://github.com/mdn/translated-content/pull/20858
- [ ] https://github.com/mdn/translated-content/pull/20859
- [ ] https://github.com/mdn/translated-content/pull/20860
- [ ] https://github.com/mdn/translated-content/pull/20861
- [ ] https://github.com/mdn/translated-content/pull/20862
- [ ] https://github.com/mdn/translated-content/pull/20863
- [ ] https://github.com/mdn/translated-content/pull/20864
- [ ] https://github.com/mdn/translated-content/pull/20865
- [ ] https://github.com/mdn/translated-content/pull/20866
- [ ] https://github.com/mdn/translated-content/pull/20867
- [ ] https://github.com/mdn/translated-content/pull/20868
- [ ] https://github.com/mdn/translated-content/pull/20870
- [ ] https://github.com/mdn/translated-content/pull/20871
[...]

PR #20841, opened 2024-05-26
[zh-cn]: update the translation of `element/math`

tianyeeT:
- add l10n
- checked with the latest upstream files.

#20746


yin1999 (commented):
没必要专门加这个了，后面有内容更新了再加吧

PR #20949, opened 2024-05-29
docs(zh-cn): add instructions on how to add images

yin1999:
### Description

add instructions on how to add images


yin1999:
As we are moving assets into [mdn/shared-assets](https://github.com/mdn/shared-assets), the instructions on how to add images are not necessary.

PR #21057, opened 2024-05-31
[zh-cn]: sync translation for `Blob.{size,type,slice()}`

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/API/Blob/size
https://developer.mozilla.org/en-US/docs/Web/API/Blob/type
https://developer.mozilla.org/en-US/docs/Web/API/Blob/slice


yin1999 (approved):
:+1:

PR #21058, opened 2024-05-31
[zh-cn]: sync translation for `Blob.{arraybuffer(),stream(),text()}`

skyclouds2001:
### Description

[Blob.arrayBuffer()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer)
[Blob.stream()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream)
[Blob.text()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/text)


files/zh-cn/web/api/blob/arraybuffer/index.md
```diff
@@ -27,7 +29,7 @@ arrayBuffer()
```
jasonren0403, line 29: 28行，中间那句可以优化翻译如下

“例如，如果用于获取 blob 数据的 reader 抛出异常，就会发生这种情况。”

jasonren0403 (commented):
只有一处小问题

jasonren0403 (approved):
👍 感谢

PR #21176, opened 2024-06-02
[zh-cn]: create the translation of `bookmarks.onImportEnded` event

tianyeeT:


jasonren0403 (approved):
LGTM, sorry for late review

PR #21361, opened 2024-06-05
[fr] Import sidebars from en-US

caugner:
### Description

Adds sidebar macro calls from the en-US version.

### Motivation

Sidebars have been added to almost all en-US pages, and many translated pages don't have them yet.

### Additional details

This was created using a custom version of the sync-translated-content tool (https://github.com/mdn/yari/commit/86ba2eac8690f3b8c1bca3cf5ad1aa813be4285c).


cw118:
Thank you for this @caugner ! I saw some XSLT files have both the XsltRef() and XsltSidebar macros in this PR, but I checked en-US and it seems the English files only have XsltSidebar. Is keeping both macros intentional here?

caugner:
> I saw some XSLT files have both the XsltRef() and XsltSidebar macros in this PR, but I checked en-US and it seems the English files only have XsltSidebar. Is keeping both macros intentional here?

Good question!

Technically, `XsltRef` is not a sidebar although it may serve a similar navigation purpose.

And actually https://developer.mozilla.org/en-US/docs/Web/EXSLT/set and https://developer.mozilla.org/fr/docs/Web/XSLT both already use `XsltRef` together with a sidebar, so I wouldn't think this is a blocker. However, it might make sense to remove the `XsltRef` macro from mdn/content, then deprecate it, and remove it from mdn/translated-content eventually.

cw118 (approved):
I see, thank you so much for the detailed explanation! This looks great 🙏

PR #21366, opened 2024-06-05
[zh] Import sidebars from en-US

caugner:
### Description

Adds sidebar macro calls from the en-US version.

### Motivation

Sidebars have been added to almost all en-US pages, and many translated pages don't have them yet.

### Additional details

This was created using a custom version of the sync-translated-content tool (https://github.com/mdn/yari/commit/86ba2eac8690f3b8c1bca3cf5ad1aa813be4285c).

PR #21370, opened 2024-06-06
[fr] Import sidebars from en-US (fixup)

caugner:
### Description

Fixes sidebar macro calls that were imported from the en-US version.

### Motivation

The `QuicklinksWithSubPages` macro requires a path including the current locale (see [here](https://github.com/mdn/translated-content/pull/21366#discussion_r1628618027)).

Follow-up of https://github.com/mdn/translated-content/pull/21361.


caugner:
@cw118 This fixes a small issue with my previous PR. Since you reviewed that one, would you like to take a look?

cw118 (approved):
Looks good, thank you for fixing this! Not sure how I missed these last time 😅

PR #21443, opened 2024-06-10
Fix incorrect translation

prahal:
Lossy was translated by the French for lossless (JPEG).

### Description

For JPEG, lossy ("Avec perte") was translated with the French word for lossless ("Sans perte")

### Motivation

Even though the rest of the document tells JPEG is lossy this might confuse newbies.


cw118 (approved):
Merci pour cette correction @prahal !

PR #21763, opened 2024-06-20
fix(zh-cn): the statement "pushState is async" is not correct

liuwenzhuang:


yin1999 (approved):
Thank you @liuwenzhuang

PR #21766, opened 2024-06-20
[fr] fix broken link to W3C in banner_role page

floriannari:
### Description
Update link from https://www.w3.org/TR/wai-aria/roles#banner to https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/banner.html

### Motivation
https://www.w3.org/TR/wai-aria/roles#banner is broken


cw118 (approved):
Thank you for this correction @floriannari !

PR #21924, opened 2024-06-25
zh-cn: fix typo in `toLocaleUpperCase()`

paddyW1120:
“The toLocaleUpperCase() method returns the value of the string converted to upper case according to any locale-specific case mappings” should be translated as "toLocaleUpperCase（）方法返回根据任何特定于区域设置的大小写映射转换为大写的字符串的值".


yin1999 (approved):
Good catch @paddyW1120. Thank you!

yin1999:
Welcome to the community 🎉

PR #21947, opened 2024-06-26
Update index.md

matiasluduena23:
### Description

I change a grammar error in the line 15.
Change the word 'caneda' for 'cadena'.

### Motivation

The word 'caneda' don't exist and could be confuse for some reader. The correct word is 'cadena', it means string


Graywolf9 (approved):
Thank you so much for your time!

Welcome to MDN!

PR #21977, opened 2024-06-27
fix(env): update required node engine to v18.18

OnkarRuikar:
- porting from yari PR https://github.com/mdn/yari/pull/11373


.nvmrc
```diff
@@ -1 +1 @@
-v18
+v18.18
```
OnkarRuikar, line 1:
```suggestion
v18
```

argl (approved):

yin1999:
Hey @OnkarRuikar. I think the problem metioned in mdn/yari#11373 would not affect this repo. I've made a test with `node v18.0.0`, and I found that the min version required by dependencies in this repo could be `18.12.0`:

image

And I can install the dependencies when the node version set to `v18.12.0`:

image

---

Should we set it to `>=18.12.0` instead?

package.json
```diff
@@ -24,7 +24,7 @@
     "prepare": "husky || true"
   },
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=18.18.0"
```
OnkarRuikar, line 27: Unlike Content, this repo doesn't have a dependency on Yari. So we can use the minimum required `18.12.0`:
```suggestion
    "node": ">=18.12.0"
```

yin1999 (approved):
Thank you @OnkarRuikar

PR #21981, opened 2024-06-27
fix-l10n(fr): misleading sentence

hormesiel:
### Description

The sentence in french was misleading and incorrect. I had to switch to the english version to make sure I get it right.


cw118 (approved):
Looks great, thank you for this contribution @flawyte !

PR #22155, opened 2024-07-04
Improve translation for ja in JS index

okuramasafumi:
### Description

Improve translation for Japanese in JS index page.

### Motivation

I noticed that the translation is partially wrong.


mfuji09 (approved):
OKです。

PR #22382, opened 2024-07-12
zh-CN: correct the presence of response body of HTTP TRACE method

ghost:
### Description
correct presence of response body of TRACE method

### Motivation
N/A

### Additional details
N/A

N/A


yin1999 (approved):
Good catch. Thank you @yuqqs.

PR #22530, opened 2024-07-22
[ko] Math.atanh() 신규 번역 외

wisedog:
- [Math.atanh()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atanh) 신규 번역
- [Math.SQRT2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/SQRT2) 최신화


hochan222 (approved):
LGTM :)

PR #22532, opened 2024-07-23
zh-cn: fix content format

Yukiniro:
https://developer.mozilla.org/zh-CN/docs/Web/API/BaseAudioContext/createBuffer


yin1999 (approved):
Thank you @Yukiniro

PR #22553, opened 2024-07-23
zh-cn: update Markdown in MDN docs to use GFM syntax

yin1999:
### Description

update Markdown in MDN docs to use GFM syntax

upstream: mdn/content#33384


jasonren0403 (approved):
👍

PR #22554, opened 2024-07-24
[ja] Fix mistranslation in performance.now()

magicant:
### Description

Part of the Japanese translation for the "Security requirements" section in the "Performance: now() method" page seems to be incorrect. The original English text says "microseconds", but the Japanese version says "ミリ秒" (milliseconds). This pull request tries to fix the mistranslation.


mfuji09 (approved):
OKです。
なお、日本語訳のレビュアーは全員が日本人ですので、プルリクエストは日本語でどうぞ。

PR #22650, opened 2024-07-27
zh-tw: convert noteblocks in `/glossary` to GFM syntax

yin1999:
### Description

convert noteblocks in `/glossary` to GFM syntax


t7yang:
LGFM

PR #22651, opened 2024-07-27
zh-tw: convert noteblocks in `/learn` to GFM syntax

yin1999:
### Description

convert noteblocks in `/learn` to GFM syntax


t7yang:
LGFM
CC @yin1999 @irvin

PR #20108, opened 2024-05-07
fr: update glossary

PassionPenguin:
sync with mdn/content#33413

file changed: from abstration to cms (except canvas)

related pr:

- #20106 `Blink` the moteur de rendu / blink
- #20107 `Autorité de certification` / Certificate authority

PR #22954, opened 2024-08-06
fr: update Glossary/ATAG

PassionPenguin:
part of #20108


tristantheb (approved):
Thanks for this addition 👍🏻

PR #22976, opened 2024-08-07
fix(zh-cn): correct the possible values for `justify-content` and other properties

webkong:
fix mistake

### Description

fix misktake for justify-content

### Motivation

It should be justify-content but written into aligin-content


yin1999 (approved):
Thanks @webkong. I've pushed a commit to update the possible values for those properties.

PR #23004, opened 2024-08-08
fix(zh-cn): correct the closing tag

scriptLin-bjtu:
Translation error,<script>should be added before</body>instead of</head>

### Description
Translation error,<script>should be added before<body/>instead of</head>

### Motivation
Prevent ambiguity


yin1999 (approved):
Thanks

yin1999:
Welcome aboard @scriptLin-bjtu :tada:

PR #23005, opened 2024-08-08
ru: Convert noteblocks to GFM alerts (part 1)

queengooborg:
This PR converts the noteblocks for the Russian locale to GFM Alerts syntax, using a [conversion script](https://github.com/queengooborg/mdn-toolkit/blob/main/upgrade-noteblock.js). This is part 1.

Note: manual adjustments have also been made to correct some issues, including capitalization, syntax, duplicated keywords and more.


leon-win (approved):

leon-win:
Hi, @queengooborg !
Thanks for your help!

I have supplemented this request, corrected several files.

I have noticed two interesting points:
1. In some places the syntax was not updated (i.e. the old format remained). It turns out that you made the changes manually, and not with your script?
2. Autolinter in some cases expects an incorrect format, wraps to one line `> [!NOTE]` and the next line. This happens if the message begins with a macro call (for example `{{domxref()}}`) or the symbol "`". There may be other cases.

queengooborg:
Hey @leon-win!  Since you had used the script, I had made an update to ensure that it doesn't convert any noteblocks that have a special character in front, as those cannot yet be converted due to the Prettier bug.  In such cases, I may have made manual changes to remove duplicated keywords, but left the magic keyword alone if they weren't ready to be converted!

Please feel free to modify any of my PRs as much as you'd like, by the way!

PR #23150, opened 2024-08-19
chore(zh-tw): introducing new markdownlint rules

yin1999:
### Description

introducing new markdownlint rules

### Motivation

We've applied those rules in [en-us](https://github.com/mdn/content/blob/main/.markdownlint.jsonc).


jasonren0403 (approved):
👍

PR #23151, opened 2024-08-19
[zh-tw] unify translation of HT"ML"

TSinChen:
### Description
Unify translation of HT"ML"
In some places, it is translated as '標記語言' while in other places, it is translated as '標示語言'
Additionally, neither English nor Simplified Chinese refer to HTML as a 'programming language' (程式語言).

### Motivation
Ensure translation consistency.


yin1999 (approved):
Good catch. Thank you @TSinChen :+1:

PR #23199, opened 2024-08-23
zh-cn: update the translation of `Element.prefix`

yin1999:
### Description

update the translation of `Element.prefix`

### Additional details

The translation of "namespace-aware" inspired by https://docs.python.org/zh-cn/3/library/xml.etree.elementtree.html#functions


jasonren0403 (approved):
👍

PR #23247, opened 2024-08-25
fix(ko): fix where the text wasn't being correctly bolded.

C17AN:
### Description

- Making some text adjustments to ensure the content is properly bolded.

### Additional details

[1. 예외 (Exception)](https://developer.mozilla.org/ko/docs/Glossary/Exception)

**- AS-IS**
image

**- TO-BE**
image

[2. EXIF](https://developer.mozilla.org/ko/docs/Glossary/EXIF)

- AS-IS
image

- TO-BE
image


wisedog (approved):
수정 감사합니다!

PR #23410, opened 2024-09-05
[zh-cn] fix the syntax of the CSS `ellipse()` function

Ortonzhang:
fix  Try it in ellipse syntactic error

### Description

Try it in ellipse have one syntactic error, the code doesn't work.

### Motivation

The code doesn't work.


Ortonzhang:
oops, The fix is the code for the Syntax module

yin1999 (commented):
Hey @Ortonzhang. Could you please update the [en-US doc](https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path) first?

yin1999 (approved):
Thank you @Ortonzhang. The upstream PR has been merged, so we can merge this one now :100:

PR #23708, opened 2024-09-20
chore(fr): handling redirects of chrome.com

PassionPenguin:
1. handling redircts of `chrome.com`


SphinxKnight (approved):
Thx @PassionPenguin !

PR #23754, opened 2024-09-23
Update arial-label with aria-labelledby links

manudefrutosvila:
### Description

Update aria-labelledby links

### Motivation

Broken links


Graywolf9 (approved):

Note: The links needs to point to `es`  in that way when the page exists it is already pointing to that, else the bot point to `en-US` automatically

PR #23764, opened 2024-09-24
[zh-CN]: fix the broken link to the `path()` CSS function

xionkq:
### Description
update `path` reference link to Chinese on `clip-path` page

### Motivation
I noticed that the `path` reference link on the `clip-path` page points to an English link, but actually `path` has a Chinese page.

image
image


yin1999 (approved):
Thank you @xionkq :+1:

PR #23788, opened 2024-09-26
fix: textarea typo

Pablo-Rio:
### Description

I've corrected a French word: “taille” which takes two letters L 

### Motivation

This correction helps to read the documentation in French.
This is the translation of the word "size".


cw118 (approved):
Thank you for this correction @Pablo-Rio !

Issue #23779, opened 2024-09-26
[zh-CN]: add & sync translation for Keyboard API & VirtualKeyboard API & Pointer Lock API


skyclouds2001:
```[tasklist]
- [x] [VirtualKeyboard API](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API)
- [x] [HTMLElement.virtualKeyboardPolicy](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/virtualKeyboardPolicy)
- [x] [Navigator.virtualKeyboard](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/virtualKeyboard)
- [x] [VirtualKeyboard](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard)
- [x] [VirtualKeyboard.boundingRect](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard/boundingRect)
- [x] [VirtualKeyboard.overlaysContent](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard/overlaysContent)
- [x] [VirtualKeyboard.hide()](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard/hide)
- [x] [VirtualKeyboard.show()](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard/show)
- [x] [VirtualKeyboard: geometrychange](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard/geometrychange_event)
- [ ] https://github.com/mdn/translated-content/pull/23836
- [ ] https://github.com/mdn/translated-content/pull/24286
- [x] [Navigator.keyboard](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/keyboard)
- [ ] https://github.com/mdn/translated-content/pull/23856
- [ ] https://github.com/mdn/translated-content/pull/23890
- [ ] https://github.com/mdn/translated-content/pull/24206
- [ ] https://github.com/mdn/translated-content/pull/24207
- [ ] https://github.com/mdn/translated-content/pull/23855
- [ ] https://github.com/mdn/translated-content/pull/23950
- [ ] https://github.com/mdn/translated-content/pull/24187
- [ ] https://github.com/mdn/translated-content/pull/23952
- [ ] https://github.com/mdn/translated-content/pull/24085
- [ ] https://github.com/mdn/translated-content/pull/24086
[...]

PR #23799, opened 2024-09-27
[zh-CN]: sync translation for `MouseEvent.{movementX, movementY}`

skyclouds2001:
### Description

[MouseEvent.movementX](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX)
[MouseEvent.movementY](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementY)

part of the https://github.com/mdn/translated-content/issues/23779


yin1999 (approved):
:+1:

PR #23826, opened 2024-09-29
zh-cn: correct the example in CSS `@layer`

typenoob:
### Description

Corrected css layer example in zh-CN to match the discription below.

### Motivation

The example in zh-CN is different with the [en-US](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer)

### DIFF
--- a/files/zh-cn/web/css/@layer/index.md
+++ b/files/zh-cn/web/css/@layer/index.md
@@ -142,7 +142,7 @@ p {
 ```html
 <div class="item">
   I am displayed in <code>color: rebeccapurple</code> because the
-  <code>type</code> layer comes after the <code>base</code> layer. My green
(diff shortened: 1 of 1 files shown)


files/zh-cn/web/css/@layer/index.md
```diff
@@ -142,7 +142,7 @@ p {
 ```html
 <div class="item">
   I am displayed in <code>color: rebeccapurple</code> because the
-  <code>type</code> layer comes after the <code>base</code> layer. My green
+  <code>special</code> layer comes after the <code>base</code> layer. My green
   border, font-size, and padding come from the <code>base</code> layer.
```
yin1999, line 146: 看一下是否方便把这个示例翻译一下（翻译成中文）

yin1999 (approved):
Thank you @typenoob

PR #23961, opened 2024-10-10
[ru] update box-sizing: change position-relative to position: relative

anton-chernianu:
### Description

change **position-relative** to **position: relative**


saionaro (approved):
@anton-chernianu спасибо за вклад в MDN!

PR #24009, opened 2024-10-13
Fix #23258

### LINKED ISSUE #23258: [fr] Does DELETE permit body?
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE

### What specific section or headline is this issue about?

DELETE

### What information was incorrect, unhelpful, or incomplete?

My coworker and I had a discussion and since our office is bilingual (french and english) we noticed, after some arguing, that the information is different in the documentation depending on the language. I am not 100% sure which one is correct, but depending on the language you choose, the value for `Request has body` changes.

**English:**
Request has body: **May** (meaning that is is indeed possible and supported)

**French:**
La requête a un corps: **Non** (meaning that it is not possible)

**Español:**
Petición con cuerpo: **Puede** (If i'm not mistaken it means that it is possible and supported)

### What did you expect to see?

Every languages needs to display the same correct information

### Do you have any supporting links, references, or citations?

_No response_

### Do you have anything more you want to share?

_No response_

SphinxKnight:
An update from en-US to close #23258

### DIFF
--- a/files/fr/web/http/methods/delete/index.md
+++ b/files/fr/web/http/methods/delete/index.md
@@ -1,41 +1,40 @@
 ---
 title: DELETE
 slug: Web/HTTP/Methods/DELETE
+l10n:
+  sourceCommit: 803a7a460338dab4900fd44cf7ab569e3bf5799a
 ---
 
 {{HTTPSidebar}}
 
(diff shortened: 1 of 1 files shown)


tristantheb (approved):
ça me semble bon 👍🏻

PR #24137, opened 2024-10-19
zh-cn: fix markdown rendering error

yin1999:
### Description

fix markdown rendering error


yin1999 (commented):
minor fixes. self-approval

PR #24189, opened 2024-10-21
[es] Fixed code inline to show correctly

cluroc:
### Description

Fixed code inline to show correctly in Django/Models


Graywolf9 (approved):

PR #24263, opened 2024-10-28
zh-cn: update the browser compatibility for `AVIF`

imba-tjd:
see https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types

图片


yin1999 (approved):
Great. Thank you @imba-tjd

PR #24315, opened 2024-11-02
Fix ja String.prototype.isWellFormed() return value explanation

MaySoMusician:
### Description

The relationship of `String.prototype.isWellFormed()` return value between `true` and `false` was reversed in Japanese translation.

### Motivation

It's a confusing mistranslation.

### Additional details

N/A

N/A


MaySoMusician:
Close per the problem is fixed in 24304

PR #24362, opened 2024-11-05
zh-cn: update the translation of `DataTransferItemList.length`

yin1999:
### Description

update the translation of `DataTransferItemList.length`


jasonren0403 (approved):
👍

PR #24414, opened 2024-11-09
zh-tw: replace the polyfill with the link to `core-js`

Jasper19901029:
When using typeof func(){}, the result should display as "function" instead of "Function".


yin1999 (approved):
Thanks @Jasper19901029. Because polyfills are expensive to maintain and are prone to security risks, we now "only" provide links to `core-js` polyfills.

PR #24459, opened 2024-11-11
fix mistyped macro call

argl:
### Description

Fixes a a typo in a `jsxref` macro call

### Motivation

discovered by rari/yari parity checks


1ilsang (approved):
LGTM 👍

Issue #24188, opened 2024-10-21
[zh-CN]: add translation for Navigation API


skyclouds2001:
- [x] https://github.com/mdn/translated-content/pull/24519
- [x] [Window.navigation](https://developer.mozilla.org/en-US/docs/Web/API/Window/navigation)
- [x] https://github.com/mdn/translated-content/pull/24402
- [x] https://github.com/mdn/translated-content/pull/24389
- [x] https://github.com/mdn/translated-content/pull/24523
- [x] https://github.com/mdn/translated-content/pull/24524
- [x] https://github.com/mdn/translated-content/pull/24390
- [x] https://github.com/mdn/translated-content/pull/24505
- [x] https://github.com/mdn/translated-content/pull/24578
- [x] https://github.com/mdn/translated-content/pull/24544
- [x] https://github.com/mdn/translated-content/pull/24579
- [x] https://github.com/mdn/translated-content/pull/26228
- [x] https://github.com/mdn/translated-content/pull/26229
- [x] https://github.com/mdn/translated-content/pull/26230
- [x] https://github.com/mdn/translated-content/pull/24593
- [x] https://github.com/mdn/translated-content/pull/24619
- [x] https://github.com/mdn/translated-content/pull/26221
- [x] https://github.com/mdn/translated-content/pull/26222
- [x] https://github.com/mdn/translated-content/pull/26223
- [x] https://github.com/mdn/translated-content/pull/25299
- [x] https://github.com/mdn/translated-content/pull/25670
- [x] https://github.com/mdn/translated-content/pull/25673
- [x] https://github.com/mdn/translated-content/pull/25684
- [x] https://github.com/mdn/translated-content/pull/25697
- [x] https://github.com/mdn/translated-content/pull/25699
- [x] https://github.com/mdn/translated-content/pull/25701
- [x] https://github.com/mdn/translated-content/pull/25499
- [x] https://github.com/mdn/translated-content/pull/24858
- [x] https://github.com/mdn/translated-content/pull/24869
- [x] https://github.com/mdn/translated-content/pull/24870
- [x] https://github.com/mdn/translated-content/pull/25301
- [x] https://github.com/mdn/translated-content/pull/25671
- [x] https://github.com/mdn/translated-content/pull/25674
- [x] https://github.com/mdn/translated-content/pull/25685
- [x] https://github.com/mdn/translated-content/pull/25698
- [x] https://github.com/mdn/translated-content/pull/25700
[...]

PR #24578, opened 2024-11-18
[zh-CN]: add translation for `Navigation.back()`

skyclouds2001:
### Description

[Navigation.back()](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/back)

part of the https://github.com/mdn/translated-content/issues/24188


yin1999 (approved):
:+1:

PR #24672, opened 2024-11-22
[zh-cn]: fix the description of the `top` CSS property

ZeroOneJs:
### Description

fix typo


yin1999 (approved):
Thanks @zero-one-code. I've pushed another commit to fix the context.

PR #24715, opened 2024-11-23
zh-cn: fix typo

CHAAAAA:
"SystaxError: missing } after function body"
change to
"SyntaxError: missing } after function body"


yin1999 (approved):
Good catch @CHAAAAA! Welcome to the community :tada:

PR #24780, opened 2024-11-26
zh-cn: correct the description in CSS `float` property

Tmaof:
### Description

浮动元素碰到包含块的内容区域边界就会停下，而不是到达边框区才停下。
Floating elements stop when they hit the boundary of the content area containing the block, rather than when they reach the border area.

### Motivation
这样描述可以更加准确的表明浮动的行为。
This description gives a more accurate indication of the behavior of the float.

### Additional details

https://developer.mozilla.org/zh-CN/docs/Web/CSS/float


yin1999 (approved):
感谢

PR #24870, opened 2024-12-01
[zh-CN]: add translation for `NavigationTransition.{from,navigationType}`

skyclouds2001:
### Description

[NavigationTransition.from](https://developer.mozilla.org/en-US/docs/Web/API/NavigationTransition/from)
[NavigationTransition.navigationType](https://developer.mozilla.org/en-US/docs/Web/API/NavigationTransition/navigationType)

part of the https://github.com/mdn/translated-content/issues/24188


yin1999 (approved):
Thank you @skyclouds2001

PR #24906, opened 2024-12-02
Add french translation for numerical separators

romaindurand:
add french translation for numerical separators

### Description

Numerical separators were not translated in the french documentation.

### Motivation

I'd like to link to the french documentation but it doesn't provide the necessary informations


tristantheb (approved):
👌🏻

PR #25013, opened 2024-12-11
[es]: auto-fix links

argl:
### Description

This fixes plain markdown links to point to the actual destination without going through the whole redirect stack.
(MP-1798)
### Motivation

To continue our transition to rari, we have to weed out any links in documentation that rely on "fundamental redirects". Fundamental redirects (as opposed to our main redirect map) are only there to support requests coming in from really outdated links on the internet.

### Additional details

Luckily, yari already has tooling around fixing those links, so this is basically a version of translated content with the fix-flaws/broken-links tooling applied. The result looks very good, and even shows improvements on linking less often to the en-US content when there is a translated version of the page available.

Using the rari diff tool, I have attached a HTML file for each locale that highlights all differences before/after the change on the simplified HTML rendering of the content:
[Diffs.zip](https://github.com/user-attachments/files/18124441/Diffs.zip)


fiji-flo (approved):
This is an automated clean-up and we depend on landing this. Please reach out if this caused any issues which it shouldn't. You can look at the actual changes in this [diff](https://hansknall.org/diff-es.html).

PR #25032, opened 2024-12-12
[ja]: auto-fix links

argl:
### Description

This fixes plain markdown links to point to the actual destination without going through the whole redirect stack.
(MP-1798)
### Motivation

To continue our transition to rari, we have to weed out any links in documentation that rely on "fundamental redirects". Fundamental redirects (as opposed to our main redirect map) are only there to support requests coming in from really outdated links on the internet.

### Additional details

Luckily, yari already has tooling around fixing those links, so this is basically a version of translated content with the fix-flaws/broken-links tooling applied. The result looks very good, and even shows improvements on linking less often to the en-US content when there is a translated version of the page available.

Using the rari diff tool, I have attached a HTML file that highlights all differences before/after the change on the simplified HTML rendering of the content.
[Diffs.zip](https://github.com/user-attachments/files/18124510/Diffs.zip)


mfuji09 (changes requested):
This PR has too many files to review at once.
Please split this PR into several PRs which have less than 100 files.

Please refer to <https://developer.mozilla.org/en-US/docs/MDN/Community/Pull_requests#open_a_pull_request>,

fiji-flo:
@mfuji09 Thanks for looking at this. This PR does two things.
- It put's into markdown what yari did magically during deployment.
- It improves some redirects that were broken in yari. E.g [Games/Techniques#ウェブゲーム用の効率的アニメーション](https://developer.mozilla.org/ja/docs/Games/Techniques#%E3%82%A6%E3%82%A7%E3%83%96%E3%82%B2%E3%83%BC%E3%83%A0%E7%94%A8%E3%81%AE%E5%8A%B9%E7%8E%87%E7%9A%84%E3%82%A2%E3%83%8B%E3%83%A1%E3%83%BC%E3%82%B7%E3%83%A7%E3%83%B3) was linking to an en-US page and it shouldn't.

You can look at the actual differences in the [diff](https://hansknall.org/diff-ja.html)

We want to have single PR's per locale to keep the noise down and to be able to revert if anything happens (we have it on stage an we didn't see any issues).

We want go ahead with this as this is more of an platform change than content. We are happy to help with any inconvenience this might create.

mfuji09:
@fiji-flo 
However, my browser freezes when I am trying to write review comments, perhaps due to too many files.

fiji-flo:
Sorry for that, yes GitHub is not great at this. Feel free to push to this PR and/or we just talk on discord in the #mdn-platform-🌐 channel?

fiji-flo (approved):
This is an automated clean-up and we depend on landing this. Please reach out if this caused any issues which it shouldn't. You can look at the actual changes in this [diff](https://hansknall.org/diff-ja.html).

@mfuji09 I apologize for this inconvenience, please let me know if I can help in any way.

fiji-flo:
We're merging this since otherwise links on production will break.

PR #25161, opened 2024-12-19
chore(docs): use the GFM alerts syntax instead

yin1999:
### Description

use the GFM alerts syntax instead


pepelsbey (approved):
Looks good! Thank you :)

PR #25174, opened 2024-12-20
zh-cn: replace the deprecated functions in example

Stan-BK:
### Description

Exchange a function which aims to import and instantiate a wasm instance.

### Motivation

Previous function has been deprecated and now we have the more advanced function.


yin1999 (approved):
Thanks @Stan-BK. I've pushed a commit to sync the example from en-US. As the arrow function format is more concise.

PR #25176, opened 2024-12-20
fix: correcting function result

Stan-BK:
### Description

Add a result for the function which ends up with not null stack.

### Motivation

Correct the function in order to run successfully.


yin1999 (approved):
Thank you @Stan-BK

PR #25204, opened 2024-12-22
feat(package): Add packageManager for corepack users

1ilsang:
### Description

image

image

Add packageManager field for Corepack users.

It is the same work as https://github.com/mdn/content/issues/33946


yin1999 (approved):
This is a sync with upstream content, I think we can merge it

PR #25469, opened 2025-01-16
Over-translation in ja

ogasawara-k:
Translation is different from other pages

### Description

If you follow the example in the documentation, your buttons will be labeled "Edit" and "Cancel" in any language.
Documentation on other pages is also listed according to the button labels.

### Motivation

The notation will be unified and easier to understand.


mfuji09 (approved):
OKです。
なお、日本語版記事のレビュアーは全員が日本人ですので、PRは日本語でどうぞ。

PR #25538, opened 2025-01-22
chore(scripts): replace cld with franc to detect language

yin1999:
### Description

The `cld` package works well, but it requires the user to install MSVC to compile the C source code. But some contributors may not be C/C++ developers, and they don't want to install such a _heavy_ compiler on their computers (MSVC build tools requires 8GB of disk space). Additionally, the `cld` package itself requires 100MB of disk space (see [the warning](https://github.com/dachev/node-cld/tree/master?tab=readme-ov-file#warning)).

After comparing with some language detectors ([cld](https://github.com/dachev/node-cld), [franc](https://github.com/wooorm/franc), [cld3-asm](https://github.com/kwonoj/cld3-asm)), I believe `franc` is a well substitute for us, as it has more users and is lighter (implemented in pure javascript).

Note that, the replacement will increase the time used for language detection. I've tested to run through all the files:

| package | Used time (seconds) |
| --- | --- |
| cld | 16.863 |
| franc | 95.401 |

### Motivation

Makes it easier for contributors (especially for Windows users) to setup the environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,9 +28,10 @@
   },
   "dependencies": {
     "autocorrect-node": "2.13.0",
-    "cld": "^2.10.0",
     "fdir": "^6.4.3",
+    "franc-min": "^6.2.0",
     "front-matter": "^4.0.2",
+    "html-to-text": "^9.0.5",
     "husky": "9.1.7",
     "lint-staged": "15.4.3",
     "markdown-it": "^14.1.0",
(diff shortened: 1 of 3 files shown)


yin1999:
Hey @queengooborg, this is a draft to replace the `cld` package (the motivation is describded above). Are you interested in taking a look?

caugner:
LGTM if @queengooborg has no objections.

queengooborg:
No objections here!

caugner (approved):

yin1999:
Thanks for your review @caugner, @queengooborg. Let's try the new language detector :)

PR #25687, opened 2025-02-01
[zh-CN]: Sync translation for `Permissions-Policy`: `web-share`

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy/web-share


yin1999 (approved):
:+1:

PR #25715, opened 2025-02-04
[fr] fix typo: missing words in sentence

vincent-peugnet:
Hi ! I suggest a little correction in french HTTP/Headers/cache-control section 😃

### Description

Add two words to a sentence about cache control so that it make sense.

### Motivation

I was reading this and it seems obvious that there where missing words. I think a text is best without missing words.


tristantheb (approved):
Merci pour la correction 👍🏻

PR #25721, opened 2025-02-05
fix(workflows/pr-check_redirects): validate with rari

caugner:
Reverts mdn/translated-content#25629 and removes the no longer supported `--strict` option.


caugner (commented):
Self-approval, as this a trivial change.

PR #25766, opened 2025-02-07
[fr] add a section on using CSS animations with Flexbox

Milrer:
Description
Added a section explaining how to use CSS animations with Flexbox. This new section includes a demonstration using https://github.com/Keyframes to animate flexible elements, as well as practical tips on animating Flexbox properties such as order, flex-grow and justify-content.

Motivation
This enhancement gives developers a better understanding of how to combine Flexbox and CSS animations to create more dynamic and interactive interfaces. Currently, the page doesn't cover this topic, even though it's common practice in modern web development.


Milrer:
Hello, is it possible to do anything about this modification? Everything seems to have been ready for over a week. Please let me know if there's anything I can do to help, even if it means becoming an official contributor.

cw118:
Hi @Milrer, sorry for the delay here. As caugner mentioned, French content is translated from the en-US documentation, and the additions in this PR do not seem to be on the [en-US page](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations). Could you first create a PR at [mdn/content](https://github.com/mdn/content) to suggest these changes in English?

If it is accepted for the en-US version, then we can consider it here too. Thank you :)

Milrer:
Hi @cw118 😄 

No problem, it's a pleasure to be able to contribute to the institution that is MDN! On the other hand, the French translation is already different from the English version, so if I modify the English version it will always be different from the French version (breakings changes).

cw118:
> No problem, it's a pleasure to be able to contribute to the institution that is MDN! On the other hand, the French translation is already different from the English version, so if I modify the English version it will always be different from the French version (breakings changes).

Yes, that is because the French translation has unfortunately not been updated vs. en-US in some time, but hopefully that can be done soon 😅

What we meant is that we don't add new sections to the French page if they are not already found in the English version, which appears to be true here. So could you please suggest adding your example on mdn/content first? If it is accepted there, then it can be added to the French content as well. Thank you for understanding :)

PR #25789, opened 2025-02-08
[zh-CN]: Sync translation for `KeyboardLayoutMap.{entries,keys,values}`

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardLayoutMap/keys
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardLayoutMap/values
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardLayoutMap/entries


yin1999 (approved):
:+1:

PR #25879, opened 2025-02-13
fix: correct typo from 'Unit8Array' to 'Uint8Array'

DDDDD12138:
### Description

Corrected a typo in the documentation where `Unit8Array` was mistakenly written instead of `Uint8Array`.


yin1999 (approved):
Thank you @DDDDD12138

PR #25902, opened 2025-02-15
[zh-CN]: Sync translation for Web Locks API

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API


jasonren0403 (approved):
👍

PR #25904, opened 2025-02-15
[zh-CN]: Sync translation for `MessagePort.{message, messageerror}_event`

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/message_event
https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/messageerror_event


yin1999 (approved):
👍

PR #25912, opened 2025-02-16
[zh-cn] correct the format of <form>

Lampese:
image

In the current Simplified Chinese document, the format of the form title is not accurate.


yin1999 (approved):
Thanks

PR #26002, opened 2025-02-23
[zh-CN]: Add translation for `NavigateEvent.info`

skyclouds2001:
### Description

[NavigateEvent.info](https://developer.mozilla.org/en-US/docs/Web/API/NavigateEvent/info)


yin1999 (approved):
:+1:

PR #26128, opened 2025-03-03
zh-cn: fix the typo in JavaScript `static` keyword

supriome:
Fix: methods call with wrong name, should be SquaredTriple.calculate(3) rather than SquaredTriple.tripple(3)

### Description

Typo: Change `SquaredTriple.tripple(3)` to  `SquaredTriple.calculate(3)`


yin1999 (approved):
Thank you @supriome. Welcome to the community :tada:

PR #26155, opened 2025-03-04
[es]: Migrate tabbed CSS interactive examples

LeoMcA:

This PR migrates the tabbed CSS interactive examples, reflecting the changes made in en-US in https://github.com/mdn/content/pull/38356

Like we did with HTML, we've tested these examples with a couple of diff tools to ensure they all work and remain the same as before. See the zipfile of the visual diff in en-US in the above PR for the few small changes in default styling we've made.

No full manual review is necessary, but please do flag if anything seems off: you can check the examples in the preview urls.


Graywolf9 (approved):
Thank you so much for your help! Everything looks good I also fix some flaw

PR #26168, opened 2025-03-04
more casing fixes

argl:
### Description

Slug casing fixes

### Motivation

Problems after slug case changes in en-us content


bsmth (approved):
+1 from me here, same as https://github.com/mdn/content/pull/38446#pullrequestreview-2658206729

PR #26202, opened 2025-03-07
[ko] images are broken for border of edge, corner of edge

social-archive:
### Description

add images src

### Motivation

correct src exists for images


marvin-seo-dev:
@sunhpark42

PR #26208, opened 2025-03-07
Remove "experimental icon" for let and const statements

urffin:
Remove "experimental icon" for let and const statements

### Description

Remove experimental icons for "let" and "const" statements

### Motivation

Consistent state with en version


saionaro (approved):
@urffin thank you for your MDN contribution!

PR #26247, opened 2025-03-08
[zh-CN]: Sync translation for `{ServiceWorkerGlobalScope,Window}.cookieStore`

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/API/Window/cookieStore
https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/cookieStore


yin1999 (approved):
👍

PR #25577, opened 2025-01-24
feat(workflows): migrate pr-test to rari

caugner:
### Description

Migrates the `pr-test` workflow to rari.

### Additional details

The existing `pr-test` workflow was renamed to `pr-test-legacy`, and disabled.

Both workflows now use the `pull_request` event, while `pr-review-companion` uses the `workflow_run` event.

### Related issues and pull requests

Related content PRs:

- https://github.com/mdn/content/pull/37572
- https://github.com/mdn/content/pull/37576
- https://github.com/mdn/content/pull/37586
- https://github.com/mdn/content/pull/37766
- https://github.com/mdn/content/pull/37771

PR #26355, opened 2025-03-13
ci: remove pr-test-legacy workflow

caugner:
### Description

Remove the `pr-test-legacy` workflow, which was added in https://github.com/mdn/translated-content/pull/25577 during the transition to rari.

### Motivation

We no longer use it, and it's already disabled.

Same as:

- https://github.com/mdn/content/pull/38611


caugner (commented):
Self-approval.

PR #26356, opened 2025-03-13
fix(macros): no more multiline macros

fiji-flo:
### Description

We remove support for those.

Programmatically replaced `(\{\{[^}\n]+)\n *([^}]+}})` and tested that it produces the exact same output.


mfuji09 (approved):
LGTM for `ja`.

caugner (approved):
LGTM, verified that those are just whitespace changes.

yin1999 (approved):
LGTM for l10n-zh. Thank you @fiji-flo 🙏

fiji-flo:
Thanks everyone. Merging since we verified this renders the exact same.

PR #26372, opened 2025-03-14
[zh-cn]: correct the Webkit Bug link releated to "the focus movement issue"

YiiChitty:
### Description

[WebKit bug 3869](https://webkit.org/b/3869) is takling about  "use HTML Image element inplace of JS Image object"，it‘s none of business with Selection API。But [WebKit bug 38696](https://webkit.org/b/38696) , which is mention in  [W3C Bug  14383](https://www.w3.org/Bugs/Public/show_bug.cgi?id=14383)，is “WebKit moves focus to where selection is instead of moving selection to the focused element like other browsers”。
I think is a mistake ：）

### Motivation

The incorrect description could mislead readers about the nature of the bug. This correction ensures the issue is accurately represented, helping developers and readers better understand the behavior.

### Additional details
null

null


yin1999 (approved):
Good catch @YiiChitty. Welcome to the community 🎉

PR #26432, opened 2025-03-17
[zh-CN]: fix a broken link in "font-size" page

EienMiku:
### Description

`[CSS](/zh-CN/docs/Web/docs/CSS)`, which directs to a page that does not exist, should be `[CSS](/zh-CN/docs/Web/CSS)`.


yin1999 (approved):
Thank you @EienMiku

PR #26475, opened 2025-03-19
Update DataView.setUint16 fr reference

jlebras:
- Replace `32` bits occurrence by `16` as setUint16 works with 16 bits values.
- Fix typo in example title.


cw118 (approved):
Thank you for these corrections @jlebras !

PR #26482, opened 2025-03-19
feat: Add Lefthook for precommit hooks

### LINKED ISSUE #26472: Port Lefthook changes over from en-US repo
In en-US content, we're trying out Lefthook to speed up authoring. We can look at porting these changes over here.

See:

- [x] https://github.com/mdn/content/pull/38399

__PR:__

- [ ] https://github.com/mdn/translated-content/pull/26482

bsmth:
### Motivation

Porting over changes from en-US content

See:

- [x] https://github.com/mdn/content/pull/38399

Fixes https://github.com/mdn/translated-content/issues/26472


lefthook.yml
```diff
@@ -0,0 +1,39 @@
...
+            run: yarn markdownlint-cli2 --fix {staged_files}
+            stage_fixed: true
+          - name: Check URLs per locale
+            run: yarn node ./scripts/check-url-locale.js --fix {staged_files}
+            stage_fixed: true
+          - name: Autocorrect
+            run: yarn autocorrect --fix {staged_files}
+            stage_fixed: true
+          - name: Check changed xrefs
+            run: yarn node scripts/update-moved-file-links.js --check
```
yin1999, line 24: I'm not sure if this is correct, because there is no such script file in translated-content

bsmth: Should be fixed in 74c1ffad7d5ed44e07f4101e037afe83d6370d3a

yin1999 (approved):
Thank you @bsmth. The new precommit hook runs very fast (especially on my Windows).

lefthook.yml
```diff
@@ -0,0 +1,40 @@
...
+      glob: "*.md"
+      group:
+        parallel: true
+        jobs:
+          - name: Markdownlint
+            run: yarn markdownlint-cli2 --fix {staged_files}
+            stage_fixed: true
+          - name: Check URLs per locale
+            run: yarn node ./scripts/check-url-locale.js --fix {staged_files}
+            stage_fixed: true
```
bsmth, line 19: @yin1999 - this script hangs for me if I try to trigger it with this in `files/ja/web/api/fetch_api/using_fetch/index.md`:

```diff
+ [フェッチ API](/es/docs/Web/API/Fetch_API) 
- [フェッチ API](/ja/docs/Web/API/Fetch_API) 
```

What do you think about leaving this out for now until we get it running? 

```suggestion
```

Aside from being able to run this script independently of pre-commit, this could be the type of thing that's built into rari directly in future. See https://github.com/mdn/rari/pull/158

yin1999: Hey, @bsmth. I tested this on my device and it seems to work fine without hangs. Did you test with the commit that only contained changes to this link? This may produce an empty commit (requiring you to fill in the commit message interactively).

bsmth: Actually the problem is the ora spinner which has some TTY interference. I've removed it in https://github.com/mdn/translated-content/pull/26482/commits/f69fa8535a0dec5f002e4147bf7ee536586108d9 which works a lot better, any thoughts on this?

yin1999: Thanks for your thorough investigation, I have no objection to removing the ora spinner from the script. I think it would be better if rari could integrate these functions directly in the future 😄

lefthook.yml
```diff
@@ -0,0 +1,40 @@
...
+      glob: "*.md"
+      group:
+        parallel: true
+        jobs:
+          - name: Markdownlint
+            run: yarn markdownlint-cli2 --fix {staged_files}
+            stage_fixed: true
+          - name: Autocorrect
+            run: yarn autocorrect --fix {staged_files}
+            stage_fixed: true
```
caugner, line 19: Could we move autocorrect to a separate rule that only runs on `zh-cn`, see: https://github.com/mdn/translated-content/blob/68dc499b9a2fa51a508df9f866c06aee6601fb64/.autocorrectignore#L6-L7

scripts/check-url-locale.js
```diff
@@ -244,16 +239,17 @@ async function main() {
 
   let exitCode = 0;
 
+  console.info(`Crawling ${files.length} files…`);
```
caugner, line 242: Why move this down, when the crawling happens above?

scripts/check-url-locale.js
```diff
@@ -12,14 +12,11 @@
 import fs from "node:fs/promises";
 import * as path from "node:path";
 import { fdir } from "fdir";
-import ora from "ora";
 import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
 import { fromMarkdown } from "mdast-util-from-markdown";
 import { visit } from "unist-util-visit";
 
-const spinner = ora().start();
```
caugner, line 21: Instead of removing `ora`, let's set `no_tty` in the lefthook config: https://lefthook.dev/configuration/no_tty.html

caugner (changes requested):

lefthook.yml
bsmth: >I think it would be better if rari could integrate these functions directly in the future 

You're very welcome and I totally agree.

scripts/check-url-locale.js
bsmth: Well spotted, I didn't know about this config

scripts/check-url-locale.js
bsmth: I'll see if `no_tty` works and if so, will revert these changes

scripts/check-url-locale.js
bsmth: See https://github.com/mdn/translated-content/pull/26482/commits/31d2eb49013fff33114608e6bb32abbdd16de0e9

scripts/check-url-locale.js
bsmth: Resolved in https://github.com/mdn/translated-content/pull/26482/commits/31d2eb49013fff33114608e6bb32abbdd16de0e9

lefthook.yml
bsmth: done in https://github.com/mdn/translated-content/pull/26482/commits/31d2eb49013fff33114608e6bb32abbdd16de0e9

caugner (approved):

PR #26511, opened 2025-03-21
ci(system-file-changes): pass if author/actor is an admin

caugner:
### Description

Prevents the `system-file-changes` workflow from failing, if both PR author and actor are admin.

### Motivation

Reduces noise when admins update system files, and reduces the chances of missing other failing workflows.

Same as:

- https://github.com/mdn/content/pull/38610


bsmth (approved):
ty!

PR #26602, opened 2025-03-27
[ru]: migrate remaining interactive examples

LeoMcA:
These should be the final migrations outside of the kitchensink and writing guidelines, part of the wider project described here: https://github.com/orgs/mdn/discussions/782

We haven't QAed these as thoroughly as the previous migrations (difficult to do, as they weren't detected by our scripts, so we've had to manually migrate these), so I'd suggest a more "hands-on" review compared to the other migrations: thankfully it's only a few examples!


leon-win:
Thank you very much, @LeoMcA !
I have corrected some inaccuracies and added translations of some examples.

leon-win (approved):

leon-win:
Added the changes and merged PR :ok_hand:

PR #26618, opened 2025-03-28
chore(css): use embedlivesample for css-examples [zh-cn]

bsmth:
### Description

I'm porting over some of the embedded examples into translated content.

I look at embeds like `{{embedgh.*css-examples.*}}`

### Motivation

These are already live samples in en-US.

### For reviewers

The easiest way to review is to compare with en-US version in the previews. The actual code is copied across from the content repo, so there's no technical review to do, just a quick verify that it looks okay!

### Additional details

- [x] https://github.com/mdn/mdn/issues/597


jasonren0403 (approved):
Thank you!

PR #26740, opened 2025-04-09
es: remove interactive-examples references

caugner:
### Description

Removes references to the interactive-examples repo.

### Motivation

We no longer include examples from there, so we should no longer point there.


Graywolf9 (approved):
Thank you so much!

PR #26743, opened 2025-04-09
ru: remove interactive-examples references

caugner:
### Description

Removes references to the interactive-examples repo.

### Motivation

We no longer include examples from there, so we should no longer point there.


files/ru/web/javascript/reference/global_objects/function/name/index.md
```diff
@@ -23,8 +23,6 @@ console.log(object.func2.name);
 // Expected output: "func2"
 ```
```
leon-win, line 25:
```suggestion

## Значение

Строка.

```

leon-win:
Hi @caugner!

Thank you for removing odd references! :+1: 

I added a small suggestion describing the return value.

PR #26754, opened 2025-04-09
Lefthook nits

caugner:
### Description

1. Rename lefthook config to `.lefthook.yml`, to ensure it's treated as a system file.
2. Remove husky leftovers in `auto-cleanup-bot` and labeler config.

_Note_: By default, lefthook doesn't install hooks in CI, so we don't need to set `LEFTHOOK=0`.

Same as:

- https://github.com/mdn/content/pull/38825


bsmth (approved):
Ty!

PR #26755, opened 2025-04-09
Rename `.front-matter-config.json`

caugner:
### Description

Ensures that it's treated as a system file by default, and picked up by the `system-file-changes` workflow.


bsmth (approved):
thanks!

PR #27058, opened 2025-04-30
[fr] Typofix for CSS `var()`

oub:
### Description

Fix typo on word "subsitution" -> "substitution"


cw118 (approved):
Merci pour cette correction @oub !

PR #27063, opened 2025-05-01
fix typo

Llldmiao:
### Description
There is a typo in chinese version of https://developer.mozilla.org/zh-CN/docs/Mozilla/Add-ons/WebExtensions/Extending_the_developer_tools
image

### Motivation
I was reviewing this document and noticed this typo, so I fixed it on the spot.


yin1999 (approved):
:+1:

PR #27200, opened 2025-05-07
Replace interactive-examples assets with shared-assets

caugner:
### Description

Replaces interactive-examples assets with shared-assets.

### Motivation

We're already replaced interactive-examples, and are deprecating the interactive-examples repo.

* Same as: https://github.com/mdn/content/pull/39440


caugner (commented):
Self-approval, after testing all examples manually.

Note: Some examples are broken, but also on prod.

PR #27365, opened 2025-05-22
chore: update index.md

YunYouJun:
optimize translation

### Description

Use `中与` instead of `的与`

The accessibility tree contains [accessibility](https://developer.mozilla.org/en-US/docs/Glossary/Accessibility)-related information for most HTML elements.

HTML elements contain accessibility-related information


yin1999 (approved):
Thanks

PR #27394, opened 2025-05-25
doc: add openomy to display active contributors

kiner-tang:
### Description

Motivate more open source contributors to contribute more to the community by displaying the community activity ranking in the past three months, such as PR, ISSUE, Discussion

image


pepelsbey:
Thank you for the suggestion, but we’re trying not to rely on third-party services too much. The gain is not worth the trouble.

PR #27400, opened 2025-05-26
[zh-cn]: fix description of viewport meta attributes

RosalynHUANG:
### Description
1. 'width' -> `width`
2. height value example: width=600 -> height=400

### Motivation
1. it causes display issue
2. height value example use width value may cause misunderstanding, and  English version use height=400 example.

### Additional details
English version:
Controls the (minimum) size of the viewport (see [viewport width and screen width](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Viewport_meta_element#viewport_width_and_screen_width)). It can be set to a specific number of pixels like height=400 or to the special value device-height, which is the physical size of the device screen in CSS pixels.

none


yin1999 (approved):
Thanks @RosalynHUANG

PR #27404, opened 2025-05-27
zh-cn: fix translation error

AlexLin625:
"have no way around" should be translated to 无法避免

### Description

Change content to correct wording.


yin1999 (approved):
Thank you @AlexLin625. Welcome to the community 🎉

PR #27665, opened 2025-06-13
[zh-CN]: fix `fit-content` example

xiaoxiyao:
### Description

Remove the `-moz-` prefix in `fit-content` example for zh-cn.

### Motivation

Keep it consistent with the English version. No prefix is needed.


jasonren0403 (approved):
👍

PR #27753, opened 2025-06-18
fix element name <article>

GNUXDAR:
### Description

fix word

### Motivation

HTML element names (e.g., <div>, <article>) must remain untranslated in the documentation

### Additional details

https://developer.mozilla.org/es/docs/Learn_web_development/Core/Styling_basics/Getting_started#combinaciones_de_selectores_y_combinadores


GNUXDAR (commented):
HTML element names (e.g., div, article) must remain untranslated in the documentation.

PR #27781, opened 2025-06-21
zh-cn: create the translate of HTML `content` attribute

cznno:
### Description

Create zh-CN translate of HTML attribute: content: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/content

### Motivation

Easier to read and understand for native Chinese speakers.


yin1999 (approved):
欢迎来到 MDN @cznno。期待你的下次贡献，如果你想了解有关简中 MDN 文档贡献的更多信息，欢迎访问：https://github.com/mdn/translated-content/blob/main/docs/zh-cn/translation-guide.md

PR #26462, opened 2025-03-18
ci: pass data betwen steps via GITHUB_OUTPUT

caugner:
### Description

Ensures workflows pass data between steps via `GITHUB_OUTPUT`.

### Motivation

Avoids unintended side-effects of using `GITHUB_ENV`.

### Related issues and pull requests

Same as:

- https://github.com/mdn/content/pull/38702

PR #27822, opened 2025-06-26
ci(pr-test): fix GIT_DIFF_FILES env

caugner:
### Description

Fixes the `pr-test` workflow, properly setting the `GIT_DIFF_FILES` variable.

### Motivation

It looks like the `filecheck` step of the `pr-test` is broken since https://github.com/mdn/translated-content/pull/26462.

### Additional details

`${{` and `$}}` can only be omitted in `if`.

See: https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions


caugner (commented):
Self-approval, as the fix is trivial.

PR #27879, opened 2025-07-01
chore(deps): bump node from v20 to v22

caugner:
### Description

Bumps Node.js from v20 to v22.

### Motivation

Node.js v22 is the currently active LTS version.

### Related issues and pull requests

Same as: https://github.com/mdn/content/pull/40141

Part of https://github.com/mdn/mdn/issues/699.

PR #28000, opened 2025-07-09
ci(workflows): use content node version

caugner:
### Description

Uses the Node version from mdn/content whenever content is checked out.

Note: Also adds names to all checkout steps.

### Motivation

When we bumped the Node version in content, some workflows in translated-content started to fail, because they used Node 20 (the translated-content version) in the context of content (requiring Node 22).

Follow-up of: https://github.com/mdn/translated-content/pull/27879#issuecomment-3025534882

### DIFF
--- a/.github/workflows/auto-cleanup-bot.yml
+++ b/.github/workflows/auto-cleanup-bot.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Setup Node.js environment
         uses: actions/setup-node@v4
         with:
-          node-version-file: ".nvmrc"
+          node-version-file: mdn/content/.nvmrc
 
       - name: Install all yarn packages
         run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/auto-cleanup-bot.yml
```diff
@@ -27,9 +27,11 @@ jobs:
...
-      - uses: actions/checkout@v4
+      - name: Checkout
+        uses: actions/checkout@v4
 
-      - uses: actions/checkout@v4
+      - name: Checkout (content)
+        uses: actions/checkout@v4
         with:
           repository: mdn/content
           path: mdn/content
```
yin1999, line 37: line 42, should we use `mdn/content/.nvmrc`? Step `Install all yarn packages for mdn/content` will install packages for `mdn/content`.

yin1999 (approved):
One question. Otherwise LGTM.

.github/workflows/auto-cleanup-bot.yml
caugner: Good catch, thanks! Fixed in 4e817a79e74.

PR #28102, opened 2025-07-22
[ja] Add explicit `なし。` to the `default-src による代替` section of CSP: base-uri, consistent with other directives

llamakko:
### Description

In the Japanese MDN pages for directives such as [CSP: form-action](https://developer.mozilla.org/ja/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/form-action) and [CSP: frame-ancestors](https://developer.mozilla.org/ja/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors), which do not fall back to default-src, the `default-src による代替` section explicitly states `なし。`, clearly indicating that no fallback occurs.

However, in the case of [CSP: base-uri](https://developer.mozilla.org/ja/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/base-uri), this explicit statement was missing. To ensure consistency and clarity, I have updated the page to explicitly indicate that there is no fallback to default-src.


mfuji09 (approved):
OKです。
なお、日本語版記事のレビュアーは全員が日本人です。
英語より日本語の方がコミュニケーションしやすいので、プルリクエストは日本語でどうぞ。

PR #28110, opened 2025-07-23
fix: incorrect description of -0 and +0 equality in map and set docs

dariasavinova:
### Description
Removed the incorrect phrase `(в отличие от ===)` (unlike ===) from the description of `-0` and `+0` equality in the SameValueZero algorithm section. 
The original English text is correct, but the Russian translation contains an error. 

### Motivation

This change corrects a misleading translation error, ensuring readers understand that `-0` and `+0` are equal in both `===` and SameValueZero.


saionaro (approved):
@dariasavinova спасибо за вклад в MDN!

PR #28200, opened 2025-07-31
[zh-cn] add translation of "Credentialed requests and wildcards"

gausszhou:
### Description

Synchronize English document content for 'Credentialed requests and wildcards'

### Motivation

Synchronize English document content

### Additional details

Add  chinese translate for  `The server must not specify the * wildcard for the Access-Control-Expose-Headers response-header value, but must instead specify an explicit list of header names; for example, Access-Control-Expose-Headers: Content-Encoding, Kuma-Revision`


yin1999 (approved):
Thank you @gausszhou

PR #28258, opened 2025-08-04
fix some erroneous comments

nano71:
### Description

Modified some code comments to make their descriptions more understandable and accurate


yin1999 (changes requested):
请按英文文档中的注释进行翻译：https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Control_flow_and_error_handling#the_finally_block

PR #28304, opened 2025-08-08
ar: Add Arabic Language

0xSm0ky:
### Description

Add Arabic Language Translation
And translate to this file: `files\ar\learn_web_development\about\index.md`


dipikabh:
We don't have Arabic as a [supported locale](https://developer.mozilla.org/en-US/docs/MDN/Community/Translated_content).

PR #28365, opened 2025-08-12
[zh-cn] update preventDefault()

rinsoko39:
### Description

Updated the description from “…if the event does not get explicitly handled, its default action should not be taken as it normally would be” to “…if the event does not get explicitly handled, its default action should not be taken as it normally would be”.

Relates to <https://github.com/mdn/content/commit/d495d8a165c7598dee686d5672268cb045f164c6>.


yin1999 (approved):
Thank you @rinsoko39

PR #28386, opened 2025-08-14
ci(pr-test): use PWD instead of CONTENT_TRANSLATED_ROOT

argl:
### Description

Fixes a double `files/` path component introduced in the last change

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -153,7 +153,9 @@ jobs:
 
           ARGS=()
           for file in "${files_to_build[@]}"; do
-            ARGS+=("-f" "$CONTENT_TRANSLATED_ROOT/$file")
+              # Remove the 'files/' prefix from the path (bash-ism, removes shortest match from beginning)
+              file_without_prefix="${file#files/}"
+              ARGS+=("-f" "$CONTENT_TRANSLATED_ROOT/$file_without_prefix")
           done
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test.yml
```diff
@@ -153,7 +153,9 @@ jobs:
 
           ARGS=()
           for file in "${files_to_build[@]}"; do
-            ARGS+=("-f" "$CONTENT_TRANSLATED_ROOT/$file")
+              # Remove the 'files/' prefix from the path (bash-ism, removes shortest match from beginning)
+              file_without_prefix="${file#files/}"
+              ARGS+=("-f" "$CONTENT_TRANSLATED_ROOT/$file_without_prefix")
```
caugner: What is the reason to go with this, rather than replicating [what we do in the content workflow](https://github.com/mdn/content/blob/791d6b6a04b1b8c09bc291ebadbc0f99a57dd5f7/.github/workflows/pr-test.yml#L144-L146)?

```suggestion
              ARGS+=("-f" "$PWD/$file")
```

argl: I was replicating what was there in the first place. But I don't have a strong opinion on either.

argl: But being somewhat in line with the content repository's workflows, I am happy to change.

caugner: Let's do this, and get this out asap.

caugner (approved):

PR #28408, opened 2025-08-14
ci(pr-test): qualify files to build properly

caugner:
### Description

Fixes the pr-test workflow.

### Motivation

Fix the `pr-test` workflow. Since `$PWD` refers to mdn/content, it doesn't work without changing folders first.

Follow-up of https://github.com/mdn/translated-content/pull/28386.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -153,7 +153,8 @@ jobs:
 
           ARGS=()
           for file in "${files_to_build[@]}"; do
-              ARGS+=("-f" "$PWD/$file")
+              file_without_prefix="${file#files/}"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test.yml
```diff
@@ -153,7 +153,8 @@ jobs:
 
           ARGS=()
           for file in "${files_to_build[@]}"; do
-              ARGS+=("-f" "$PWD/$file")
+              file_without_prefix="${file#files/}"
```
caugner, line 156: Note: I checked locally that this removes `files/` only once (given that there are some pages about `files` attributes that include `/files/` twice).

yin1999 (approved):
It looks like this PR will fix the broken workflow, can we merge it sooner?

argl (approved):

PR #28498, opened 2025-08-19
es: fix title of input button page

caugner:
### Description

Remove unnecessary space from page title.

### Motivation

Avoid break in new front-end.


Graywolf9 (approved):
Thank you so much!

PR #28556, opened 2025-08-24
[zh-CN]: update CSSStyleRule: selectorText property

KeanuoLeung:
### Description

update translation for https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleRule/selectorText
https://developer.mozilla.org/zh-CN/docs/Web/API/CSSStyleRule/selectorText

### Motivation

sync the latest content from English version of man doc


yin1999 (approved):
Thank you @KeanuoLeung. Welcome aboard 🎉

PR #28301, opened 2025-08-07
fix(workflows): use bash array instead of relying on xargs

argl:
### Description

Replaces the use of `xargs` in github actions with a bash array

### Motivation

`xargs` usage considered harmful

### Additional details

File names with special characters an spaces caused the action to fail.

### Related issues and pull requests

Fixes [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1980970)

Same as:

- https://github.com/mdn/content/pull/40648
- https://github.com/mdn/translated-content-de/pull/112

PR #28646, opened 2025-08-30
[zh-cn] update 'execution model'

jasonren0403:
### Description

Full update 'execution model', resolve bad image link


jasonren0403:
@yin1999 not sure if I've add localized image in proper way 🤔 

image

Edit: it used to work in #18149

yin1999:
> @yin1999 not sure if I've add localized image in proper way 🤔

We should use absolute paths in this step https://github.com/mdn/translated-content/blob/76a5e9fe6904a6415fca02354c937545ebca5ac3/.github/workflows/pr-test.yml#L189-L200

Which has been removed by #28301

image

Hey @caugner do you have time to have a look :)

jasonren0403:
it's still using the en-us one in review environment 🤔 

image

yin1999:
> it's still using the en-us one in review environment 🤔
> 
> image

It seems that the problem is caused by the gateway (cloud-function?), as the localized image is in [build](https://github.com/mdn/translated-content/actions/runs/17498334383/artifacts/3938785132).

/cc @caugner

yin1999 (approved):
Thank you @jasonren0403. The production environment should be able to see the localized images normally. Let's merge this PR.

PR #28681, opened 2025-09-02
zh-cn: correct the steps of using `prettier`

HermitHan:
fix first index.js (it was the same as the version after --write)

### Motivation

    The version of index.js before 'prettier --check' already uses the Prettier code style, which is different from the English type. So, I fixed it to make sure that it was a bad code style before.


yin1999 (approved):
Thank you @HermitHan. Welcome aboard 🎉

PR #28701, opened 2025-09-05
fix(pr-test): run filecheck with absolute file paths

yin1999:
### Description

Run filecheck with absolute file paths, to fix the problem mentioned in https://github.com/mdn/translated-content/pull/28646#issuecomment-3243498044

https://github.com/mdn/translated-content/pull/28646#issuecomment-3243498044


caugner (approved):

caugner:
Thanks, @yin1999, good catch!

PR #29127, opened 2025-09-20
zh-CN: fix wrong translate in `Window.crypto`

shc0743:
Original content:

虽然该属性自身是只读的，但它的所有方法（以及其子对象 {{domxref("SubtleCrypto")}} 的方法）不仅是只读的，因此容易受到 {{glossary("polyfill")}} 的攻击。

“不仅是只读的” means "Not only it is read-only", which distorted the original meaning. The correct translate should be "并非是只读的" or "并不是只读的".

### Description

**Fix a logical error in the Chinese translation of a sentence about read-only properties.** Corrects "不仅是只读的" (not only read-only) to "并不是只读的" (not read-only) to accurately reflect that the methods are writable and thus vulnerable to polyfill attacks.

### Motivation

The original translation "不仅是只读的" distorted the original meaning and created a logical contradiction. The correct meaning is that while the property itself is read-only, its methods are **not** read-only (writable), which is the key reason they are vulnerable to polyfill attacks. This fix ensures the technical accuracy and logical coherence of the documentation for Chinese readers.

### Additional details

This is a straightforward linguistic/logic fix. No changes to technical content are made, only a correction of a mistranslation.

### DIFF
--- a/files/zh-cn/web/api/window/crypto/index.md
+++ b/files/zh-cn/web/api/window/crypto/index.md
@@ -9,7 +9,7 @@ l10n:
 
 {{domxref("Window")}} 接口的 **`crypto`** 只读属性返回当前窗口的作用域的 {{domxref("Crypto")}} 对象。此对象允许网页访问某些加密相关的服务。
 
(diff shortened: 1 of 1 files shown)


jasonren0403 (approved):
LGTM 👍

PR #29439, opened 2025-09-30
fix(zh-cn): correct the description of the `endOffset` parameter of `Range.setEnd()`

yiTuChuan:
endOffset 是从节点的起始位置开始计算的偏移量，而非结束位置。

### Description

Corrects the mistranslation of the endOffset parameter for the Range.setEnd() method in the Chinese (zh-CN) documentation.
This change updates the description of endOffset from "the offset from the end of endNode" to "the offset for the end of the Range relative to the start of endNode".
This aligns the Chinese translation strictly with the original English documentation ("from the start of endNode").

### Motivation

The current Chinese documentation for endOffset contains a critical error, incorrectly stating that the offset is calculated from the node's "end position" when the correct behavior is from the "start position".
This mistranslation can severely mislead Chinese-speaking developers, potentially causing them to write incorrect code when using the Range API and waste significant time on unnecessary debugging.
This PR aims to fix this critical error, ensuring the accuracy and reliability of the Chinese documentation and providing correct guidance to the large Chinese-speaking developer community.

### DIFF
--- a/files/zh-cn/web/api/range/setend/index.md
+++ b/files/zh-cn/web/api/range/setend/index.md
@@ -1,35 +1,35 @@
 ---
-title: Range.setEnd()
+title: Range：setEnd() 方法
 slug: Web/API/Range/setEnd
 ---
 
 {{ApiRef("DOM")}}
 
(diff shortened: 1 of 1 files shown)


yin1999 (approved):
Thank you @Nasir1423. Welcome aboard :tada:

PR #29697, opened 2025-10-08
ci(pr-review-companion): use step output via env variable

fkiriakos07:
### Description

Use the defined env variable instead of github parameter directly to prevent potential command injection


caugner (approved):
LGTM.

Note that we already sanitize the input further above.

PR #29713, opened 2025-10-09
chore(ko): replace old note card style with GFM alerts syntax (part 2)

yin1999:
### Description

replace old note card style with GFM alerts syntax

### Motivation

See: #29712

Continue: #29712


wisedog (approved):
LGTM!

PR #29765, opened 2025-10-11
[zn-cn] correct incorrect Chinese translation of sort

l5ng:
fix #29764 

### Description

correct incorrect Chinese translation of sort

### Motivation

correct incorrect Chinese translation of sort

“派寻” -> "排序"

### Additional details

“派寻” -> "排序"

https://developer.mozilla.org/zh-CN/docs/Web/API/URLSearchParams/sort#%E7%A4%BA%E4%BE%8B

Image

对应英文版

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/sort#examples

Image


jasonren0403 (approved):
👍 感谢更正

PR #29830, opened 2025-10-14
fix(zh-cn): fix slug cases

yin1999:
### Description

fix slug cases

### Motivation

see: https://github.com/mdn/translated-content/actions/runs/18461876202/job/52601951782


yin1999 (commented):
I'm going to allow myself merge this PR. So that we can quickly receive upstream content changes

PR #29865, opened 2025-10-14
ci(workflows): disable automatic setup-node caching

caugner:
### Description

Sets `package-manager-cache: false` to all `actions/setup-node` workflow steps that don't already have caching configured via either `package-manager-cache` or `cache`.

### Motivation

The new default behavior introduced in [`actions/setup-node v5.0.0`](https://github.com/actions/setup-node/releases/tag/v5.0.0) is not safe.


argl (approved):

yin1999:
Hey @caugner. Are we now ready to bump [setup-node](https://github.com/mdn/translated-content/pull/28699) from 4 to 5?

PR #29937, opened 2025-10-17
[ko] Fix Bitwise OR page path

caugner:
### Description

Fixes a path that is not consistent with the slug.

### Motivation

Fixes our failing build: https://github.com/mdn/translated-content/pull/29698 was merged with test failures.


argl (approved):

1ilsang:
Oh, Thanks!

PR #30115, opened 2025-10-21
fix(zh-CN): fix four-value syntax explanation for border-width

BobLiu0518:
### Description

Fix zh-CN translation: Correct the clock position descriptions for four-value border-width syntax.

### Motivation

The original Chinese translation incorrectly described the clock positions as "12 o'clock, 2 o'clock, 3 o'clock, 4 o'clock" instead of the correct "12 o'clock, 3 o'clock, 6 o'clock, 9 o'clock" (matching the English documentation). This correction ensures accurate understanding of the border directions in four-value syntax.


jasonren0403 (approved):
LGTM, good catch 👍

PR #30312, opened 2025-11-01
zh-cn: fix wrong sentence

agoudbg:
### Description

The text `如何你的用户代理上的` makes no sense and seems like unexpectly inserted into content. I've removed it.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/audio#accessibility


jasonren0403 (approved):
👍 感谢指出

PR #30550, opened 2025-11-15
[ja] Update Document.visibilityState property description

SegaraRai:
### Description

Updated the description of the `Document.visibilityState` property for clarity and consistency in language.

### Motivation

Because the writing style was a bit confusing due to the passive voice and the use of state-indicating verbs.


mfuji09 (approved):
OKです。

PR #30578, opened 2025-11-18
fix the incorrect function name in the comment

tomo-x7:
### Description

The comment explaining `with` mistakenly uses `toSorted`.

### Motivation

Incorrect comments can lead to misunderstandings.

### Additional details

The latest English version correctly uses “with.”
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/with#calling_with_on_non-array_objects


potappo (approved):
修正ありがとうございます。OK です。

PR #30713, opened 2025-11-26
[zh-tw] fix syntax error in :lang() interactive example

tp6gw94:
### Description
Fix CSS syntax error in the `:lang()` interactive example by removing the asterisk before the pseudo-class selector.

### Motivation
The interactive example had invalid CSS syntax with `*:lang(zh_Hant)` which should be `:lang(zh-Hant)`. This correction ensures the example demonstrates proper CSS pseudo-class syntax and works correctly for readers learning about the `:lang()` selector.

### Additional details
Changed line in the interactive example:
- Before: `*:lang(zh_Hant) {`
- After: `:lang(zh-Hant) {`


jasonren0403 (approved):
👍 Good catch. Welcome aboard!

PR #30746, opened 2025-11-27
fix(ja): fix two macro invocations

caugner:
### Description

Fixes two Macro invocations in Japanese that were missing quotes.

### Motivation

These showed up in build: https://github.com/mdn/dex/actions/runs/19721304771/job/56536893288#step:22:131

#### Before

image

#### After

image


bsmth (approved):
ty

PR #30748, opened 2025-11-27
fix(ko): escape macro-like code

caugner:
### Description

Escapes macro-like code on two Korean pages, e.g. `\{{ request.path }}` instead of `{{ request.path }}`.

### Motivation

Avoid it from being recognized as a macro in rari, see: https://github.com/mdn/dex/actions/runs/19721304771/job/56536893288#step:22:132

#### Before

image

#### After

image


bsmth (approved):
TY 👍🏻

PR #30859, opened 2025-12-01
[zh-tw] add a translation for "typescript" glossary term

PikachuTW:
### Description

I add a translation for glossary/typescript.

Also i run the code in my machine for localized error messages.

By the way, I noticed that in "TypeScript also enables the programmer to annotate their code, to indicate, for example, the types of parameters to a function or **the properties of an object**:", I don't see any example for the object properties type annotation.

But i am not really sure should i open a issue in English(original) repo, which modifying the source first, or i can change in my language directly, which make there are additional contents in translation repo.

### DIFF
--- a/files/zh-tw/glossary/typescript/index.md
+++ b/files/zh-tw/glossary/typescript/index.md
@@ -1,0 +1,37 @@
+---
+title: TypeScript
+slug: Glossary/TypeScript
+---
+
+TypeScript 是一個程式語言，為 JavaScript 增加了 {{glossary("static_typing", "靜態型別")}} 功能。
+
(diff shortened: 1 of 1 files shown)


yin1999:
> But i am not really sure should i open a issue in English(original) repo, which modifying the source first, or i can change in my language directly, which make there are additional contents in translation repo.

Please file a issue in mdn/content first (Unless there are special circumstances, we do not accept additional changes.)

yin1999 (approved):
Thanks @PikachuTW

PikachuTW:
> > But i am not really sure should i open a issue in English(original) repo, which modifying the source first, or i can change in my language directly, which make there are additional contents in translation repo.
> 
> Please file a issue in mdn/content first (Unless there are special circumstances, we do not accept additional changes.)

Thanks for the information, i understand.

PR #31005, opened 2025-12-05
fix(pt-br): avoid prettier bug

caugner:
### Description

Updates a pt-BR code sample, moving the comment behind the semicolon.

### Motivation

Avoid a prettier bug that keeps reformatting it, see: https://github.com/prettier/prettier/issues/5022


bsmth (approved):
+1 from me 👍🏻

PR #31093, opened 2025-12-09
fix(fr): fixing typo of a note `lien` → `link` on role value

lgiraudel:
Do not translate attribute in code

### Description

The code example translates `role="link"` in french (`role="lien"`).

### Motivation

To avoid misunderstanding with invalid attribute value.


tristantheb (approved):
Bien vu @lgiraudel, merci pour la correction 👍🏻

PR #31178, opened 2025-12-13
zh-cn: refine import module description

java6688:
Delete the redundant identical translations and supplement and improve the translation content


yin1999 (approved):
Thank you @java6688

PR #31730, opened 2026-01-01
fix: remove ambiguous expression

easesu:


yin1999 (approved):
LGTM. Thank you @easesu

PR #31819, opened 2026-01-06
zh-cn: update the description of the `pushState()` method

ryuujo1573:
Make translation more natural in Mandarin.

### Motivation

The original sentence feels like machine translated Chinese.
Updated the structure to make it more... natural.


yin1999 (approved):
Thank you @ryuujo1573

PR #31946, opened 2026-01-09
Fix Korean mistranslation in critical rendering path

itsnowkwon:
### Description

Fixes a mistranslation in the Korean version where the subject of the sentence incorrectly stated that the server received the response.

The original English text states that the server sends the HTML response.

### Motivation

The incorrect subject could mislead readers about the HTTP request–response flow
in the critical rendering path.


1ilsang (approved):
Thanks a lot. LGTM

PR #33459, opened 2026-02-18
fix: line 85

princepink:
### Description

Verbatim and corresponding vocabulary optimization

### Motivation

The first phrase in the original text was grammatically incorrect, the second was a misuse of the corresponding vocabulary, and was also inconsistent with the same expression in line 84, so the expressions have been unified.
This translation will help Japanese readers to correctly understand the explanation of this item.


mfuji09 (approved):
ＯＫです。
なお、レビュアーは全員が日本人ですので、プルリクエストは日本語でお願いします。

princepink:
失礼いたしました。また機会がありましたら、次からは日本語で。

PR #34007, opened 2026-03-09
feat(zh-CN): translate CSS background-repeat property

ZheYanyan:
Translate the background-repeat documentation into Chinese (zh-CN).
No structural changes were made.


yin1999:
文件的路径与slug不对应

PR #34512, opened 2026-03-23
Fix index in message text

agoudbg:
### Description

Fixes wrong fixed count in the code.

### Additional details

Original article: https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API


yin1999 (approved):
Thank you @agoudbg

PR #34602, opened 2026-03-25
fix(ja): clarify debounce behavior in search event description

RiddhiPachehara:
### Description

Clarified the behavior of the `search` event in the Japanese documentation to correctly reflect debounce behavior. Updated wording to avoid misinterpretation that the event fires repeatedly.

### Motivation

The previous translation could be misunderstood as the event firing multiple times with each keystroke. However, according to the specification and English documentation, the event is debounced and only fires once after the user stops typing. This change improves accuracy and helps readers better understand the actual behavior.

### Additional details

Reference (English version):
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/search_event

Reference (Debounce explanation):
https://developer.mozilla.org/en-US/docs/Glossary/Debounce

Fixes #30141

### DIFF
--- a/files/ja/web/api/htmlinputelement/search_event/index.md
+++ b/files/ja/web/api/htmlinputelement/search_event/index.md
@@ -10,7 +10,7 @@ l10n:
 
 **`search`** イベントは、 {{HTMLElement("input")}} 要素の `type="search"` にて検索が開始されたときに発生します。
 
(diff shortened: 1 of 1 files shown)


RiddhiPachehara:
Hey @fj68 and @AllMaintainers ,

kindly review my PR and let me know if any edits are needed

fj68:
Thank you @Riddhi8077 .
I guess it's ok, so waiting for an approval by @mfuji09 .

PR #35029, opened 2026-04-08
chore: replace `Glossary("Forbidden header name")`

caugner:
### Description

Replaces 252 occurrences of `{{Glossary("Forbidden header name"}}` by `{{Glossary("Forbidden request header"}}`.

### Motivation

Resolve flaws.


tristantheb (commented):
There are glossary macros containing translations with the same incorrect key, like `{{Glossary("Forbidden header name", "en-têtes de requête interdits")}}`, it is possible that all locales are affected ^^'

caugner:
> There are glossary macros containing translations with the same incorrect key, like `{{Glossary("Forbidden header name", "en-têtes de requête interdits")}}`, it is possible that all locales are affected ^^'

Good point. If the `Glossary` macro would take the short-title of the page, would that remove the need for the localized parameter?

JuanVqz (approved):
ES Team, LGTM

tristantheb:
> > There are glossary macros containing translations with the same incorrect key, like `{{Glossary("Forbidden header name", "en-têtes de requête interdits")}}`, it is possible that all locales are affected ^^'
> 
> Good point. If the `Glossary` macro would take the short-title of the page, would that remove the need for the localized parameter?

_I mainly said that because I can’t see the pages with a second argument in the changes for `fr` (`Glossary/CORS-safelisted_response_header 38:5`, `Web/API/XMLHttpRequest/setRequestHeader 10:109`, `Web/API/XMLHttpRequest/setRequestHeader 14:130` according to the VSCode search) xD_

But that might be an idea. I don't know, however, to what extent each language version uses the `short-title`. For example, in the French glossary, very few pages (5/625) have one, as we generally follow the English page format unless we really need to make an exception.

image

yin1999 (approved):
LGTM for `l10n-zh`. Thank you @caugner

caugner:
> here are glossary macros containing translations with the same incorrect key, like `{{Glossary("Forbidden header name", "en-têtes de requête interdits")}}`, it is possible that all locales are affected ^^'

Once this PR is merged, I'll file a separate issue for those, as those need updated translations, and it's probably easier for localization teams to do this.

tristantheb (approved):
lgtm `fr`

caugner:
I'm going ahead, as this has multiple approvals, and the trivial search and replace changes are the same across all locales.

PR #35051, opened 2026-04-09
chore: fix URLs that neither exist nor redirect

caugner:
### Description

Fixes links to pages that moved in mdn/content, but don't exist or redirect in a the locale.

### Motivation

Resolve broken links, and resolve flaws.

### Additional details

Ran `cargo run -- content fix-flaws --content` with rari from https://github.com/mdn/rari/pull/618, which resolves these broken links using the en-US redirect.

### Related issues and pull requests

Part of https://github.com/mdn/fred/issues/1462.

PR #35064, opened 2026-04-10
chore(pt-br): fix URLs that neither exist nor redirect

caugner:
### Description

Fixes links to pages that moved in mdn/content, but don't exist or redirect in a the locale.

### Motivation

Resolve broken links, and resolve flaws.

### Additional details

Ran `cargo run -- content fix-flaws --content` with rari from https://github.com/mdn/rari/pull/618, which resolves these broken links using the en-US redirect.

_(Split from https://github.com/mdn/translated-content/pull/35051.)_


caugner (commented):
Self review: LGTM.

Two nits:
1. In some cases a page was removed in en-US, and now redirects to another _related yet not equivalent_ page. To fix this adequately, the affected link texts would need to be reviewed by a native speaker. Ultimately, however, this is a limitation of our redirect system, which doesn't distinguish "moves" from "deleted in favor of", and something to consider in the future.
2. In some cases, redirects point to a "conflicting" page. On the one hand, this ensures these conflicting pages are still reachable, but they would also still be reachable due to the redirect. In theory, it would be better to not follow redirects to "conflicting" pages, but because our redirect system doesn't capture the redirect chain, this is currently non-trivial.

PR #35186, opened 2026-04-13
ReferenceError: must call super constructor before using 'this' in derived class constructor を新規翻訳

sigma7863:
### Description

[ReferenceError: must call super constructor before using 'this' in derived class constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Super_not_called) ページを翻訳しました。

Relates to https://github.com/mozilla-japan/translation/issues/868


mfuji09 (approved):
OKです。

PR #35885, opened 2026-05-06
fix(fr): invalid `Array.from` interactive example output

homersimpsons:
### Description

Fixes js example output in french for Array.from

### Motivation

The current example is incorrect


tristantheb (approved):
lgtm, thanks 👍🏻

PR #35997, opened 2026-05-11
[ja] docs(http): HTTP の概要説明の文言を調整

gobo-cello:
### Description

1. `connection` を「接続」と翻訳
2. `meaning` 節前後の文言を調整

### Additional details

該当部分の原文
`HTTP follows a classical client-server model, with a client opening a connection to make a request, then waiting until it receives a response from the server. HTTP is a stateless protocol, meaning that the server does not keep any session data between two requests, although the later addition of cookies adds state to some client-server interactions.`

なし


mfuji09 (approved):
OKです。

PR #36173, opened 2026-05-22
fix(fr): example in note don't explain ths shorten version properly

regseb:
### Description

Correct the bad example so that it remains bad. Just like they do in [English](https://github.com/mdn/content/blob/52a81d8138473b6ac4bec77d0be4261cb0b76d41/files/en-us/learn_web_development/core/frameworks_libraries/react_getting_started/index.md?plain=1#L74-L78).

### Motivation

The note doesn't make sense, because the example is the same as the correct one.


tristantheb (approved):
Merci pour la correction, bien vue la duplication du code qui s'est glissée dans l'exemple secondaire 👍🏻

PR #36447, opened 2026-06-09
Web/Accessibility/ARIA/Guides/Techniques を更新

naimonon77:
### Description

元英文：https://github.com/mdn/content/blob/main/files/en-us/web/accessibility/aria/guides/techniques/index.md?plain=1


mfuji09 (approved):
OKです。

PR #36668, opened 2026-06-23
Numberの精度説明で小数点以下を有効数字に修正

Juke-99:
### Description
Update the Japanese wording for the precision of JavaScript `Number`.

This changes the phrase from "小数点以下 15 ～ 17 桁程度" to wording that describes it as approximately 15–17 significant decimal digits.

### Motivation
The current Japanese wording says "digits after the decimal point", but the surrounding explanation is about the precision of the mantissa/significand.

This precision should be described as significant decimal digits, not as a fixed number of digits after the decimal point. This helps avoid the impression that `Number` always preserves 15–17 fractional digits regardless of the magnitude of the value.

### Additional details
The corresponding English wording currently says "about 15 to 17 decimal places". Since the English wording has the same ambiguity, I plan to open a corresponding PR for `mdn/content` as well.

Relates to mdn/content#44562


mfuji09 (approved):
OKです。
なお、日本語記事のレビュアーは全員が日本人ですので、プルリクエストは日本語でどうぞ。
（英文の場合はレビューに時間がかかることがあります。）

PR #37219, opened 2026-07-23
[zh-cn] Removed redundant description for exceptions

sheep-realms:
### Description

Removed redundant description for exceptions.


yin1999 (approved):
Good catch @sheep-realms 👍

PR #37400, opened 2026-07-31
[zh-cn] update the translation of `@layer`

zyblui:
### Description

This PR updates the Simplified Chinese translation so it is in sync with upstream content.

### Motivation

The old translation marks `@layer` as a Experimental feature, which is outdated. It also lacks some detailed info.


yin1999 (approved):
Thank you @zyblui. Welcome to the community 🎉

PR #37508, opened 2026-08-03
fix: use a real minus sign in the BigInt operator example

munzzyy:
### Description

The arithmetic example on the `BigInt` page has U+2013 EN DASH where the subtraction operator belongs, and the comment under the negation line carries the same character in front of the result. It is the same two lines in `ko`, `pt-br`, `ru`, `zh-cn` and `zh-tw`. Both are a plain ASCII `-` now.

### Motivation

That block is there to be copied. With the en dash in it, pasting it into a console or a file fails to parse instead of printing the values the comments promise. `es` already has the ASCII minus, and the `fr` and `ja` pages have moved on to a different example, so I left those alone.

The last line of the block picked up a semicolon because Prettier can format the code once it parses. Without that reformat, `prettier -c` on the five files fails.

### Additional details

Before, against `origin/main`:

```console
$ git show origin/main:files/zh-cn/web/javascript/reference/global_objects/bigint/index.md | sed -n '52,74p' > /tmp/bigint.js
$ sed -n '13p' /tmp/bigint.js | hexdump -C
00000000  63 6f 6e 73 74 20 73 75  62 74 72 20 3d 20 6d 75  |const subtr = mu|
00000010  6c 74 69 20 e2 80 93 20  31 30 6e 3b 0a           |lti ... 10n;.|
0000001d
$ node --check /tmp/bigint.js 2>&1 | grep SyntaxError
SyntaxError: Invalid or unexpected token
```

The `e2 80 93` in the dump is the en dash.

After, with this branch checked out:

```console
$ { git show HEAD:files/zh-cn/web/javascript/reference/global_objects/bigint/index.md | sed -n '52,74p'; echo 'console.log(subtr, mod, bigN * -1n);'; } > /tmp/bigint-run.js
$ node /tmp/bigint-run.js
18014398509481972n 2n -18014398509481984n
```

Those three values match the three comments in the sample, and the other four locales print the same line. `markdownlint-cli2`, `prettier -c`, `autocorrect --lint` and `scripts/check-url-locale.js` are clean on the five changed files.

None that I could find.


yin1999 (approved):
LGTM for `l10n-zh`. Thank you @munzzyy

leon-win (approved):

leon-win:
Approved for `ru` also. Thank you @munzzyy !

yin1999 (commented):
Let's get it merged, changes for different locales look the same ;)

PR #38072, opened 2026-08-29
zh-cn: correct `tabindex` default value description

patienry:
### Description

Refer to https://html.spec.whatwg.org/multipage/interaction.html#the-tabindex-attribute:~:text=Return%200%20if%20this%20is%20an%20a%2C%20area%2C%20button%2C%20frame%2C%20iframe%2C%20input%2C%20object%2C%20select%2C%20textarea%2C%20or%20SVG%20a%20element%2C%20or%20MathML%20a%20element%2C%20or%20is%20a%20summary%20element%20that%20is%20a%20summary%20for%20its%20parent%20details%3B%20otherwise%20%E2%88%921.

The current description may lead readers to believe that the default `tabindex` value is `0` for all elements. In fact, only certain focusable elements have a default value of `0`.


yin1999 (approved):
这么修改是可以的，不过还是建议对全文进行更新（中文翻译有些过时了）。感谢 @patienry

PR #38234, opened 2026-09-04
fix(ru): replace unsupported `{{CSSSyntax}}` invocations

caugner:
### Description

Fix 36 pages in `ru` that call `{{CSSSyntax}}` on CSS pseudo-class, pseudo-element and keyword pages:

- Replace each invocation with the corresponding code block from the en-US page.

### Motivation

Ensure the syntax section renders, instead of failing with "CSS syntax not supported".

### Additional details

The `CSSSyntax` macro only has syntax data for properties, functions and at-rules. For the other page types, en-US spells the syntax out in a code block, which this copies over verbatim, leaving the translated prose around it untouched.


leon-win (approved):

leon-win:
Thank you @caugner !

PR #38237, opened 2026-09-04
fix(zh-tw): remove `{{deprecated_header}}` macro calls

caugner:
### Description

Remove all `{{deprecated_header}}` macro calls from zh-tw (22 files).

### Motivation

### Additional details

Audited every file against web-features and its en-US source page: 16 keep the banner, and 6 have no banner but either show no deprecation in en-US either or already carry the translated note. None loses its deprecation signal.

Escaped mentions (`\{{Deprecated_Header}}`) under `mdn/writing_guidelines/` stay; rewriting that prose belongs to a re-sync with en-US, which dropped the macro from those pages entirely. The two `page_types` templates drop it from their banner line so contributors no longer copy it.

Worth a look:

- `web/api/canvas_api/tutorial/drawing_text`: the call flagged the `mozDrawText()` and `mozMeasureText()` sections, both long gone from en-US.


yin1999 (approved):
👍

PR #38238, opened 2026-09-04
fix(ja): remove `{{deprecated_header}}` macro calls

caugner:
### Description

Remove all `{{deprecated_header}}` macro calls from ja (397 files).

### Motivation

### Additional details

Audited every file against web-features and its en-US source page: 353 keep the banner, 41 have no banner but either show no deprecation in en-US either (stale calls, for example the legacy `web/api/navigator/*` properties, and `web/api/screendetailed/top`, which en-US marks `status: experimental`) or already carry the translated note, and 3 would lose the signal, handled in #38245.

Escaped mentions (`\{{Deprecated_Header}}`) under `mdn/writing_guidelines/` stay; rewriting that prose belongs to a re-sync with en-US, which dropped the macro from those pages entirely. The 13 `page_structures/page_types/*` templates drop it from their banner line so contributors no longer copy it.

Worth a look:

- `web/api/css_object_model`: en-US marks the heading `{{deprecated_inline}}`; the list items already do.
- `web/svg/reference/element/radialgradient`: the call was inline in a definition list whose term already carries `{{deprecated_inline}}`.
- `web/api/idbtransaction` passed an obsolete `(13)` argument.

Followed by #38245, which adds the three missing notes.


caugner:
It looks like our Reviewdog workflow does not work well with stacked PRs.

mfuji09 (approved):
LGTM

PR #38240, opened 2026-09-04
fix(fr): remove `{{deprecated_header}}` macro calls

caugner:
### Description

Remove all `{{deprecated_header}}` macro calls from fr (93 files).

### Motivation

### Additional details

Audited every file against web-features and its en-US source page: 71 keep the banner, and 22 have no banner but either show no deprecation in en-US either (stale calls, for example `web/api/history/back`) or already carry the translated note, such as `mozilla/add-ons/webextensions/api/extension/sendrequest`. None loses its deprecation signal.

Two occurrences stay, mirroring mdn/content#45500: escaped mentions (`\{{Deprecated_Header}}`) under `mdn/writing_guidelines/`, and `mdn/kitchensink`, which exists to exercise every Rari macro. The 13 `page_structures/page_types/*` templates drop the macro from their banner line so contributors no longer copy it.

Worth a look:

- `web/svg/reference/element/radialgradient`: the call was inline in a definition list whose term already carries `{{deprecated_inline}}`.
- `web/mathml/reference/element/mfenced` and `mozilla/add-ons/webextensions/manifest.json/options_page`: both keep a `> [!WARNING]` saying the feature is deprecated.
- `mozilla/add-ons/webextensions/api/runtime/onbrowserupdateavailable`: en-US has `status: deprecated`, but Rari's fallback (`get_mocked_baseline_status`) only applies to slugs under `Web/` or `WebAssembly/`, so no banner is generated. en-US carries no note either, so that page loses its deprecation on both sides once mdn/content#45500 lands; probably worth raising there rather than papering over it here.


tristantheb:
I add this to the current update cycle of the week that is themed "MDN Cleanup update", where i've initialy a todo with "remove deprecated_header macros, sidebar's macros, old macros on XSL, etc." xD

tristantheb (approved):
lgtm

PR #38241, opened 2026-09-04
fix(ru): remove `{{deprecated_header}}` macro calls

caugner:
### Description

Remove all `{{deprecated_header}}` macro calls from ru (57 files).

### Motivation

### Additional details

Audited every file against web-features and its en-US source page: 43 keep the banner, and 14 have no banner but either show no deprecation in en-US either (stale calls, for example the legacy `web/api/navigator/*` properties) or already carry the translated note. None loses its deprecation signal.

Worth a look:

- `web/svg/reference/element/lineargradient` and `web/svg/reference/element/use`: the call was inline in the `xlink:href` definition list item, and neither term line carries `{{deprecated_inline}}` as en-US does.
- `web/html/reference/elements/menu`: the call flagged a `### Context menu` example section that no longer exists in en-US.
- `web/api/htmlslotelement`: still a translation of the old `HTMLContentElement` article.
- `web/javascript/reference/global_objects/date/getyear` passed an empty argument.


leon-win (approved):

leon-win:
Thank you @caugner !

PR #38243, opened 2026-09-04
fix(ko): remove `{{deprecated_header}}` macro calls

caugner:
### Description

Remove all `{{deprecated_header}}` macro calls from ko (44 files).

### Motivation

### Additional details

Audited every file against web-features and its en-US source page: 41 keep the banner, and 3 have no banner but either show no deprecation in en-US either or already carry the translated note (`web/api/window/unload_event`). None loses its deprecation signal.

Every call was a top-of-page banner, so the change is purely mechanical. The escaped mention in `mdn/writing_guidelines/page_structures/banners_and_notices` stays.

`web/css/guides/scroll_snap` is one of the stale ones: en-US documents the current, non-deprecated CSS scroll snap module, while ko still translates the old Scroll Snap Points content. Dropping the banner is correct; the page needs a re-sync.


hochan222 (approved):
LGTM :)

PR #38245, opened 2026-09-04
fix(ja): add missing deprecation notes

caugner:
### Description

Add the deprecation notes that en-US carries in prose to three ja pages.

- `games/tools/asm.js`: the specification is deprecated; WebAssembly is the alternative.
- `web/css/reference/at-rules/@font-face/font-stretch`: the descriptor was renamed to `font-width` and kept as a legacy alias.
- `web/mathml/reference/values`: the legacy MathML lengths are deprecated.

### Motivation

Removing `{{deprecated_header}}` in #38238 leaves these three pages with no deprecation signal. Rari generates the banner only when web-features marks the feature as discouraged, or when the source page has `status: deprecated` and a slug under `Web/` or `WebAssembly/` (`get_mocked_baseline_status`). `Games/Tools/asm.js` fails the slug test despite carrying `status: deprecated`, and the other two are not marked discouraged. en-US conveys the status in prose instead, which the translations lacked.

### Additional details

The `@font-face/font-stretch` wording follows the note already translated on the `font-stretch` property page. `{{cssxref("@font-face/font-width")}}` falls back to en-US, since that descriptor page is not translated yet.

Across all 397 files in #38238, these were the only three where the deprecation would otherwise become invisible.

**Depends on:** #38238.


mfuji09 (approved):
LGTM.

PR #38239, opened 2026-09-04
fix(zh-cn): remove `{{deprecated_header}}` macro calls

caugner:
### Description

Remove all `{{deprecated_header}}` macro calls from zh-cn (216 files).

### Motivation

The macro is being removed from en-US and will then be removed from Rari. Rari generates the banner itself when web-features marks the feature as discouraged, or when the source page has `status: deprecated` and a slug under `Web/` or `WebAssembly/`. Translated pages copy `status` and `browser-compat` from their en-US source page (`copy_meta_from_super`), so the banner survives.

### Additional details

Audited every file against web-features and its en-US source page: 191 keep the banner, 23 have no banner but either show no deprecation in en-US either (stale calls, for example the legacy `web/api/navigator/*` properties) or already carry the translated note, and 2 would lose the signal, handled in #38246.

Escaped mentions (`\{{Deprecated_Header}}`) under `mdn/writing_guidelines/` stay; rewriting that prose belongs to a re-sync with en-US, which dropped the macro from those pages entirely.

Worth a look:

- `web/api/css_object_model`: en-US marks the heading `{{deprecated_inline}}`; the list items already do.
- `web/http/reference/status/102`: the banner depends on the `status: deprecated` that mdn/content#45500 adds to en-US.
- `web/api/cssvalue`, `web/api/navigator/getusermedia`, `web/api/document/fgcolor`, `web/api/document/linkcolor`: the call opened the intro paragraph, so only the call is gone.

### Related issues and pull requests

Related to mdn/content#45500, which removes the last calls from en-US.
Followed by #38246, which adds the two missing notes.

PR #38246, opened 2026-09-04
fix(zh-cn): add missing deprecation notes

caugner:
### Description

Add the deprecation notes that en-US carries in prose to two zh-CN pages.

- `web/api/webvr_api/concepts`: the WebVR API is replaced by the WebXR API.
- `web/mathml/reference/values`: the legacy MathML lengths are deprecated.

### Motivation

Removing `{{deprecated_header}}` in #38239 leaves these two pages with no deprecation signal: neither en-US source page has `status: deprecated`, and web-features does not mark either feature as discouraged, so Rari generates no banner. en-US conveys the status in prose instead, which the translations lacked.

### Additional details

The WebVR wording follows the note already translated on the `Navigator.activeVRDisplays` page.

Across all 216 files in #38239, these were the only two where the deprecation would otherwise become invisible.

**Depends on:** #38239.

### DIFF
--- a/files/zh-cn/web/mathml/reference/values/index.md
+++ b/files/zh-cn/web/mathml/reference/values/index.md
@@ -14,6 +14,9 @@ l10n:
 
 ## 旧版 MathML 长度
 
+> [!WARNING]
+> 这些长度已弃用，不再推荐使用。
+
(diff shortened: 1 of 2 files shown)


files/zh-cn/web/mathml/reference/values/index.md
```diff
@@ -14,6 +14,9 @@ l10n:
 
 ## 旧版 MathML 长度
 
+> [!WARNING]
+> 这些长度已弃用，不再推荐使用。
```
yin1999: It would be best to add the word "值" to clarify that they represent values, rather than units or anything else.

```suggestion
> 这些长度值已弃用，不再推荐使用。
```

yin1999 (approved):
One suggestion, otherwise LGTM. Thank you @caugner

PR #38247, opened 2026-09-04
fix: remove obsolete numeric args from InheritanceDiagram macro

caugner:
### Description

Fix `{{InheritanceDiagram}}` macro calls that passed obsolete `width, height` numeric arguments, causing content build validation to fail with "must be a string":

- `ru/web/api/htmlbuttonelement`
- `ru/web/api/htmlscriptelement`
- `zh-cn/web/api/htmlmediaelement`

### Motivation

Ensure these pages pass content validation; the macro no longer accepts numeric width/height arguments.


yin1999 (approved):
LGTM for l10n-zh. Thank you @caugner

PR #38314, opened 2026-09-06
zh-cn: add application role

btea:
### Description
https://github.com/mdn/content/tree/main/files/en-us/web/accessibility/aria/reference/roles/application_role


yin1999 (approved):
Thank you @btea

PR #38365, opened 2026-09-08
ci: run independent steps in parallel

caugner:
### Description

Update CI workflows to overlap independent, network-bound setup steps using the GitHub Actions `parallel` and `background` step keywords.

- `auto-cleanup-bot.yml`, `sync-translated-content.yml`: run the `translated-content` and `mdn/content` checkouts in a `parallel` group.
- `pr-review-companion.yml`: run the `mdn/content` checkout alongside `Setup gcloud`, and run `Upload to GCS` as a `background` step so `Setup`/`Install (mdn/content)` (an `npm ci`) overlap with it. A `wait-all` barrier joins the upload before `Analyze PR build`.

### Motivation

Speed up CI by overlapping independent checkout, install, and upload steps instead of running them serially.

### Additional details

Uses the [parallel step keywords](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsparallel) introduced in GitHub Actions on [2026-06-25](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/). Same approach as mdn/rari#775.

`parallel` starts every nested step at once, which is not enough in `pr-review-companion.yml`, where `Setup`/`Install (mdn/content)` are a serial chain. The upload therefore uses the underlying `background` primitive plus an explicit `wait-all`. Semantics are preserved: a failed background step fails the job at the next `wait-all`, so an upload failure still stops the run before `Analyze PR build` posts a comment linking to the review deployment.

Savings, from step-level timings of recent successful runs on `main`:

| Workflow | Job median | Saving (median) | Range | Share | Samples |
| --- | --- | --- | --- | --- | --- |
| `pr-review-companion` | 50s | ~14s | 10-21s | 28% | 21 |
| `sync-translated-content` | 62s | ~8s | 6-10s | 13% | 8 |
| `auto-cleanup-bot` | 147s | ~8s | 4-9s | 5% | 8 |

The `pr-review-companion` saving grows with build size: the upload is 7s median but reaches 26s on large builds, against a steady 14s of setup and install. The two checkout workflows are bounded by the `translated-content` checkout (~19-22s) against `mdn/content` (~6-9s), hence the flat ~8s; as 8-way locale matrices, that is ~64s of runner time per daily run. The figures assume perfect overlap and so are upper bounds, though mdn/rari#775 (39s to 19s for seven checkouts) suggests contention is small.

Left unchanged: `pr-reviewdog.yml`, where the `lint-results` artifact download measures 0-1s across 25 runs, so overlapping it saves ~1s; and `pr-test.yml`, `pr-check-lint_content.yml`, `pr-check_redirects.yml`, where the `mdn/content` checkout is gated on a step (changed-file detection or `dorny/paths-filter`) that itself needs the first checkout.

**Depends on:** #38622.

Part of mdn/fred#1862.


.github/workflows/auto-cleanup-bot.yml
```diff
@@ -27,17 +27,18 @@ jobs:
...
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          repository: mdn/content
-          path: mdn/content
-          persist-credentials: false
+          - name: Checkout (content)
+            uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+            with:
+              repository: mdn/content
+              path: mdn/content
```
LeoMcA, line 40: I think this can race with the above checkout: it uses no path, so if this one starts before the above one it will delete this checkout's directory

.github/workflows/sync-translated-content.yml
```diff
@@ -34,17 +34,18 @@ jobs:
...
+          - name: Checkout
+            uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+            with:
+              persist-credentials: false
+
+          - name: Checkout (content)
+            uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+            with:
+              repository: mdn/content
+              path: mdn/content
```
LeoMcA, line 47: Same race here

LeoMcA (changes requested):

.github/workflows/auto-cleanup-bot.yml
caugner: Good catch and point, I opened https://github.com/mdn/translated-content/pull/38622 stacked below this PR to separate the checkouts.

.github/workflows/sync-translated-content.yml
caugner: dito

LeoMcA (approved):

PR #38369, opened 2026-09-08
docs(webassembly): replace `npm adduser` with `npm login`

caugner:
### Description

Update the Rust to Wasm guide to use `npm login` instead of `npm adduser`, in `ja`, `ko`, `ru` and `zh-cn`.

### Motivation

`npm adduser` was removed in npm 12, so readers are told to run a command that no longer exists.

### Additional details

- `npm login` has been the documented alias for years, so this is a rename rather than a change in instructions. In `ja`, the link now points at `npm-login` in the current (v11) CLI docs instead of `npm-adduser` in the v10 docs.
- `ko`, `ru` and `zh-cn` still show the legacy interactive prompts of a much older en-US revision. Only the command name is changed here, leaving the wider staleness for a proper re-translation.

Ports https://github.com/mdn/content/pull/45568.

Part of mdn/fred#1868.

### DIFF
--- a/files/ja/webassembly/guides/rust_to_wasm/index.md
+++ b/files/ja/webassembly/guides/rust_to_wasm/index.md
@@ -393,7 +393,7 @@ hello-wasm-0.1.0.tgz
 [INFO]: 🎒  packed up your package!
 ```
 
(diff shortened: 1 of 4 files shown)


yin1999 (approved):
LGTM for `l10n-zh`

mfuji09 (approved):
LGTM for `ja`.

PR #38384, opened 2026-09-08
fix(pt-br): correct `jsxref` arguments for property accessors

caugner:
### Description

Fix the two `jsxref` calls on `orphaned/Web/JavaScript/Reference/Errors/Undefined_prop`:

- pass `""` instead of `0` as argument 3 (anchor)
- remove the stray spaces in `{{jsxref (` and `"Operators / Property_Accessors"`
- use the correct slug casing `Operators/Property_accessors`

### Motivation

Prevent a rendering error: `jsxref` argument 3 (anchor) must be a string.

### Additional details

Argument 4 (`1`, no code formatting) is unchanged, so the rendered link text stays plain text.

Part of mdn/fred#1462.


nathipg (approved):
Thank you for your contribuition!

PR #38389, opened 2026-09-08
chore: drop stale + already-clean entries from `.prettierignore`

caugner:
### Description

Update `.prettierignore` to drop 28 entries from the block ignored pending prettier/prettier#11828:

- 18 paths that no longer exist in the repository.
- 10 files (7 fr glossary pages, 3 ja pages) whose Prettier output is already identical to the file.

### Motivation

Prevent stale ignore entries from accumulating now that the upstream issue is resolved.

### Additional details

No content files change. `npx prettier --check` passes for the 10 un-ignored files. The 6 remaining entries in the block are handled in follow-up PRs, one per locale, stacked on this one.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -11,37 +11,9 @@ build/
 /files/**/orphaned/**/*.md
 
 # XXX Ignored until https://github.com/prettier/prettier/issues/11828 is fixed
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval, as there is no code owner for `.prettierignore`.

PR #38398, opened 2026-09-08
ci(pr-labeler): label PRs by size

caugner:
### Description

Update the PR labeler workflow to also label pull requests by size (`size/xs` to `size/xl`), ported from mdn/content.

- Add a `label-by-size` job using `codelytv/pr-size-labeler` (pinned to v1.10.4) with the same thresholds as mdn/content.
- Exclude `files/*/_redirects.txt` and `files/*/_wikihistory.json` from the line count.
- Align workflow/job names and permissions with mdn/content.

### Motivation

Make it easier for reviewers to triage pull requests by size at a glance, consistent with mdn/content.

### Additional details

The `size/*` labels have been created in this repository with the same descriptions and colors as in mdn/content.

Since the workflow runs on `pull_request_target`, the new job only takes effect after this is merged.

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,18 +1,52 @@
-name: "Pull Request Labeler"
+name: PR labeler
+
 on:
   - pull_request_target
 
 permissions:
-  contents: read
+  # Patch issues, see: https://github.com/CodelyTV/pr-size-labeler/pull/89
+  issues: write
+  # Label pull requests.
   pull-requests: write
+  # Fetch files (used by actions/labeler to get config).
+  contents: read
 
 jobs:
-  triage:
-    # do not run on PRs in forks
+  # Docs: https://github.com/actions/labeler
+  label-by-path:
+    # do not run on forks
     if: github.repository == 'mdn/translated-content'
+    name: Label by path
     runs-on: ubuntu-latest
     steps:
       - uses: actions/labeler@bf12e9b00b37c5c0ca2b87b79b2daf7891dbda13 # v7.0.0
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           sync-labels: true
+
+  # Docs: https://github.com/CodelyTV/pr-size-labeler
+  label-by-size:
+    # do not run on forks
+    if: github.repository == 'mdn/translated-content'
+    needs: label-by-path
+    name: Label by size
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-labeler.yml
```diff
@@ -1,18 +1,52 @@
...
+          s_label: "size/s"
+          s_max_size: "50"
+          m_label: "size/m"
+          m_max_size: "500"
+          l_label: "size/l"
+          l_max_size: "1000"
+          xl_label: "size/xl"
+          fail_if_xl: "false"
+          message_if_xl: ""
+          files_to_ignore: |
```
LeoMcA, line 50: It looks like setting `files_to_ignore` means that any files over the first 100 are ignored:

https://github.com/CodelyTV/pr-size-labeler/blob/095a41fca88b8764fd9e008ad269bcdb82bb38b9/src/github.sh#L23

so the `size/m` and `size/l` could fail to apply if e.g. the first 100 files change one line each, and the 101st file has 1000 changes in it

I suppose it's a judgement call whether that matters more or less than counting redirects and wikihistory changes in the labelling

LeoMcA (approved):
Approved given it matches content's behaviour, albeit with the same first-100 files issue:

.github/workflows/pr-labeler.yml
caugner: Good catch, but I think we can live with this edge-case. At least those PRs won't get a `size/xs` or `size/s`.

caugner: I ran a short analysis over the 9,184 PRs merged in the last 12 months:

- 87 PRs (<1%) changed more than 100 files, mostly bot sync/auto-fix runs and locale-wide cleanups.
- For those 87, I simulated the labeler with our thresholds, summing only the first 100 files vs. all files (ignoring `_redirects.txt`/`_wikihistory.json` in both cases). The label differs in 50 of 87 cases, and always in the same direction: `size/xl` PRs get `size/m` or `size/l`, and `size/l` PRs get `size/m`.
- True distribution of those 87: 39 xl, 30 l, 18 m. With the truncation, `size/xl` would never be applied, since 100 files rarely carry >1000 changes on their own.

So as you hinted, `files_to_ignore` effectively caps the labeler at `size/l` and makes `size/xl` unreachable, which defeats the purpose for exactly the PRs where the size label matters most. Counting `_redirects.txt` and `_wikihistory.json` changes only inflates PRs that are already large, so that seems the lesser evil.

